PR #4, opened 2026-03-03
Feature/crate publish

dazwin:
Adds crate publishing to github actions.
- CARGO_REGISTRY_TOKEN needs to be added as a github secret
- Now use 'cargo release X.Y.Z --execute' to bump and release a new version
- There is a 30s delay between each publish to ensure the create exists on crates.io
- Note: Only a dry run has been tested (it needs a registry token for a full run)


RezaRassool:
Thanks Darren
  PR #7 — Bugfix/go linux aarch64                                                                                     
   
  Key improvements to setup.sh:                                                                                       
                                                  
  1. Fixes missing Linux aarch64 support — the old code hardcoded amd64 for Linux Go downloads. Now it detects $(uname
   -m) = aarch64 and uses the arm64 tarball.
  2. Eliminates code duplication — installs and upgrades now share a single code path via
  GO_ACTION="install"/"upgrade", removes ~30 lines of repeated wget/tar/rm blocks.
  3. Smart shell RC detection — instead of blindly appending to both ~/.bashrc and ~/.zshrc, it checks $SHELL and only
   updates the active shell's RC file. Uses grep -qxF for idempotency (won't add duplicate lines).
  4. Fixes regex portability — grep -oP (Perl regex) → grep -oE (extended regex), which is PR #5's fix folded into
  Linux logic too.
  5. PATH ordering fix — /usr/local/go/bin:$PATH (prepend) instead of $PATH:/usr/local/go/bin (append), so the freshly
   installed Go takes priority over any stale system Go.

  This is a good, low-risk change. The logic is clean and the testing coverage (macOS arm64/x86_64, Linux
  arm64/x86_64) is solid. Worth merging.


  - #5 — fix: use extended regular expressions (PCRE not supported on macOS)                                          
  - #4 — Feature/crate publish

PR #29, opened 2026-03-30
feat(nix): MicroVM lifecycle testing, NixOS modules, K8s manifests, cross-cache optimization

randomizedcoder:
> **Working draft** — core functionality is complete and tested on x86_64 (KVM). Cross-architecture (aarch64/riscv64 TCG) tests are still being run. Early review welcome while we finish the remaining test matrix.

## Summary

- **NixOS service modules** — `services.kwaainet`, `services.kwaainet-map-server`, `services.kwaainet-summit-server` with shared systemd security hardening baseline (score ≤ 2.5), `TimeoutStopSec=10` for clean VM shutdown
- **Kubernetes manifest generation** — Namespace, Deployment (kwaainet + map-server pod), ClusterIP Service from shared constants, validated by kube-linter
- **MicroVM lifecycle testing** — 7 VM variants × 3 architectures (x86_64 KVM, aarch64 TCG, riscv64 TCG) using astro/microvm.nix with shared `/nix/store` via 9P. Phases: boot → console → SSH → service health → security audit → deep validation → restart recovery → clean shutdown
- **Cross-compilation cache optimization** — `cross-cache.nix` overlay pins build-host-only tools (remarshal) to native pkgs, eliminating ~235 unnecessary derivations (~2.3 GiB) per cross target. Cross builds now compile only 2 Rust derivations; everything else fetches from cache.nixos.org
- **Dual-stack P2P networking** — p2pd was IPv4-only (`/ip4/0.0.0.0/tcp/8080`), preventing peer connectivity on IPv6 networks. Now listens on both IPv4 and IPv6 (`/ip6/::/tcp/8080`). Discovered and verified via two-node MicroVM tests over an IPv6 TAP bridge
- **CLI `config set initial_peers`** — enables runtime peer injection (`kwaainet config set initial_peers "/ip6/…/p2p/Qm1, /ip6/…/p2p/Qm2"`), used by two-node tests to inject bootstrap peers after both VMs are running
- **Build infrastructure** — data-driven cross-compilation matrix in flake.nix, `full-rebuild.nix` orchestrator (writeShellApplication), Makefile lifecycle/network/k8s targets, comprehensive `clean` target
- **Documentation** — cross-cache optimization, architecture diagrams, NixOS module reference, shutdown behaviour, network setup (bridge-nf-call, multiqueue TAPs), full test results

## Bug fixes found during testing

| Bug | Impact | Fix |
|-----|--------|-----|
| p2pd IPv4-only binding | Peers on IPv6 networks can't connect to each other's P2P port | Add `/ip6/::/tcp/<port>` to `-hostAddrs` (node.rs) |
| Two-node VM port collision | VM-B fails to start (serial/virtio TCP port conflict with VM-A) | Give VM-B separate `portOffset` (600/700) |
| Two-node VM MAC collision | Both VMs had identical MAC on the same bridge | Parameterise `macAddress` in mkMicrovm |
| `bridge-nf-call` filtering | Host nftables drops bridged L2 traffic between VMs | Disable `bridge-nf-call-iptables` in network-setup.nix |

## Test results (x86_64 — 2026-03-31)

| Variant | Result | Checks | Time |
|---------|--------|--------|------|
| single-node | PASS | 20/20 | ~50s |
| map-server | PASS | 23/23 | ~1m10s |
| docker | PASS | 8/8 | ~3m30s |
| k8s | PASS | 8/8 | ~1m15s |
| full-stack | WARN | 28/30 | ~35s | summit-server not yet in workspace (expected) |
| **two-node** | **PASS** | **30/30** | **2m55s** |
| **two-node-services** | **PASS** | **23/23** | **6m32s** |

All 7 x86_64 variants pass. Two-node tests verify: IPv6 ping, cross-VM TCP :8080, peer discovery, DHT bootstrap, and clean two-phase shutdown.

## Test plan

- [x] `nix flake check` passes (clippy + cargo test + smoke)
- [x] `make -j all` — native binaries
- [x] `make -j containers` — 3 OCI images
- [x] `make cross-{x86_64-musl,aarch64-gnu,aarch64-musl,riscv64-gnu}` — all 4 targets
- [x] `make test-cross` — QEMU smoke tests
- [x] `make test` + `make test-containers` — integration tests
- [x] Cross-cache: `nix build .#kwaainet-aarch64-linux-gnu --dry-run` shows 2 derivations
- [x] x86_64 lifecycle: single-node, map-server, docker, k8s — all pass
- [x] x86_64 two-node: 30/30 checks (IPv6 + TCP + P2P discovery)
- [x] x86_64 two-node-services: 23/23 checks (P2P + map-server)
- [ ] `make test-lifecycle-all-aarch64` (TCG, ~45min)
- [ ] `make test-lifecycle-all-riscv64` (TCG, ~25-50min)


RezaRassool (commented):
## Review — Storage Fabric + MicroVM Lifecycle Testing

Overall this is high-quality work with real testing behind it. The IPv6 dual-stack fix, cross-cache optimisation, and systemd hardening baseline are all valuable additions. A few things need addressing before this is ready to merge, including one genuine bug.

---

### Bug: `--initial-peers` persists to disk via the model-map save path

In `main.rs`, the CLI overrides are applied to `cfg` in memory, but `cfg.save()` is called at line 218 when the network map lookup succeeds and picks a model. Since `initial_peers` is now part of `cfg`, a `kwaainet start --initial-peers "/ip6/..."` invocation will **permanently write those peers to `~/.kwaainet/config.yaml`** if the map lookup succeeds.

The NixOS module relies on passing `--initial-peers` as a transient startup flag every restart. After the first successful boot the peers get baked into config.yaml, and on subsequent restarts the flag value conflicts with the persisted value in undefined ways.

**Fix:** Save a snapshot of the `initial_peers` before the CLI overrides block, and restore it before the `cfg.save()` call, so transient flags don't leak into the persisted config. The same applies to `--start-block`.

```rust
// Save transient overrides before the map save path
let transient_peers = args.initial_peers.clone();
let transient_start_block = args.start_block;

// ... existing overrides ...

// Inside the cfg.save() call at line 218, restore:
let peers_to_save = cfg.initial_peers.clone();
let block_to_save = cfg.start_block;
if let Some(ref p) = transient_peers { cfg.initial_peers = p.clone(); }
// Actually simpler: reload from disk after the map-save, then re-apply transient flags
```

Or simpler: pass `initial_peers` and `start_block` directly to `run_node()` rather than mutating `cfg`, so they never touch the serialisation path.

---

### Minor: `ExecStartPre = kwaainet setup` on every restart

In `nix/modules/kwaainet.nix`, `ExecStartPre` runs `kwaainet setup` before every start. In a Nix package, p2pd is already bundled in the same derivation (built from source), so `setup --get-deps` is a no-op — but `kwaainet setup` without flags still does config initialisation and a network request to check for updates. That adds latency on every service restart and fails in air-gapped environments.

Recommend either removing `ExecStartPre` entirely (since Nix handles the binary), or scoping it to `kwaainet setup --no-update-check` if the config init is needed.

---

### Question: `SocketBindAllow = ["tcp" "udp:..."]`

The `"tcp"` entry (no port restriction) combined with `SocketBindDeny = "any"` allows the service to bind any TCP port. The comment says this is for the ephemeral `127.0.0.1:0` RPC socket. That makes sense, but worth confirming: does `SocketBindAllow = "tcp"` in systemd mean "allow binding to any TCP port" or "allow TCP socket family"? The semantics changed in systemd 250 — on older systems it means socket family, on newer it means port policy. On NixOS unstable this should be ≥ 250, so the port-policy interpretation applies and this is intentional, but a comment here would help future readers.

---

### Minor: `set_var("HOME", ...)` in tests

`std::env::set_var` is safe today but became `unsafe` in Rust 1.80+ (pending stabilisation of the UB lint). The `ENV_LOCK` mutex correctly serialises these tests against each other, but it doesn't protect against parallelism with tests in other modules that happen to read `HOME`. The pattern is fine for now, but consider `KWAAINET_HOME` env var instead of `HOME` since `config.rs` already supports it — that would make the test isolation narrower and compatible with parallel test execution.

---

### What's working well

- **IPv6 dual-stack** (`node.rs`): correct fix, real gap closed. `/ip6/::/tcp/{port}` with `bindv6only=0` (NixOS default) handles both families on a single socket pair correctly.
- **Cross-compilation refactor**: the `crossTargetDefs` map eliminating 4 repetitive `import ./nix/cross.nix` blocks is clean. Data-driven is the right call here.
- **Cross-cache overlay**: 235 derivations / 2.3 GiB saved per cross target is meaningful. Pinning build-host tools (`remarshal` etc.) to `pkgsNative` so they match binary cache hashes is the correct approach.
- **`hardening.nix` baseline**: the shared systemd security profile is thorough. `CapabilityBoundingSet = ""` (empty string, not empty list) is the correct way to drop all capabilities — good catch on that systemd gotcha.
- **`config.rs` test isolation**: `ENV_LOCK` mutex + tempfile + `setup_config()` is done right. The tests covering all bool/numeric/string/invalid cases are comprehensive.
- **`summit-server or null` guard**: avoids attribute errors while the crate isn't in the workspace yet. Clean.
- **MicroVM two-node test results**: 30/30 checks including IPv6 ping, cross-VM TCP, peer discovery, DHT bootstrap, and clean shutdown at 2m55s is solid evidence this works.

---

### Merge readiness

The PR is still DRAFT with `aarch64` and `riscv64` TCG tests pending — that's the right blocker to hold on. After those pass:
1. Fix the `--initial-peers` / `--start-block` config persistence bug
2. Address `ExecStartPre`
3. Mark ready for review

Happy to discuss any of the above.

RezaRassool (commented):
Review posted as comment — see thread.

RezaRassool (commented):
## Proposed fix for the `--initial-peers` / `--start-block` persistence bug

> This follows up on my earlier review comment with a concrete diff.

### Root cause

The PR inserts the two new overrides **before** the network-map lookup block in `main.rs`:

```rust
// (line ~140 in main.rs, inside Command::Start)
if args.no_relay {
    cfg.no_relay = true;
}
// ← PR inserts these here
if let Some(sb) = args.start_block {
    cfg.start_block = sb;
}
if let Some(peers) = args.initial_peers {
    cfg.initial_peers = peers;
}

// ── Read the network map ──
if !explicit_model {
    // ...
    match map::pick_best_model(...) {
        Some(choice) => {
            // ...
            let _ = cfg.save();   // ← persists everything in cfg, including
                                  //   initial_peers and start_block
        }
    }
}
```

When the map lookup succeeds and picks a model, `cfg.save()` writes the whole config to disk, including whatever was in `cfg.initial_peers` and `cfg.start_block` at that moment. So transient CLI flags leak into `~/.kwaainet/config.yaml` on the first successful boot.

The NixOS module passes `--initial-peers` on every restart specifically because these are topology-specific bootstrap addresses that should not be persisted. After one successful boot they'd be baked into config and conflict with future invocations.

### Fix: move the two overrides to after the save point

```diff
--- a/core/crates/kwaai-cli/src/main.rs
+++ b/core/crates/kwaai-cli/src/main.rs
@@ -138,12 +138,6 @@ async fn main() -> Result<()> {
             if args.no_relay {
                 cfg.no_relay = true;
             }
-            if let Some(sb) = args.start_block {
-                cfg.start_block = sb;
-            }
-            if let Some(peers) = args.initial_peers {
-                cfg.initial_peers = peers;
-            }
 
             // ── Read the network map and select the best locally-available model ──
             if !explicit_model {
@@ -236,6 +230,13 @@ async fn main() -> Result<()> {
                 print_separator();
             }
 
+            // Apply transient overrides — these must come AFTER cfg.save() in the
+            // map-lookup block above so they are never persisted to config.yaml.
+            if let Some(sb) = args.start_block {
+                cfg.start_block = sb;
+            }
+            if let Some(peers) = args.initial_peers {
+                cfg.initial_peers = peers;
+            }
+
             // Auto-detect public IP if neither announce_addr nor public_ip is set
             if cfg.announce_addr.is_none() && cfg.public_ip.is_none() {
```

The moved block lands between `print_separator()` (line 236) and `detect_public_ip()` (line 241), after the only `cfg.save()` call in this path. Both fields are still applied to `cfg` before `run_node(&cfg)` is called at line 310, so the node sees them — they just never touch the serialised config.

### The other overrides (`--model`, `--blocks`, `--port`, etc.) are intentionally persistent

Those are already in the block before the map lookup and correctly get saved — the user explicitly asked to change their model or block count. The distinction is:

| Flag | Intent | Should persist? |
|------|--------|----------------|
| `--model`, `--blocks`, `--port` | user reconfiguring the node | ✅ yes |
| `--initial-peers`, `--start-block` | NixOS/test topology injection | ❌ no |

`--start-block` also belongs in the transient group because the NixOS module sets it declaratively per-service, not as a user preference — and the existing `kwaainet config set start_block N` path already handles the persistent case.

---

Happy to discuss if you want `--start-block` treated differently (some users may want it to persist). In that case the minimal fix is just moving `--initial-peers` after the save. Let me know how you'd like to handle it.

RezaRassool:
Closing this draft PR as it has been open for 39 days without activity. Feel free to reopen or submit a fresh PR if this work is ready for review.

PR #42, opened 2026-05-09
feat(cli): add p2p subcommand for live diagnostics

dazwin:
Adds a `kwaainet p2p` subcommand for inspecting the local node's view of the p2p network. Built while debugging dcutr / NAT / relay paths — there was no in-tree way to ask a running node "what does your p2p layer actually see?" without grepping p2pd logs.

Three read-only commands, all talking only to the local p2pd over its IPC socket:

```
kwaainet p2p info               # peer ID, listen + observed addrs, reachability verdict
kwaainet p2p peers list         # active connections, direct vs relay, bootstrap tagged
kwaainet p2p peers find <peer>  # active DHT lookup for a peer's currently-advertised addrs
```

`info` and `peers list` return p2pd's in-memory view (IDENTIFY and LIST_PEERS); `peers find` is the only one that triggers network traffic (DHT FIND_PEER, with a `--timeout` flag, default 10s).

### Sample output

```
╭─────────────────────────────────────────────────────────────────────╮
│              🛰  KwaaiNet P2P — Active Connections                  │
╰─────────────────────────────────────────────────────────────────────╯

  [direct] QmQhRuheeCLEsVD3RsnknM75gPDDqxAb8DhnWgro7KhaJc  (bootstrap)
           /ip4/18.219.43.67/tcp/8000
  [direct] Qmd3A8N5aQBATe2SYvNikaeCS9CAKN4E86jdCPacZ6RZJY  (bootstrap)
           /ip4/52.23.252.2/tcp/8000
  [ relay] 12D3KooWJckHG2fX8HCNy1n4yNyqsLbGJYoSAkeGD7h5Svf9PBzd
           /ip4/52.23.252.2/tcp/8000/p2p/Qmd3.../p2p-circuit

  💡 Total 3 connection(s): 2 direct, 1 via relay; 2 to bootstrap peer(s)
```

### What's deliberately NOT here

Several `p2p` commands were considered and dropped after looking at what p2pd's IPC actually exposes:

- **`peers info <peer>`** — `LIST_PEERS` only returns `(id, addrs)`. There's no per-connection metadata (direction, age, RTT, streams) without patching p2pd, so `info` would have been a thin filter over `list`.
- **`dht get` / `dht providers`** — these query p2pd's native libp2p Kademlia DHT, which this codebase barely populates. All the interesting records (VPK registry, shard block coverage) live in the separate hivemind DHT layer, accessed via `call_unary_handler` to `DHTProtocol.rpc_find/store` — different store, different protocol.
- **`pubsub topics`** — gossipsub is not used anywhere in this codebase (no `subscribe()` / `publish()` call sites). The PUBSUB op exists in the protobuf but would always return empty.

### Follow-up work

The genuinely high-value diagnostics for dcutr debugging require patching the maintained `docker/p2pd-patched` daemon to expose data libp2p tracks internally but doesn't surface over IPC: dcutr attempt outcomes, AutoNAT verdict, peerstore (known-but-disconnected peers), per-connection RTT and direction, relay reservations. None of those required for this PR; called out so the CLI shape doesn't have to change later.

### Implementation notes

- New file: `core/crates/kwaai-cli/src/p2p_cmd.rs` (one async fn per subcommand, ~280 lines)
- One new method on `P2PClient`: `identify_with_addrs()` — same as `identify()` but preserves the `IdentifyResponse.addrs` field instead of discarding it
- Bootstrap peer detection: same pattern as `vpk discover` and `node.rs` — prefers `cfg.initial_peers` if set, falls back to `NetworkConfig::with_petals_bootstrap()`
- Multiaddr direct/relay classification: walks the multiaddr looking for `Protocol::P2pCircuit`

### Verified

- `cargo build --package kwaainet --bin kwaainet` clean
- `cargo fmt --check` clean on all four touched files
- Manual smoke against a live local daemon (`info`, `peers list`, `peers find`) all working as expected

### DIFF
--- a/core/crates/kwaai-cli/src/cli.rs
+++ b/core/crates/kwaai-cli/src/cli.rs
@@ -128,6 +128,9 @@ pub enum Command {
     /// Manage node identity and verifiable credentials
     Identity(IdentityArgs),
 
+    /// Inspect live p2p state (identity, connected peers) via the local p2pd
+    P2p(P2pArgs),
+
     /// Manage VPK (Virtual Private Knowledge) vector database integration
     Vpk(VpkArgs),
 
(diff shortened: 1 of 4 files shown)


RezaRassool:
Merging this. One thing to clean up when `feature/public-ip` lands: `identify_full` was renamed from `identify_with_addrs` to avoid a name collision with that branch. Whichever PR arrives second should deduplicate — pick one name and remove the redundant method from `client.rs`.

PR #45, opened 2026-05-10
feat(p2p): tag trusted-relay peers in `peers list`, sort by group

dazwin:
**Stacked on #43** (`fix/relay-config-opt-in`). The first commit here is #43's commit; rebases away cleanly once #43 lands.

PR #43 added two new config fields — `force_private` and `trusted_relays` — that decide which peers a NATed node uses to obtain a relay circuit. Once a user configures them, it's useful to see at a glance which of the currently-connected peers actually come from those settings, without cross-referencing peer IDs against the config file by hand.

This PR makes `kwaainet p2p peers list` do that visualisation:

- **Tag trusted relays** as `(trusted relay)` in gold (256-color 220), parallel to the existing cyan `(bootstrap)` tag added in #42.
- **Colour the connection-kind bracket**: green `[direct]`, yellow `[ relay]`. Kind is now classified through a small `ConnKind` enum rather than a bool, so future kinds (e.g. `[ holed]` for a hole-punch upgrade detection) can slot in without disturbing call sites.
- **Sort the rows**: bootstraps first, then trusted relays, then plain direct, then via-relay; within each group, by peer ID. Stable across polls so a `watch -n 2` loop doesn't churn.
- **Extend the summary line** with a `trusted relay(s)` count alongside `bootstrap peer(s)`.

### Why 256-color for gold

The standard 16-colour palette has no orange/gold — closest is yellow, which is already used for `[ relay]`. Going 256-color (`\x1b[38;5;220m`) gives a distinct gold. Every modern terminal supports 256-color (iTerm2, kitty, alacritty, gnome-terminal, Terminal.app, Windows Terminal), so the compatibility risk is essentially zero. Inline ANSI keeps us off the dep treadmill — if colour use spreads beyond two strings, that's the time to pull in a colour crate.

### Verified

- `cargo build --package kwaainet --bin kwaainet` clean
- `cargo fmt --check` on the changed file: clean
- Manual smoke against a live local daemon: `(bootstrap)` and `(trusted relay)` tags render correctly; sort order stable across `watch`-style repeat invocations.

### DIFF
--- a/core/crates/kwaai-cli/src/config.rs
+++ b/core/crates/kwaai-cli/src/config.rs
@@ -81,6 +81,22 @@ pub struct KwaaiNetConfig {
     #[serde(default = "default_peers")]
     pub initial_peers: Vec<String>,
 
+    /// Multiaddrs of peers to use as trusted circuit relays (for AutoRelay).
+    /// When set, AutoRelay reserves circuits with these peers instead of (or
+    /// in addition to) discovering relays via the DHT. Useful when DHT
+    /// discovery is unreliable (small networks, NAT-isolated test topologies)
+    /// or when you want to force traffic through specific known-good relays.
+    /// Empty means "let AutoRelay discover relays via DHT" (the default).
+    #[serde(default)]
+    pub trusted_relays: Vec<String>,
+
+    /// When true, pre-declare this node as private (`-forceReachabilityPrivate`)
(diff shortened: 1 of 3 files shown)


dazwin:
Combining with #43 since this only makes sense once that PR's config is in place. The combined PR is now #43, with both the config and the CLI display work in scope. Closing here.

PR #48, opened 2026-05-10
chore(p2p-daemon): demote IPC wire-format logs to trace!

dazwin:
## Summary

`P2PClient` and the DHT helpers logged every protobuf request type, every hex-prefix of the encoded payload, and every connect/connected/upgrade handshake at `debug!`. Useful when debugging the IPC layer itself, useless (and noisy) for everyone running `RUST_LOG=debug` on the CLI to investigate something at a higher layer.

Compare to the existing `peers list` UX, which is clean at debug level because its higher-level helpers don't trip these wire-format messages. This PR brings the rest of the `kwaainet p2p` commands (and any other caller of `P2PClient`) up to that same level of polish.

## What moved

**Demoted to `trace!`:**
- `client.rs`:
  - "Connecting to daemon at <addr>", "Connected to daemon"
  - "Connecting to daemon stream at <addr>", "Connected to daemon stream"
  - "Successfully upgraded to persistent connection"
  - "Connecting to peer with multiaddr: …" + "Connected successfully"
  - "Request type field = N", "Encoded N bytes - hex: …", "Received response (N bytes)"
  - "STREAM_HANDLER request: type=…, protocols=…"
  - "StreamInfo received: proto=…, peer_len=…, addr_len=…"
- `dht.rs`:
  - "DHT PUT_VALUE: key len=…, value len=…"
  - "DHT GET_VALUE: key len=…"
  - "DHT FIND_PEER: peer_id len=…"
  - "DHT FIND_PROVIDERS: cid len=…, count=…"
  - "DHT PROVIDE: cid len=…"

**Kept at debug!:**
- TCP / Unix-socket retry warnings (signal a real, actionable problem)
- "Stream handler registered/removed for protocols: …"
- "Opening stream for protocols: …"
- "Upgrading to persistent connection for unary handlers"

Anyone who genuinely needs the wire-level detail can ask for it via RUST_LOG=trace.

## Test plan

- [x] cargo fmt --all -- --check clean
- [x] cargo clippy --all-targets --all-features -- -D warnings clean

### DIFF
--- a/core/crates/kwaai-p2p-daemon/src/client.rs
+++ b/core/crates/kwaai-p2p-daemon/src/client.rs
@@ -41,11 +41,11 @@ impl P2PClient {
     /// - **Windows**: `addr` should be a multiaddr like `/ip4/127.0.0.1/tcp/5005`
     /// - **Unix**: `addr` should be a multiaddr like `/unix/tmp/p2pd.sock`
     pub async fn connect(addr: &str) -> Result<Self> {
-        debug!("Connecting to daemon at: {}", addr);
+        trace!("Connecting to daemon at: {}", addr);
 
         let stream = Self::connect_stream(addr).await?;
 
-        debug!("Connected to daemon");
+        trace!("Connected to daemon");
 
         Ok(Self {
             stream,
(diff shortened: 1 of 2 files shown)


dazwin:
Withdrawing — on reflection, RUST_LOG=debug is meant to surface this kind of detail. The original complaint was about it leaking into a normal CLI invocation, but if the user has explicitly turned on debug logging, getting wire-format detail is the correct behavior.

PR #55, opened 2026-05-17
feat(rpc): in-process gRPC server for the kwaainet daemon

dazwin:
## Summary

Adds a long-running gRPC management surface (`kwaai-rpc` crate) that the Flutter GUI talks to. The listener binds early in `run_node` so clients can attach within seconds of `kwaainet start`, before p2p discovery finishes.

- New `kwaai-rpc` crate (proto + generated bindings + lifecycle tests)
- Multiplexed bidi `Session` rpc — `Ping`, `Generate`, `ShardRun`, `Status`, `Cancel` bodies with per-id correlation
- Structured `Error.Code` enum (`NO_PEERS_FOR_MODEL`, `INSUFFICIENT_COVERAGE`, `ALL_CANDIDATES_FAILED`, `MODEL_LOAD_FAILED`, …) so clients can render friendly messages without grepping strings
- `shard_run` wired through the existing block dispatcher
- `KWAAINET_NO_AUTO_UPDATE` env var so a GUI-spawned daemon doesn't surprise-update mid-session
- gRPC listener moved ahead of p2p init in `run_node` so the management surface is reachable immediately

## Test plan

- [x] `cargo build -p kwaainet --bin kwaainet` clean (one pre-existing unused-import warning)
- [x] `cargo test -p kwaai-rpc` — proto round-trip + server lifecycle/handler tests
- [x] Local: `kwaainet start` then GUI connects via `Session` and runs `shard_run`
- [ ] Reviewer: sanity-check early-bind ordering doesn't race `config` loading

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #56, opened 2026-05-17
feat(gui): Flutter desktop GUI for the kwaainet daemon

dazwin:
## Summary

Introduces `apps/gui/` — a Flutter desktop app that drives the kwaainet daemon over the new gRPC `Session` rpc (see #55). Single-process companion that hosts chat, settings, and daemon lifecycle in one window.

**Depends on:** #55 — the GUI is the client for that PR's gRPC surface.

### Key surfaces
- **Main chat:** pill-shaped composer (`KwaaiChatComposer`) with Enter-to-send / Shift-Enter newline, expanding multi-line input, inline circular Send + "New Chat" affordance to the left of the input
- **Settings:** rail nav with Status / Contribute / Network / Appearance / Developer / About tabs, including a 3-mode daemon picker (bundled / system / external) and a 10-variant theme picker with semantic info/warning/success/error colors
- **Daemon lifecycle:** spawn / monitor / external-mode short-circuit, with serialized gRPC probe + reset to avoid wedging the UI on stale sockets
- **About tab:** logo, app version via `package_info_plus`, link to https://kwaai.ai via `url_launcher`

### Notable infra
- Generated `kwaai.v1` Dart bindings consumed via `grpc-dart`
- `SessionClient` multiplexes per-id streams on one channel
- `ChatTranscriptNotifier` Riverpod family — one transcript per `ChatPath` (`shardRun` / `generateLocal`)
- Structured `ChatError(code, message)` so the UI can render friendly text for `NO_PEERS_FOR_MODEL` etc.
- macOS `network.client` entitlement added to Release (needed by `url_launcher` under the App Sandbox)
- Flutter window restoration via native AppKit `setFrameAutosaveName`

## Test plan

- [x] `flutter analyze` clean on touched files
- [x] `flutter run -d macos` — daemon connects, chat streams, theme/settings persist across restart
- [ ] Reviewer: smoke-test on Linux + Windows desktop targets (scaffolding included, untested by author)
- [ ] Reviewer: confirm `package_info_plus` / `url_launcher` plugin registration on each platform


RezaRassool:
Hey @dazwin — #55 just landed so the gRPC dependency is in. Before we can merge #56 there are a few items from the earlier review still outstanding:

1. **README** — replace the boilerplate Flutter template README with real setup instructions
2. **Hardcoded path** — `generated/README.md` has `/Volumes/Projects/kwaaiai/KwaaiNet-GUI/...` in it, needs to be a portable relative path or removed
3. **`DaemonStartResult.kind`** — convert the raw strings (`'spawned'`, `'alreadyRunning'`, etc.) to a Dart `enum`
4. **CI** — add a `flutter analyze && flutter test` step for `apps/gui/` so it's gated on the same CI as the Rust code

Happy to merge once those are in.

dazwin:
Addressed https://github.com/Kwaai-AI-Lab/KwaaiNet/pull/56#issuecomment-4480045935

RezaRassool (commented):
## Review — PR #56 Flutter GUI scaffold

Overall: well-engineered for an alpha scaffold. The gRPC session multiplexer, probe loop serialisation, and daemon lifecycle code are solid. A few issues to address before merge.

---

### Blockers / correctness issues

**1. `builtInDebugDaemonPath` resolves to `core/target/debug/kwaainet`**

`paths.dart` always looks for the *debug* artifact. In any installed scenario (no `core/target/` tree next to the app bundle) this path won't exist, so `DaemonMode.builtIn` will always present "binary not found". Fine for dev use, but the shipped app needs a release-adjacent path (e.g. sibling to the `.app` bundle, or falling back to `~/.cargo/bin/kwaainet`). Suggest adding:

```dart
// After the walk() check:
final cargo = '${Platform.environment['HOME']}/.cargo/bin/kwaainet';
if (File(cargo).existsSync()) return cargo;
```

or a separate `builtInReleaseDaemonPath` that checks the app bundle's `Contents/Resources/` on macOS.

**2. `_StubClient extends KwaaiRpcClient` fires a live probe timer in tests**

`KwaaiRpcClient`'s constructor calls `_startProbeLoop()`, which creates a `Timer.periodic` and immediately schedules a `_probe()`. The test stubs inherit this — they'll try to open a gRPC channel every 3 s during the test run. This won't fail tests (the channel open is fire-and-forget), but it leaks timers and can produce noisy stderr. Recommend either:
- Extract an `abstract class KwaaiRpcClientBase` with just the `chatStream`/`generateLocal`/`connectionState` interface, or  
- Have the stub call `super()` then `setProbingEnabled(false)` immediately.

---

### Non-blocking issues

**3. `flutter analyze` in CI doesn't enforce warnings**

The `gui:` CI job runs `flutter analyze` without `--fatal-warnings`. Warnings accumulate silently. Add `--fatal-warnings` (or at minimum `--fatal-infos`) to make the gate meaningful.

**4. Generated Dart protobuf files are checked in but no regen check**

`chat/generated/kwaai.pb.dart` etc. are committed. They can drift from the `.proto` definition in the Rust crate without anyone noticing. Suggest adding a CI step (or at least a README note in `generated/README.md`) explaining the regen command and what proto they track. A drift-detection step (`protoc --dart_out=... && git diff --exit-code`) would be ideal.

**5. `gui:` CI runs on `ubuntu-latest` only**

Unit tests pass; macOS-specific code (`macos_window_utils`, dock icon, window restoration via `setFrameAutosaveName`, `AppDelegate.swift`) isn't exercised. That's acceptable for now since #55's gRPC server also needs macOS smoke-test. The PR description already flags "reviewer: smoke-test on Linux + Windows" as unchecked — carry this forward to a follow-up issue.

**6. No test for `DaemonController` or `KwaaiRpcClient` probe state**

The existing 4 transcript tests are well-written. Worth adding at least one `DaemonController` unit test (no-binary path → error result; already-running → alreadyRunning) since that code path is exercised every launch. Not a merge blocker, but a follow-up issue would help.

---

### Positive callouts

- `setProbingEnabled(false)` suppressing the probe when the daemon is known-stopped is the right call — no spammy connect-refused every 3 s.
- `_probeInFlight` gate preventing overlapping probes is exactly the right fix for grpc-dart's channel state machine wedge.
- `KWAAINET_NO_AUTO_UPDATE=1` injection with the inline comment explaining *why* (locally-built feature work getting swapped out) — good institutional knowledge preserved.
- `SessionEndedError` vs `SessionOpError` distinction — callers can differentiate transport failures from server-side operation errors without string matching.
- The `_pendingReset` await-with-timeout pattern before re-opening a channel is solid defensive code.

---

### Summary

Two things to fix before merge: the `builtInDebugDaemonPath` production gap (or document clearly that `builtIn` mode is dev-only in this version), and the test timer leak from `_StubClient`. Everything else is follow-up material.

PR #61, opened 2026-05-20
fix: replace broken Windows CUDA fast path with reliable direct zip update

RezaRassool:
## Summary

- Removes the Windows CUDA detection block that silently downloaded a 404 HTML page on every `kwaainet update` (CI never produces a Windows CUDA artifact — only Linux CUDA builds exist)
- Always downloads the standard `kwaainet-x86_64-pc-windows-msvc.zip` (the only Windows artifact CI produces)
- Switches the post-swap restart from batch `start` builtin to `Start-Process` in PowerShell — reliable in DETACHED_PROCESS/no-console context
- Adds `scripts/update-windows.ps1` for one-liner manual update: `irm https://raw.githubusercontent.com/Kwaai-AI-Lab/KwaaiNet/main/scripts/update-windows.ps1 | iex`

## Root cause

Two stacked bugs in v0.4.67 made Windows updates silently fail:
1. `download_to` had no HTTP status check — a 404 response was written as the zip file, `Expand-Archive` failed quietly, binary stayed at old version
2. The CUDA detection fast path tried to download `kwaainet-x86_64-pc-windows-msvc-cuda-exes.zip` which CI never builds — guaranteed 404 on every Windows machine with an NVIDIA GPU

## Test plan
- [ ] `kwaainet update` on Windows downloads the correct zip and replaces the binary
- [ ] `kwaainet restart` triggers after swap (verify in `%TEMP%\kwaainet-update.log`)
- [ ] `irm .../scripts/update-windows.ps1 | iex` installs from a fresh shell

🤖 Generated with [Claude Code](https://claude.ai/claude-code)

Co-Authored-By: Claude Sonnet 4.6 <noreply@anthropic.com>

### DIFF
--- a/core/crates/kwaai-cli/src/rag_cmd.rs
+++ b/core/crates/kwaai-cli/src/rag_cmd.rs
@@ -1483,7 +1483,10 @@ async fn cmd_rebuild(
             println!();
             println!("  ▶ Step 5/8  graph seed");
             cmd_graph(
-                GraphAction::Seed { file: seed_path, kb: kb.clone() },
+                GraphAction::Seed {
+                    file: seed_path,
+                    kb: kb.clone(),
+                },
                 kb.clone(),
             )
             .await?;
@@ -1495,7 +1498,11 @@ async fn cmd_rebuild(
         println!();
         println!("  ▶ Step 6/8  graph alias-scan");
         cmd_graph(
-            GraphAction::AliasScan { auto: true, dry_run: false, min_hits: 1 },
+            GraphAction::AliasScan {
+                auto: true,
+                dry_run: false,
+                min_hits: 1,
+            },
             kb.clone(),
         )
         .await?;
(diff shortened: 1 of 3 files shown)


RezaRassool (commented):
Two issues to fix before merge:

**1. `restart` is not a valid kwaainet subcommand**

The generated PS1 ends with:
```powershell
Start-Process -FilePath '...\kwaainet.exe' -ArgumentList 'restart' -WindowStyle Hidden
```
`kwaainet restart` doesn't exist — the daemon will silently not come back after the swap. Should be `'start', '--daemon'` (matching the original batch respawn and what `update-windows.ps1` correctly uses at the bottom: `'start', '--daemon'`).

**2. `p2pd` won't match in `-Include` on Windows**

Both the generated PS1 and `scripts/update-windows.ps1` use:
```powershell
Get-ChildItem -Path $tmp -Recurse -Include '*.exe','p2pd'
```
On Windows, the p2pd binary ships as `p2pd.exe`, so the bare `'p2pd'` pattern never matches. Since `p2pd.exe` already ends in `.exe`, `'*.exe'` alone covers both binaries — just drop the `'p2pd'` entry.

Everything else looks correct: the CUDA dead-path removal is clean, `-LiteralPath` instead of `-Path` for `Expand-Archive` is the right call, `$ErrorActionPreference = 'Stop'` is good practice, and narrowing `nvidia_smi_async` cfg to Linux-only is correct now that Windows no longer uses it.

PR #68, opened 2026-05-31
fix(build): prefer /usr/lib64 for x86_64 to avoid 32-bit libcuda.so

RezaRassool:
## Summary

- Adds `[target.x86_64-unknown-linux-gnu]` section to `core/.cargo/config.toml` with `-L /usr/lib64` so `lld` searches the 64-bit library path before `/usr/lib`
- Fixes linker error `libcuda.so is incompatible with elf64-x86-64` on Linux machines where `/usr/lib/libcuda.so` is a 32-bit stub but `/usr/lib64/libcuda.so` is the correct 64-bit driver library
- A build script hardcodes `-L /usr/lib` before `-L /usr/lib64` in the linker invocation; prepending `/usr/lib64` via config.toml overrides the search order

## Test plan

- [ ] `cargo build --release --features cuda` succeeds on x86_64 Linux with CUDA installed
- [ ] Existing aarch64 `rustflags` in `config.toml` are unaffected

### DIFF
--- a/core/.cargo/config.toml
+++ b/core/.cargo/config.toml
@@ -1,2 +1,7 @@
 [target.aarch64-unknown-linux-gnu]
 rustflags = ["-C", "target-feature=+fp16,+fullfp16"]
+
+# /usr/lib/libcuda.so on this machine is 32-bit; inject the 64-bit path first
+# so lld doesn't grab the wrong one before /usr/lib64 is searched.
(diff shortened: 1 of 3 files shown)


RezaRassool:
All useful commits cherry-picked to main (config.toml rustflags fix + updater restart-daemon + tests). Superseded by commits 040aa67 and 7db511b on main.

PR #70, opened 2026-06-10
fix(kwaai-rag): replace bare lock().unwrap() with poison-safe recovery

christophermayfield:
## Summary

- Replaces all 9 bare `.lock().unwrap()` calls on the shared `GraphStore` mutex in `ingestion.rs` with `.unwrap_or_else(|e| e.into_inner())`
- Adds a `warn!` log at each recovery site so poisoning is visible in logs
- Affected functions: `extract_and_store_entities_pub` (3 sites) and `refine_low_confidence_entities` (6 sites)

**Why this matters:** If any concurrent ingestion worker task panics while holding the graph store mutex, Rust marks the mutex as _poisoned_. Every subsequent `.lock().unwrap()` on that same mutex then also panics — cascading a single task failure into a full crash of the ingestion run. Using `PoisonError::into_inner()` recovers the underlying guard; data written before the panic is preserved and only the in-flight write from the panicking task is lost.

## Test plan

- [x] `cargo build -p kwaai-rag` passes with no new warnings
- [ ] Run a normal ingestion to confirm no behaviour change on the happy path
- [ ] (Optional) Confirm that a task panic during ingestion now logs a warning and continues rather than crashing

### DIFF
--- a/core/crates/kwaai-rag/src/ingestion.rs
+++ b/core/crates/kwaai-rag/src/ingestion.rs
@@ -223,7 +223,10 @@ pub async fn extract_and_store_entities_pub(
     if graph_cfg.ec_refine_only {
         println!("  EC refine-only: skipping CC extraction, re-scoring existing entities");
         {
-            let mut g = graph_cfg.store.lock().unwrap();
+            let mut g = graph_cfg.store.lock().unwrap_or_else(|e| {
+                warn!("graph store mutex was poisoned; recovering inner value");
+                e.into_inner()
+            });
             g.sync_evidence();
             if let Err(e) = g.score_all_confidences() {
                 warn!("confidence scoring failed: {e}");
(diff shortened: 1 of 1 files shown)


RezaRassool:
Thanks Christopher — the approach is correct and this is a real hazard worth fixing.

A few notes for follow-up (none blocking merge):

1. **Extract a helper** — the same 3-line closure appears 9 times. A small free function (`fn lock_recover`) would make future coverage extensions one-liners and keep the message in one place.

2. **Add call-site context to the warning** — all 9 warn! messages are identical, so in logs you can't tell which operation triggered the recovery. Worth adding the operation name at each site.

3. **Other lock sites** — `rag_cmd.rs` (~line 812) and a few other places still use bare `.lock().unwrap()`. Good follow-up scope for a second PR.

4. **Test** — a simple unit test that poisons a `Mutex`, calls the recovery path, and asserts no panic would lock this in permanently.

Merging as-is — the fix is correct and the status quo (cascading panics) is worse than the theoretical inconsistency risk.

PR #75, opened 2026-08-01
feat(rpc): block-coverage fetch/subscribe over the Session stream

dazwin:
## What

Exposes model block coverage — which peers serve which blocks, the data `kwaainet shard chain` prints — over gRPC, as both a one-shot fetch and a live subscription.

## Why

`shard chain` is currently terminal-only. A GUI or any other RPC client has no way to see whether the model is fully covered, where the gaps are, or who is serving what, short of shelling out and scraping stdout.

## How

Extends the existing `Session` bidi stream rather than adding a top-level rpc. The proto designates `Session` as the canonical surface ("Don't add new functionality here — use Session"), and it already provides per-id correlation, streaming and `Cancel` — so subscribe/unsubscribe semantics come for free and `spawn()` needs no changes.

- `BlockCoverageRequest` on `ClientFrame` (field 15), `BlockCoverageUpdate` on `ServerFrame` (field 15).
- `subscribe: false` → one update, then `Done`. `subscribe: true` → a live feed until the client sends a `Cancel` for that op id.
- `connect_for_discovery` is extracted from the existing CLI paths in `shard_cmd.rs` so the RPC handler and the CLI resolve peers identically.

Three behaviours worth flagging for review:

**Lazy p2pd connection.** The gRPC server binds before p2pd is up during daemon startup, so the handler connects lazily: a one-shot fetch fails fast with `UNAVAILABLE`, while a subscription simply retries on the next tick. A client can subscribe immediately at startup and start receiving updates once the DHT is live, rather than polling for readiness first.

**Asymmetric clamping.** A peer's announced `end_block` may exceed the model's block count. The range is clamped before indexing the coverage bitmap, but the peer's raw `end_block` is sent verbatim on the wire — so a client sees what the peer actually claimed rather than a silently truncated value.

**Unchanged updates are suppressed.** Coverage derives from DHT records with a 360 s TTL that peers re-announce every ~300 s, so at a 5 s cadence the overwhelming majority of ticks produce a byte-identical snapshot — a frame and a client rebuild each, conveying nothing. Updates now go out only when the snapshot differs from the last one sent.

The comparison deliberately excludes `server_time`, which advances every tick by construction and would mark everything as changed, suppressing nothing. Peers compare as a set, since the DHT returns them in whatever order responders answered in and a pure reordering isn't worth waking a UI for; the per-peer key covers every field a client renders, so range/name/throughput/trust changes still register.

Suppression alone would be unsafe — silence is ambiguous, and a client can't tell "nothing changed" from "the daemon wedged" — so an unchanged snapshot still goes out every 60 s. That bounds the ambiguity and keeps a client-side "last updated" display honest while still dropping roughly eleven ticks in twelve. One-shot requests are unaffected.

## Wire compatibility

Additive only, on new field numbers, per the header rule in `kwaai.proto`. Existing clients ignore fields they don't know; a new client against an older daemon simply never receives a `blockCoverage` frame. No existing field is renumbered or repurposed.

## Testing

- `cargo check -p kwaainet` and `cargo fmt --check` clean.
- `cargo test -p kwaainet`: 55 pass. Two pre-existing failures (`server_binds_and_shuts_down_cleanly`, `chat_returns_invalid_argument_on_empty_inbound_stream`) reproduce identically on unmodified `main` when a `kwaainet` daemon holds port 8093 locally — environmental, unrelated to this change.
- `coverage_update_counts_blocks_and_clamps_ranges` covers the bitmap math (gaps, overlaps, end-past-total clamping).
- `coverage_identity_ignores_time_and_peer_order` covers the diffing: time advance and peer reordering must not count as changes; departures, re-ranging and re-tiering must.

## Note

Force-pushed to add the third commit — GitHub doesn't allow retargeting a PR's head branch, so the branch was updated in place rather than opening a new PR. No reviews had been left at that point.

### DIFF
--- a/core/crates/kwaai-cli/src/grpc_server.rs
+++ b/core/crates/kwaai-cli/src/grpc_server.rs
@@ -653,6 +653,50 @@ async fn spawn_session_shard_run(
 /// doesn't ask for a specific interval. Mirrors the map-server crawler.
 const DEFAULT_COVERAGE_INTERVAL_SECS: u64 = 5;
 
+/// How long a subscription may stay silent while coverage is unchanged.
+///
+/// Unchanged snapshots are suppressed, so without this a stable network
+/// and a wedged daemon look identical from the client's side. Sending an
+/// otherwise-redundant update this often bounds that ambiguity while
+/// still dropping the vast majority of duplicate frames — at the default
+/// 5 s cadence this keeps roughly one tick in twelve.
+const HEARTBEAT: std::time::Duration = std::time::Duration::from_secs(60);
+
+/// The parts of a [`BlockCoverageUpdate`] that decide whether it tells the
+/// client anything new.
+///
+/// Deliberately excludes `server_time`: it changes every tick by
+/// construction, so comparing whole updates would report every snapshot as
+/// changed and suppress nothing. Peers are compared as a *set* — the DHT
+/// returns them in whatever order responders answered in, and a pure
+/// reordering is not a change worth waking the UI for.
+type CoverageIdentity = (u32, u32, bool, std::collections::BTreeSet<String>);
+
+fn coverage_identity(u: &BlockCoverageUpdate) -> CoverageIdentity {
+    let peers = u
+        .peers
+        .iter()
+        .map(|p| {
+            // Every field a client renders per row, so a peer changing its
+            // range, name, throughput or trust still counts as a change.
+            // Floats are formatted rather than compared bitwise: the DHT
+            // round-trips them through msgpack, and a stable rendering is
+            // what the client actually sees.
+            format!(
+                "{}|{}|{}|{}|{:.3}|{:.3}|{}",
+                p.peer_id,
+                p.start_block,
+                p.end_block,
+                p.public_name,
+                p.throughput,
+                p.trust_score,
+                p.trust_tier,
+            )
+        })
+        .collect();
+    (u.total_blocks, u.covered_blocks, u.full_coverage, peers)
+}
+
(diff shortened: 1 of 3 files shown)


RezaRassool (approved):
Approve. Small, self-contained, and it has aged 14 days without review — no reason for that given it still rebases cleanly onto tonight's `main` after nine merges.

## Checked

**Proto tags are sequentially allocated, no collision.** `main` uses up to 14 on both sides of the Session oneof (`Cancel cancel = 14`, `StatusReply status = 14`); this takes 15 on each. Clean.

**A subscriber cannot spin the server.** `interval_secs = 0` falls back to `DEFAULT_COVERAGE_INTERVAL_SECS = 5` rather than looping flat out, which is the failure mode I check first on any subscribe-style RPC.

**Block ranges are clamped** — `e.start_block.min(total_blocks)` guards against a peer advertising a range past the model's block count, so the bitmap math cannot index out of bounds on hostile input.

`1c37d00` covering the bitmap math is the right thing to have here; that arithmetic is exactly where an off-by-one hides silently.

`5692d77` (suppress unchanged updates) is worth having in the same PR rather than later — a subscribe stream that re-sends an identical payload every 5s is the kind of thing that looks fine at three nodes and is obviously wrong at thirty.

CI green: 7 checks.

## Note on merge order

**#76 is stacked on this one** — its branch contains these three commits plus two of its own, even though both declare `base: main`. So #76's diff on GitHub shows this PR's changes too, and it needs to merge *after* this. Reviewing #76 in isolation would mean re-reading these 445 lines as if they were new.

Merging.

PR #76, opened 2026-08-01
feat(rpc): VPK storage-node discovery over the Session stream

dazwin:
Exposes `kwaainet vpk discover` as a Session op, so a client can show which VPK storage nodes the network has, how much capacity they advertise, and which of them the local node can actually reach.

### Stacked on #75 — please merge that first

This branch builds on `feat/block-coverage` and **includes its three commits**. The two features touch the same code by nature: adjacent arms in the same two `oneof`s, a shared `HEARTBEAT`, and `storage_identity` sitting alongside `coverage_identity`. Rebasing storage onto `main` alone would mean writing a standalone variant of that shared machinery and then unwinding it when #75 lands, so this deliberately depends on #75 rather than duplicating it.

Once #75 merges this diff reduces to its own two commits:

- `feat(rpc): storage-node discovery over the Session stream`
- `feat(rpc): suppress unchanged storage-discovery updates`

## Two-phase first round

The DHT lookup is fast; confirming reachability means dialling every advertised node, and on a NAT-heavy network most of those dials run to a timeout. The first round therefore emits the registry immediately and the resolved snapshot once probes land, so a client renders the node list instead of holding a blank view. Measured against the live network:

```
[1.19s] probes_pending=true   9 peers, 229.5 GB advertised, all UNKNOWN
[1.41s] probes_pending=false  3 reachable / 6 unreachable, 82.3 GB reachable
```

`skip_probes` opts out for callers that only want the advertised registry, and suppresses the pending update — nothing is pending in that case, so `probes_pending` would be a lie.

## Diffing

Later rounds send at most one update, and usually none. The pending phase runs only on the first round: re-announcing every peer as unprobed would throw a client's status column back to "checking" on a table that is already correct. Resolved snapshots are then diffed, as block coverage already does, with an unchanged snapshot every `HEARTBEAT` so silence stays unambiguous.

Over 75 s at a 5 s interval (deliberately tighter than the 30 s default, to force the issue) this sends 3 updates where the naive version sent ~26:

```
[ 1.21s] pending=true   9 peers, all unknown
[ 2.62s] pending=false  3 reachable, 82.3 GB
[67.14s] pending=false  3 reachable, 82.3 GB   (heartbeat)
```

## Probing

Probes run concurrently rather than serially as the CLI does, so a round costs about one timeout instead of one per unreachable node — 1.5 s here against the CLI's ~5 s for the same 9 nodes. Where a probe succeeds its live tenant count replaces the advertised one, since the DHT record can be a full announce cycle stale.

Subscriptions default to 30 s rather than block coverage's 5 s: VPK records are re-announced on a ~120 s cycle, so polling faster spends a dial per node per round to observe a registry that has not moved.

## Shared DHT query

The DHT query is extracted out of the CLI's `discover()` and shared with the new op, so the two paths cannot drift on how records are decoded or which entries are filtered out (dedup by peer id, and dropping self — p2pd always rejects dialling self, so leaving it in would guarantee one bogus "unreachable" row). Verified behaviour-preserving: `kwaainet vpk discover --json` returns the same 9 nodes as before the refactor.

## Wire compatibility

New fields only, with explicit numbers: `ClientFrame.storage_discovery = 16`, `ServerFrame.storage = 16`. Nothing existing changes.

## Testing

- `cargo test -p kwaainet --bin kwaainet` — 59 passed (2 grpc bind tests fail locally only when a daemon already holds port 8093)
- `cargo clippy` / `cargo fmt --check` clean
- Live-network verification against a running daemon, as above

### DIFF
--- a/core/crates/kwaai-cli/src/grpc_server.rs
+++ b/core/crates/kwaai-cli/src/grpc_server.rs
@@ -917,15 +917,60 @@ const DEFAULT_STORAGE_INTERVAL_SECS: u64 = 30;
 /// timeout, not one per unreachable node.
 const STORAGE_PROBE_TIMEOUT: std::time::Duration = std::time::Duration::from_secs(5);
 
-/// Serve a `storage_discovery` op within a Session: look up the VPK node
-/// registry in the DHT, emit it immediately, then probe each node's
-/// reachability and emit the resolved snapshot.
+/// The parts of a [`StorageUpdate`] that decide whether it tells the
+/// client anything new.
 ///
-/// Each round sends two updates because the two halves have very
+/// Deliberately excludes `server_time`: it changes every round by
+/// construction, so comparing whole updates would report every snapshot
+/// as changed and suppress nothing. Peers are compared as a *set* — the
+/// DHT returns them in whatever order responders answered in, and the
+/// wire order is already normalised by sorting, so a pure reordering is
+/// not a change worth waking the UI for.
+///
+/// `probes_pending` is not part of the identity: only resolved snapshots
+/// are ever compared, so it is false on both sides by construction.
+type StorageIdentity = std::collections::BTreeSet<String>;
+
+fn storage_identity(u: &StorageUpdate) -> StorageIdentity {
+    u.peers
+        .iter()
+        .map(|p| {
+            // Every field a client renders per row, so a node changing
+            // its capacity, tenancy, reachability or trust still counts
+            // as a change. Floats are formatted rather than compared
+            // bitwise: they round-trip through msgpack, and a stable
+            // rendering is what the client actually sees.
+            format!(
+                "{}|{}|{}|{}|{:.3}|{}|{}|{:.3}|{:.3}|{}",
+                p.peer_id,
+                p.public_name,
+                p.mode,
+                p.vpk_version,
+                p.capacity_gb,
+                p.tenant_count,
+                p.reachability,
+                p.capacity_gb_free,
+                p.trust_score,
+                p.trust_tier,
+            )
+        })
+        .collect()
+}
+
+/// Serve a `storage_discovery` op within a Session: look up the VPK node
+/// registry in the DHT, probe each node's reachability, and emit the
+/// resolved snapshot.
+///
+/// The *first* round sends two updates, because the two halves have very
 /// different latencies — the DHT answers in about a second, while probing
 /// nodes that will never answer costs a full timeout. Emitting the
 /// registry first lets a client show the node list while reachability is
-/// still resolving.
+/// still resolving, rather than holding a blank view.
+///
+/// Later rounds send at most one, and usually none: re-announcing every
+/// peer as unprobed would flicker the client's status column back to
+/// "checking" on each round, and a resolved snapshot identical to the
+/// last one is suppressed outright. See the comments inline.
 async fn spawn_session_storage_discovery(
     id: u64,
     req: StorageDiscoveryRequest,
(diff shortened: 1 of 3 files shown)


RezaRassool:
Reviewed. The storage-discovery work looks good; two things before it can land.

## It needs a restack — my fault, and the fix is one command

I squash-merged #75 just now, which created a new commit on `main` rather than replaying your three. Your branch still carries the originals, so they now conflict with their own squashed equivalent (`grpc_server.rs`). A plain `git rebase origin/main` hits that conflict.

Dropping the already-merged prefix instead is clean — I verified it:

```bash
git rebase --onto origin/main 5692d77 feat/storage-discovery-upstream
```

That leaves exactly your two storage-discovery commits on top of `main`, no conflicts, and it builds. I have not pushed anything to your branch.

Worth noting for the rest of the backlog: this happens to any stacked pair where both declare `base: main`, because GitHub cannot see the relationship and does not retarget the child the way it did for the native-p2p stack.

## On the substance (reviewed against the restacked branch)

Sound. `skip_probes` and `probes_pending` are the right shape — a caller that wants the peer list immediately should not pay for reachability probes, and one that does gets told the answer is still settling rather than being handed a half-filled list as if final. `STORAGE_PROBE_TIMEOUT` at 5s bounds each probe.

One thing worth a follow-up: **`probe_storage_peers` fans out with `futures::future::join_all`, which is unbounded.** Every storage peer is probed simultaneously through one `P2PClient`. At today's fleet size that is nothing; at a few hundred storage nodes it opens a few hundred concurrent unary calls at once, and the 5s timeout bounds each probe's latency without bounding the burst. `buffer_unordered(N)` or a capped `JoinSet` would keep it flat.

Same category as the O(n) eviction in #83: fine now, cliff later, cheap to fix while the code is fresh. Not a merge blocker at current scale.

Happy to approve once it is restacked.

dazwin:
Restacked — `git rebase --onto upstream/main 5692d771`, dropping the three block-coverage commits that #75 squash-merged. The branch is now exactly the two storage-discovery commits (`2d9e84de`, `6050602b`) on current `main`, no conflicts, and GitHub reports it mergeable.

One correction to the command for anyone following along: on my checkout `origin` is my fork and the Kwaai-AI-Lab remote is `upstream`, so it was `--onto upstream/main`. Same rebase, different remote name.

On the unbounded `join_all` in `probe_storage_peers` — you are right, and I would rather not fold it into this PR now that it is one approval from landing. Filing it as a follow-up alongside the O(n) eviction in #83, since they are the same shape of problem and worth fixing together while someone has the context. `buffer_unordered` is the cheaper of the two to reason about.

Also agreed on the stacked-pair hazard: both PRs declaring `base: main` is what hid the relationship from GitHub. Worth watching for on the rest of the backlog.

RezaRassool (approved):
Approve — the restack is exactly right.

`git log origin/main..` now shows only your two storage-discovery commits (plus the `lease_v1` test fixture, which has since landed independently as #97 and merges clean). The three already-merged block-coverage commits are gone, which is what was conflicting.

Substance re-confirmed on the rebased branch, and my read from the first pass holds: `skip_probes` / `probes_pending` are the right shape — a caller wanting the peer list immediately shouldn't pay for reachability probes, and one that does gets told the answer is still settling rather than being handed a partial list as if final. `STORAGE_PROBE_TIMEOUT` bounds each probe at 5s.

The unbounded fan-out is still there (`futures::future::join_all` at `grpc_server.rs:1229`) and that is fine by me for this merge — as I said, it is a scale concern rather than a correctness one, and today's fleet is nowhere near it. Logging it as a follow-up alongside the O(n) eviction in #83's storage: both are "fine now, cliff later", and both are cheapest to fix while the code is fresh rather than when a node is falling over.

`MERGEABLE/CLEAN` against current `main`. Yours to merge.

PR #77, opened 2026-08-01
feat(config): make UPnP port mapping configurable

dazwin:
`-natPortMap` was passed unconditionally at all three p2pd launch sites, so a node had no way to decline asking the router to open a port.

`enable_upnp` defaults to `true`, so no existing node changes behaviour on upgrade.

## Why

Turning it off is the supported way to get a genuinely NATed node without touching router settings, which makes the relay and hole-punch paths testable. It also lets an operator opt out of port mapping entirely.

The distinction matters more than it might look. While investigating DCUtR behaviour on a live node, UPnP started succeeding partway through a session — the node moved from `Private`/relay-dependent to `Public` on its own, which silently changed which code paths it exercised. Being able to hold that fixed is the difference between testing NAT traversal and hoping the router cooperates.

## Changes

- `enable_upnp: bool` on `KwaaiNetConfig`, defaulting true
- Threaded to all three `.nat_portmap(...)` call sites in `node.rs`
- Passed explicitly through `restart_p2pd_with_addrs` and `discover_and_restart_with_announce`, matching how `no_relay` already travels to those sites
- Settable via `kwaainet config set enable_upnp false`; added to the documented key list

## Testing

- `kwaainet config set enable_upnp false` round-trips to `config.yaml`
- `cargo test -p kwaainet` passes; the two `grpc_server` failures are a pre-existing port-8093 conflict with a locally running daemon, present on `main` too

Takes effect at startup, since the mapping is requested when the daemon launches.

### DIFF
--- a/core/crates/kwaai-cli/src/cli.rs
+++ b/core/crates/kwaai-cli/src/cli.rs
@@ -287,6 +287,7 @@ pub enum ConfigAction {
     /// Valid keys:
     ///   model, blocks, start_block, port, use_gpu, log_level,
     ///   public_name, public_ip, announce_addr, no_relay, native_p2p,
+    ///   enable_upnp,
     ///   vpk_enabled, vpk_mode, vpk_local_port,
     ///   auto_rebalance, rebalance_interval_secs, rebalance_min_redundancy
     ///
--- a/core/crates/kwaai-cli/src/config.rs
+++ b/core/crates/kwaai-cli/src/config.rs
@@ -142,6 +142,19 @@ pub struct KwaaiNetConfig {
     /// lands and the cutover in Phase 5 flips it.
     #[serde(default)]
     pub native_p2p: bool,
+    /// Ask the local gateway to map our listen port via UPnP/IGD
+    /// (`-natPortMap`).
+    ///
+    /// On by default, which is the behaviour this replaces — the flag was
+    /// previously passed unconditionally. Turning it off is how you get a
+    /// genuinely NATed node without touching router settings, which makes the
+    /// relay and hole-punch paths testable; it also lets an operator decline
+    /// to have the node ask the router to open ports at all.
+    ///
+    /// Takes effect at startup, since the mapping is requested when the daemon
+    /// launches. Changing it needs a restart.
+    #[serde(default = "default_enable_upnp")]
+    pub enable_upnp: bool,
 
     #[serde(default)]
     pub health_monitoring: HealthConfig,
@@ -577,6 +590,9 @@ fn default_trusted_relays() -> Vec<String> {
 fn default_force_private() -> bool {
     true
 }
+fn default_enable_upnp() -> bool {
+    true
+}
 fn default_api_endpoint() -> String {
     "https://map.kwaai.ai/api/v1/state".to_string()
 }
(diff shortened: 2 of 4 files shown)


RezaRassool:
Reviewed — the change is right, but it now conflicts with `main` and needs a rebase.

## Conflict

`core/crates/kwaai-cli/src/config.rs` and `cli.rs`. #85 (`run_node takes the native path behind native_p2p`) landed tonight and added a `native_p2p` field to the same config struct, adjacent to where this adds `enable_upnp` — including the doc-comment list of settings that both PRs extend. Mechanical to resolve; a plain `git rebase origin/main` will show it.

## On the substance — no objections

**Backward compatible by construction.** `#[serde(default = "default_enable_upnp")]` returning `true`, plus `enable_upnp: default_enable_upnp()` in `Default`, means an existing `config.yaml` with no such key behaves exactly as today. Nobody's node changes on upgrade.

**Every call site converted.** All three `.nat_portmap(true)` sites in `node.rs` now read the config, and the only remaining literal is the builder's own parameter. No half-migrated path left behind.

**The rationale is the useful part.** Being able to turn UPnP off is what makes a genuinely NATed node without touching router settings — which is precisely how the relay and hole-punch paths become testable. I spent a chunk of this week debugging relay behaviour on the live fleet and would have reached for this more than once; a node that quietly maps its own port is a node whose relay path you cannot exercise.

The doc comment noting it takes effect at startup and needs a restart is the right thing to say, since the mapping is requested when the daemon launches.

Rebase and I will approve.

dazwin:
Rebased onto current `main` — `dbfcf7d8`, mergeable.

The conflict was as you described plus one more site: `config.rs` (the `native_p2p` doc block and the `Default` impl, both adjacent to `enable_upnp`), `cli.rs` (the valid-keys doc list), and the `set` key dispatch at the bottom of `config.rs`, which #85 also extended. All three were purely additive — both fields survive, nothing from #85 dropped.

Verified `cargo check -p kwaainet` clean, and the diff is still only the three `nat_portmap` call sites plus the config plumbing.

One thing I noticed but deliberately did not touch: #85's doc comment on `native_p2p` says NAT traversal is "p2pd-only" and lists UPnP among the things the native path does not do yet. That stops being true when #86 lands. It is #86's line to update, not this PR's, so I have left it alone rather than editing a neighbouring PR's prose from here.

RezaRassool (approved):
Approve — as promised once rebased.

Verified the conflict resolved correctly rather than just taking GitHub's word for it: `config.rs` now carries **both** `native_p2p` (from #85) and `enable_upnp` side by side, and all three `.nat_portmap()` sites in `node.rs` still read the config with no literal left behind. That is the failure mode a hand-resolved struct conflict invites — one field silently dropped, or one call site reverted to the hardcoded value — and neither happened.

Everything from the first pass still stands: `#[serde(default = "default_enable_upnp")]` returning `true` means no existing node changes behaviour on upgrade, and the rationale is the genuinely useful part — being able to decline the port mapping is what makes a truly NATed node without touching router settings, which is how the relay and hole-punch paths become testable at all.

`MERGEABLE/CLEAN` against current `main` (post-#97). Yours to merge.

PR #78, opened 2026-08-03
fix(kwaai-cli): log full error chain on auto-update install failure

RezaRassool:
## Summary
- \`node.rs:2405\` logged \`{e}\` on auto-update install failure — anyhow's Display, which only prints the top-level context string ("Could not rename kwaainet.exe — is another process holding it?"), never the underlying OS error.
- At least 6 prior commits (\`0ff5abc\`, \`22aa3eb\`, \`77215e5\`, \`35b7ef9\`/\`daff4a7\`, \`7c57102\`, \`33c5015\`) have attempted to fix Windows auto-update rename failures, all working from that same static message — none of them ever saw the actual Win32 error code.
- Reproduced the rename step directly on real Windows hardware (full daemon, \`DETACHED_PROCESS\`, live \`p2pd\` child, exact same \`fs::rename\` call \`install_update()\` uses) and it succeeded every time — the \`FILE_SHARE_DELETE\` assumption in the code comment held up in this repro. So whatever actually caused the July 27 failure on this machine wasn't captured by any existing test or log line.
- Changed to \`{e:?}\` so the next real occurrence logs the full anyhow chain, including the real OS error, instead of the canned context string.

## Test plan
- [x] \`cargo build -p kwaainet --release\` — clean build, no new warnings from this change
- [x] Reproduced the rename call live against a copy of the real binary + real daemon process tree (see conversation) — confirmed the diff doesn't change success/failure behavior, only what gets logged on failure
- [ ] Next real Windows auto-update failure should be diagnosed from the actual logged error instead of guesswork

### DIFF
--- a/core/crates/kwaai-cli/src/node.rs
+++ b/core/crates/kwaai-cli/src/node.rs
@@ -2402,7 +2402,7 @@ async fn maybe_auto_update() -> Option<String> {
     );
 
     if let Err(e) = checker.install_update(&update.version).await {
-        warn!("Auto-update install failed: {e}");
+        warn!("Auto-update install failed: {e:?}");
         return None;
     }
 


RezaRassool:
Reviewed (as the author, so this is a note rather than an approval — GitHub will not let me approve my own PR, and this still needs someone else's eyes before merge).

The change is correct and the reasoning holds up better now than when it was opened 11 days ago.

`{e}` on an `anyhow::Error` prints only the outermost message. `{e:?}` prints the whole chain, which for an install failure is the difference between "Auto-update install failed: io error" and knowing which path, which permission, which rename. That is exactly the information missing from every auto-update failure report we have.

Worth noting this is not an isolated papercut — the same swallowed-error pattern has cost real debugging time twice in the last few days:

- `service.rs` logged a circuit-listen failure as `error = %e`, which came out **empty** on every occurrence. The only reason the relay-candidate bug in #86 was diagnosable at all is that the address in the message gave it away; the error itself said nothing. That one is fixed on `proposal/relay-candidate-fix`.
- The same shape appears elsewhere in `node.rs` and `relay_manager.rs`.

So this is a one-line fix to one instance of a class. Worth a follow-up sweep of `error = %e` and `{e}` on error types across the crate — mechanical, and it converts a category of silent failures into diagnosable ones.

Still rebases cleanly onto tonight's `main` and CI is green.

dazwin (approved):
Approving — the change is correct and does what the summary claims.

Verified the masked error is real: `updater.rs:280` attaches `.context("Could not rename kwaainet.exe — is another process holding it?")` to the `fs::rename`, and anyhow's `Display` prints only that outermost context, discarding the underlying `io::Error`. `Debug` prints the whole chain, so the next real Windows failure will actually carry the Win32 error. Behavior is otherwise unchanged — only the text of an already-failing log line, no callers affected.

Non-blocking nit: the crate's prevailing idiom for logging an anyhow chain is `{e:#}` (26 uses, including every adjacent `warn!` in `node_native.rs` and both `main.rs` sites handling this same `install_update` failure). The 7 existing `{e:?}` uses are all in `llama_local.rs` formatting foreign non-anyhow errors. Both expose the cause chain, but `{e:#}` stays on one line, whereas `{e:?}` also prints a backtrace when `RUST_BACKTRACE` is set — plausible during exactly the Windows debugging session this serves, and it would bury the warning. Worth considering, but not a blocker.

PR #79, opened 2026-08-05
build: patched deps via checksum-pinned fetch script (native-p2p 0/7)

dazwin:
| # | PR | What |
|---|----|------|
| 0 | #79 | build: patched deps via fetch script |

---

## What this adds

A build bootstrap step that fetches `multistream-select` 0.13.0 from crates.io and applies a one-rule patch allowing slash-less libp2p protocol IDs, wired in via `[patch.crates-io]`. This is the foundation the rest of the native-p2p stack builds on — without it a Rust node cannot speak hivemind's protocol names at all.

## Why

Hivemind negotiates bare handler names (`DHTProtocol.rpc_store`, `hello`) as libp2p protocol IDs. go-libp2p accepts them; upstream `multistream-select` rejects any name without a leading `/`, in both its `TryFrom` impls. That means a native rust-libp2p node can neither call a hivemind unary handler nor serve one to a Go peer. The restriction is local validation inside rust-libp2p, not part of the wire protocol — go-libp2p negotiates these names on the live network today. Every later PR in this stack depends on the relaxed rule.

## What's in it

- `core/patches/multistream-select.patch` — a 46-line patch relaxing the leading-slash requirement. Nothing is vendored: the repo carries only the patch, and the expanded source is gitignored.
- `core/patches/fetch-multistream-select.sh` — idempotent fetch/verify/apply. Pins version 0.13.0 and a SHA-256 of the crate tarball, extracts to `core/patches/multistream-select`, applies the patch, and writes a stamp file so re-runs are no-ops unless the patch changes.
- `core/patches/README.md` — documents the delta and the rationale.
- `core/Cargo.toml` — `[patch.crates-io] multistream-select = { path = "patches/multistream-select" }`, plus `exclude = ["patches/multistream-select"]` so cargo does not auto-adopt it as a workspace member and pull its dev-deps into the lock.
- **Build entry points, all for the same reason.** Because `[patch.crates-io]` points at a path, cargo cannot even *parse* the workspace until that path exists. So the fetch step is added to `setup.sh`, the eleven GitHub Actions workflows (`ci.yml`, the six `ci-kwaai-*.yml`, `publish-crates.yml`, `release.yml`, `release-riscv.yml`, `test-cuda-cache.yml`), `flake.nix`, and the nix crane/cross builders (`nix/crane.nix`, `nix/cross.nix`).

## Risk

**No default runtime behaviour changes.** This PR adds no application code — nothing in any crate is added, removed or rewired. The only functional delta is that `multistream-select` now permits protocol names it previously rejected locally; no code in the tree yet negotiates such a name (that starts in PR3).

What a reviewer should look at closely:

- **The bootstrap requirement is the real risk, and it is a build-time one.** Anyone running `cargo` in a checkout that has not run `core/patches/fetch-multistream-select.sh` will get a workspace-parse failure, not a helpful error. Please confirm your own build entry point is covered — if there is a CI workflow, container image, or local script not in the list above, it will break.
- The fetch script requires network access to `static.crates.io` at build time. Air-gapped or offline builds will need the directory pre-populated.
- The checksum pin is what makes the fetch trustworthy; verify `SHA256=ea0df8e5…` matches the published crate.

## Test coverage

No new tests. This PR is build plumbing; the patched behaviour is exercised by PR3's tier-08 interop test (`slash-less negotiation against a real go-libp2p daemon`) and by PR1's `slashless_protocol_negotiates`. The fetch script's own correctness is checked by the checksum verification and the idempotency stamp rather than by a test.

## Stacked on

`upstream/main`. This is the base of the 8-PR native-p2p stack.

---

- `cargo test --workspace`: **633 passing, 0 failures.**

---


dazwin:
Superseded by #91: identical branch content, re-opened from the in-repo branch so the native-p2p chain can be linked as a GitHub stack (cross-fork stacks are unsupported in the preview).

PR #80, opened 2026-08-05
feat(hivemind-dht): hivemind unary wire codec + p2pd interop gate (native-p2p 1/7)

dazwin:
| # | PR | What |
|---|----|------|

---

## What this adds

A codec for hivemind's unary-RPC wire format — the uvarint-framed `PersistentConnectionRequest` envelopes that actually travel on a libp2p stream between peers — plus the fix for a long-standing bug in which our reply envelope used the wrong message type entirely.

## Why

Every later PR in this stack needs to speak hivemind's unary RPC natively. That format was not previously written down anywhere in this repo, and where it *was* implemented (`kwaai-p2p-daemon/src/stream.rs`) it was wrong: the reply was encoded as a `PersistentConnectionResponse`, a message that exists only on the daemon's local control socket, not on the peer-to-peer wire. Getting the envelope right is a prerequisite for a native responder, and documenting it as a schema of record stops the same mistake recurring. This PR also establishes the interop test tier that every subsequent phase gates on.

## What's in it

**The codec — `core/crates/kwaai-hivemind-dht/src/wire.rs`**

- Encode/decode for both directions of the hivemind unary exchange:
  - caller → `uvarint(len) ++ PersistentConnectionRequest{callId, callUnary}`
  - caller ← `uvarint(len) ++ PersistentConnectionRequest{callId, unaryResponse}`
  - on a stream whose protocol ID is the bare handler name with no leading slash (e.g. `DHTProtocol.rpc_store`), with `data` carrying the raw application protobuf.
- uvarint framing with a 10 MiB cap matching the Go daemon's `persistentConnMsgMaxSize`.
- An async `read_framed` that handles multi-byte varints and partial reads without over-consuming into the next frame.
- `core/crates/kwaai-hivemind-dht/proto/persistent_conn.proto` — the schema of record. A wire-compatible subset of `kwaai-p2p-daemon/proto/p2pd.proto` with field numbers copied verbatim. As with `proto/dht.proto` and `protocol.rs`, the Rust types are hand-written prost derives; this crate has no `build.rs`.

**The fix — `core/crates/kwaai-p2p-daemon/src/stream.rs`**

- `wrap_stream_handler_response` now emits a `PersistentConnectionRequest` with the `unaryResponse` arm (field 4) instead of a `PersistentConnectionResponse` (whose `callUnaryResponse` sits at field 2). Verified against go-libp2p-daemon `persistent_stream.go`, where `Daemon.exchangeMessages` reads the reply into `&pb.PersistentConnectionRequest{}` and calls `GetUnaryResponse()`.
- The old encoding, decoded as a `PersistentConnectionRequest`, reinterprets a `CallUnaryResponse{response: "pong"}` as `AddUnaryHandlerRequest{proto: "pong", balanced: false}` — a DHT payload read as a handler-registration name. Go's gogo proto2 rejects it outright since `required bool balanced` is absent. Either way `GetUnaryResponse()` is nil.

**Design record — `docs/NATIVE_P2P_MIGRATION.md`** (new, 331 lines): the wire format, the verification record, and the open/resolved verification items the later phases reference.

## Risk

**No default runtime behaviour changes** in the sense that matters for a dark launch — but note this PR does contain one genuine bug fix on shared code, so it deserves a closer read than the rest of the stack.

- `wrap_stream_handler_response` is corrected rather than added. **It has no callers anywhere in the workspace today** (verified by grep across `core/crates` at this commit), so in practice the change is confined to a function that nothing invokes; the native responder in PR3 is its first consumer. If a reviewer knows of an out-of-tree caller, that is the one place to look.
- `wire.rs` and `persistent_conn.proto` are new modules with no consumers at this commit.
- The `07_wire_interop` tier is gated behind `KWAAI_INTEGRATION_TESTS=1` like the other integration tiers, so it does not run in the default test path.

## Test coverage

**Unit tests in `wire.rs`** — golden byte layouts (including an explicit assertion that the reply oneof arm sits at field 4, not the historical field 2), varint width boundaries at 0/127/128/16383/16384, oversized-frame rejection on the length declaration alone, runaway varints, truncated payloads, and decode of a `DaemonError` payload.

**Unit tests in `stream.rs`** — that the emitted bytes decode as a `PersistentConnectionRequest` with the `unaryResponse` arm; that the arm's tag byte is `0x22` (field 4) rather than `0x12` (field 2); the unwrap/wrap round trip a stream handler performs; and the mis-decode above kept as an explicit regression witness.

**Tier 07 — `core/crates/kwaai-network-tests/tests/07_wire_interop.rs`** (7 tests, real go-libp2p-daemon on the other end of the stream):

- our responder frames are accepted by a p2pd caller, and the old `PersistentConnectionResponse` shape is provably rejected — the historical bug was real;
- our caller frames are accepted by a p2pd responder, including an empty payload and a multi-byte uvarint length prefix;
- the error arm propagates to the caller;
- multistream-select negotiates the slash-less `DHTProtocol.rpc_ping` protocol ID verbatim — this is the empirical proof that PR0's patch does what it claims.

Harness: `TestNode::new_wire_peer()` spawns a bare p2pd (no DHT, relay or bootstrap) on its own tmpdir socket with a throwaway identity.

**One measured finding recorded for later phases:** on a raw stream handler, `callUnary.peer` arrives unrewritten — it carries the callee's ID as the caller sent it. Only the unary-handler dispatch path rewrites it to the caller. PR3's native responder therefore takes caller identity from the connection, never from this field.

## Stacked on

`native-p2p-pr0-build-patched-deps`

---

- `cargo test --workspace`: **663 passing, 0 failures.**

---

### DIFF
--- a/core/Cargo.lock
+++ b/core/Cargo.lock
@@ -3062,6 +3062,7 @@ dependencies = [
  "tokio",
  "tracing",
  "tracing-subscriber",
+ "unsigned-varint 0.7.2",
 ]
 
 [[package]]
@@ -3111,6 +3112,7 @@ dependencies = [
  "tokio",
  "tracing",
  "tracing-subscriber",
+ "unsigned-varint 0.7.2",
 ]
 
 [[package]]
--- a/core/crates/kwaai-hivemind-dht/Cargo.toml
+++ b/core/crates/kwaai-hivemind-dht/Cargo.toml
@@ -23,6 +23,7 @@ serde_json = { workspace = true }
 rmp-serde = { workspace = true }  # MessagePack
 prost = { workspace = true }       # Protobuf
 bincode = { workspace = true }
+unsigned-varint = { workspace = true }  # uvarint framing on the hivemind unary wire
 
 # Error handling
 thiserror = { workspace = true }
--- a/core/crates/kwaai-hivemind-dht/proto/persistent_conn.proto
+++ b/core/crates/kwaai-hivemind-dht/proto/persistent_conn.proto
@@ -1,0 +1,85 @@
+syntax = "proto2";
+
+// Hivemind unary-RPC wrapper as it appears ON THE LIBP2P WIRE between peers.
+//
+// This is a minimal, wire-compatible subset of go-libp2p-daemon's `p2pd.proto`
+// (see `crates/kwaai-p2p-daemon/proto/p2pd.proto`, package `p2pd.pb`). Field
+// numbers and types are copied verbatim — do not renumber.
+//
+// The Go daemon (`persistent_stream.go`) drives the peer-to-peer exchange like
+// this:
+//
+//   Caller side (`Daemon.exchangeMessages`):
+//     1. `host.NewStream(peer, protocol.ID(proto))` — the libp2p protocol ID is
+//        the BARE handler name, no leading slash (e.g. `DHTProtocol.rpc_store`).
+//     2. Writes ONE uvarint-length-delimited `PersistentConnectionRequest`
+//        with `callId` + `callUnary{peer, proto, data}`. `data` is the raw
+//        application protobuf (StoreRequest / FindRequest / PingRequest).
+//     3. Reads ONE uvarint-delimited `PersistentConnectionRequest` back and
+//        takes `remoteResp.GetUnaryResponse()`.
+//     4. Stream closes.
+//
+//   Responder side (`Daemon.persistentStreamHandler`):
+//     Reads the `PersistentConnectionRequest`, overwrites
+//     `callUnary.peer` with the CALLER's peer ID (so a value set by the
+//     caller — as hivemind's Python client does — is ignored), dispatches to
+//     the registered handler, and writes back a `PersistentConnectionRequest`
+//     carrying `unaryResponse` (field 4).
+//
+// NOTE THE ASYMMETRY: the reply on the libp2p wire is a
+// `PersistentConnectionRequest` with the `unaryResponse` oneof arm set — NOT a
+// `PersistentConnectionResponse`. `PersistentConnectionResponse` exists only on
+// the daemon's local control socket. Encoding the reply as a
+// `PersistentConnectionResponse` makes Go callers fail to unmarshal, because
+// field 2 there is `CallUnaryResponse` while field 2 of
+// `PersistentConnectionRequest` is `AddUnaryHandlerRequest` — whose proto2
+// `required` fields are absent.
+//
+// This file documents the schema of record. The Rust types live in `src/wire.rs`
+// as hand-written `prost::Message` derives, matching how `src/protocol.rs`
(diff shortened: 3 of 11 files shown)


RezaRassool (approved):
Approve. I ran the interop tier against a real p2pd rather than trusting the green CI tick — see below, because the two are not the same thing here.

## Substance

The proportions are right: 723 lines of codec against 576 lines of interop tests and 331 of schema-of-record docs. The codec does the things a wire codec has to do and that are easy to skip:

- `MAX_FRAME_LEN` (10 MiB) bounds a length prefix that arrives from a remote peer
- `MAX_VARINT_LEN` bounds the varint decode itself
- the varint is read byte-at-a-time so it cannot over-consume into the payload
- proto2 `required` labels, matching the finding that Go enforces `required` on unmarshal — an omitted empty `peer`/`data` gets the frame dropped and the stream reset

The `stream.rs` change is the part carrying real risk, because that is **live code on today's p2pd path**: the reply envelope moves from `PersistentConnectionResponse` (a control-socket-only message) to `PersistentConnectionRequest{unaryResponse}`. That is a genuine bug fix rather than a refactor, and it is precisely the change the two most important interop tests cover.

## Verification

`07_wire_interop` is gated behind `KWAAI_INTEGRATION_TESTS=1`, so CI does not run any of it. I ran it locally against real p2pd instances:

```
slashless_protocol_negotiates ................. ok
multibyte_length_prefix_accepted_by_daemon .... ok
raw_wire_responder_is_accepted_by_daemon_caller ok
daemon_to_daemon_unary_round_trip ............. ok
raw_wire_caller_against_daemon_handler ........ ok
old_wrapper_shape_is_rejected_by_daemon_caller  ok
raw_wire_responder_error_arm .................. ok

test result: ok. 7 passed; 0 failed; finished in 0.95s
```

`old_wrapper_shape_is_rejected_by_daemon_caller` passing is what makes the `stream.rs` change safe to take: it demonstrates the shape being replaced was genuinely rejected by a Go caller, so this is repairing something already broken, not changing a working contract.

## Two notes, neither blocking

**1. The gated tests report `ok` in CI without running.** `require_integration!()` early-returns, so with the env var unset the suite prints `7 passed` in 0.00s having executed nothing. Same 7 with the gate on takes 0.95s and does the real work. Anyone reading CI output would reasonably conclude interop was verified — it wasn't. `#[ignore]` would report `7 ignored`, which is honest. This is pre-existing convention across tiers 04/05/07, so not this PR's fault; it matters more here only because the interop tier is this PR's headline contribution.

**2. Worth actually running it in CI.** The whole gated suite is **0.95 seconds**, and CI already installs Go and builds p2pd. There's no cost argument for leaving the gate that every subsequent phase depends on entirely manual — a job with `KWAAI_INTEGRATION_TESTS=1` would make the interop claim real. Suggest a follow-up.

Merging.

PR #81, opened 2026-08-05
feat(kwaai-p2p): in-process rust-libp2p swarm (native-p2p 2/7)

dazwin:
| # | PR | What |
|---|----|------|

---

## What this adds

A real in-process rust-libp2p swarm in `kwaai-p2p` — ping, identify and Kademlia behind a clonable handle — replacing the crate's dead prototype modules. This is the skeleton that eventually takes over the networking currently done by the Go `p2pd` child process.

## Why

Before this PR, `kwaai-p2p` had no event loop at all. The node's networking was entirely the Go p2pd child process, and the crate's `dht.rs` / `hivemind.rs` / `network.rs` / `protocol.rs` / `rpc.rs` were an unfinished prototype on a wire format that does not match hivemind's — no production code path called any of them. Rather than leave a misleading prototype alongside a real implementation, this replaces them. The handle's method names deliberately mirror `P2PClient`'s (`connect_peer` / `disconnect_peer` / `list_peers` / …) so the later cutover can swap the p2pd client for this handle without churning ~15 external call sites.

## What's in it

**New — the swarm**

- `behaviour.rs` — `KwaaiBehaviour`: ping + identify + kad(`MemoryStore`). Kademlia deliberately keeps the **default `/ipfs/kad/1.0.0`** protocol: the Python bootstraps run hivemind's go-libp2p daemon with no `ProtocolPrefix`, so a custom name would silently partition us. Identify sends `protocol_version` `"ipfs/0.1.0"` to match Go, with the kwaai identity in `agent_version`. Structured so relay/dcutr/AutoNAT/UPnP are additive fields in a later phase.
- `service.rs` — `NetworkService` owns the swarm on a dedicated tokio task (the swarm is not `Sync` and must be polled from one place). `SwarmBuilder` with TCP(nodelay) + noise + yamux + **DNS** — DNS is not optional here, the bootstraps are reached via `/dns/` addrs. A `select!` over commands, swarm events, and a 5-minute kad refresh. Dials park in `pending_dials` keyed by `ConnectionId`, DHT lookups in `pending_kad` keyed by `QueryId`; every entry is removed on **both** the success and the failure event, and `fail_all_pending` drains them on shutdown, so no caller is left awaiting a dead oneshot. No command handler awaits anything.
- `handle.rs` — `NetworkHandle`, the clonable facade. Also exposes `observed_addrs()` as `(addr, distinct observer count)`; counting distinct observers is what turns an observed address into evidence rather than one peer's opinion, which the later reachability work needs.
- `identity.rs` — load/generate on the same libp2p protobuf file format `kwaai-cli`'s `NodeIdentity` uses, so the two are on-disk compatible. New keys are Ed25519; RSA loads (the bootstraps' `Qm…` keys) via the `rsa` feature.
- `config.rs` — adds `port`, `initial_peers`, `dht_server`. `dht_server` forces kad Server mode for bootstrap-grade nodes and for loopback tests, where nothing will ever confirm an external address and an auto-mode kad would sit in Client mode answering no queries.

**Deleted — the prototype.** `dht.rs`, `hivemind.rs`, `network.rs`, `protocol.rs`, `rpc.rs` (~1,250 lines), plus `core/examples/test_announce.rs`. This is the largest single surface in the stack and the item most worth a reviewer's attention (see Risk).

**Also**

- `read_framed_futures` in `kwaai-hivemind-dht/src/wire.rs` — the `futures_io` twin of PR1's `read_framed`, for reading unary frames directly off libp2p substreams (which implement `futures::AsyncRead`, not tokio's). Needed by the unary behaviour in PR3.
- Workspace `libp2p` gains the `ping` and `dns` features.

## Risk

**No default runtime behaviour changes.** What makes that safe:

- **The deleted modules had zero consumers outside `kwaai-p2p` itself.** Verified by grepping the whole workspace at the parent commit for `KwaaiNetwork`, `kwaai_p2p::hivemind`, `::network`, `::dht`, `::rpc`, `::protocol` — no hits in any other crate or example. `kwaai-cli`, `map-server` and `kwaai-p2p-daemon` import `kwaai_p2p` only for `NetworkConfig`, `PETALS_BOOTSTRAP_SERVERS`, `P2PError` and `NodeCapabilities`, all of which are retained.
- Nothing constructs a `NetworkService` outside tests. The node still runs on p2pd.
- All existing `config.rs` fields and the `PETALS_BOOTSTRAP_SERVERS` constants are untouched, so `kwaai-cli` and `map-server` compile unchanged.

Worth looking at closely:

- The deletion itself — please confirm no out-of-tree consumer imports the removed modules.
- `02_unit_p2p.rs` loses its `ServerInfo` / hivemind-framing cases along with the prototype they tested. The real codec is `kwaai-hivemind-dht`'s (PR1), so this is not a coverage loss so much as a relocation, but it is a test deletion.
- The `pending_dials` / `pending_kad` cleanup discipline in `service.rs` is where a leak would show up as a hang rather than an error.

## Test coverage

**`core/crates/kwaai-p2p/tests/swarm.rs`** — in-process, every swarm binds an ephemeral loopback port with a freshly generated key, so nothing can collide with a node running on the same machine.

- Two swarms: dial; both sides see each other in `list_peers` with the correct direction; identify delivers an observed address to each; disconnect drops the peer from both peer lists.
- Three swarms: A knows B, B knows C, A resolves C via `dht_find_peer` — then dials C using **only** the address the DHT returned, so the test proves the lookup produced something usable rather than merely non-empty.
- The event-loop hazards from the migration plan, asserted directly: a dial to an unroutable address returns an error rather than hanging (a leaked `pending_dials` entry would surface as a timeout, not a wrong value); a malformed multiaddr does not poison the loop; commands after shutdown fail instead of hanging.

**`core/crates/kwaai-p2p/tests/live_bootstrap.rs`** (`#[ignore]`, needs internet) — connects to both public bootstraps, asserts identify completed, disconnects. Strictly read-only: ephemeral key, no announce, no DHT writes.

This live test exists to settle the migration plan's biggest unknown: the noise handshake against the bootstraps' RSA-2048 identities from rust-libp2p, a path that has never run in production because the Go daemon did the dialing. **It works.** Both bootstraps identify as agent `p2pd/0.1` advertising `/ipfs/kad/1.0.0`, `/ipfs/id/1.0.0`, `/libp2p/autonat/1.0.0`, `/libp2p/circuit/relay/0.2.0/{hop,stop}` and `/libp2p/dcutr` — confirming the kad protocol name assumed in `behaviour.rs`, and confirming that the relay/dcutr/autonat behaviours the NAT phase needs have peers to talk to.

## Stacked on

`native-p2p-pr1-wire-codec`

---

- `cargo test --workspace`: **670 passing, 0 failures.**

---


RezaRassool (approved):
Approve. The headline number (+1738/−1592, five modules deleted) looks alarming for a crate three live binaries depend on, so I verified the blast radius rather than trusting green CI.

## The deletions are dead prototype code

`dht.rs`, `hivemind.rs`, `network.rs`, `protocol.rs`, `rpc.rs` all go. `kwaai-p2p` is consumed by **kwaai-cli, map-server and kwaai-distributed** — none of which this PR touches — so the question is what they actually used. Answer:

```
kwaai_p2p::NetworkConfig
```

That is the entire live surface, in both kwaai-cli and map-server. Nothing referenced the removed modules, which matches the note in `02_unit_p2p.rs` that `kwaai_p2p::hivemind` was a prototype on a wire format that never matched hivemind's.

Full `cargo build --workspace` (including map-server and kwaai-distributed, which the default CI job does not necessarily exercise) builds clean.

## The one production-visible risk, checked

`node.rs` — the **live p2pd path** — calls `NetworkConfig::with_petals_bootstrap().bootstrap_peers` and hands the result to p2pd. A changed list would silently re-point every running node. Compared across `main` and this branch:

```
/ip4/18.219.43.67/tcp/8000/p2p/QmQhRuhee…
/ip4/52.23.252.2/tcp/8000/p2p/Qmd3A8N5aQ…
/ip4/127.0.0.1/tcp/8000/p2p/QmXwErKD4k7…
```

Byte-identical on both sides. `KWAAI_BOOTSTRAP_SERVERS_DNS` is added alongside, not substituted in, so the p2pd path is unaffected.

## Tests

21 pass: 9 unit (service, identity), 10 in-process swarm on loopback, 1 harness, plus the live one below.

`connects_and_identifies_against_live_bootstraps` is `#[ignore]`d **with an explicit reason** ("requires internet access to the live KwaaiNet bootstraps"), so it reports as `ignored` rather than silently passing. That is the right pattern, and worth contrasting with the `require_integration!()` macro in #80's tier, which prints `7 passed` in 0.00s having executed nothing — the follow-up I suggested there would be well served by copying what this PR already does.

I ran the ignored test anyway: passes against the real bootstraps in 0.81s.

## Note

`identity.rs` pins the on-disk key format with `on_disk_format_is_bare_libp2p_protobuf`, which is the test I would most want to exist here — the native path has to produce the same PeerId as the p2pd path from the same key file, or a node silently changes identity on cutover. Good that it is asserted rather than assumed.

Merging.

PR #82, opened 2026-08-05
feat(kwaai-p2p): hivemind unary RPC over the native swarm (native-p2p 3/7)

dazwin:
| # | PR | What |
|---|----|------|

---

## What this adds

Native hivemind unary RPC: a hand-rolled `NetworkBehaviour` + `ConnectionHandler` that speaks PR1's wire format directly on libp2p streams, exposed through `NetworkHandle` so a node can both call and serve hivemind handlers without the Go daemon.

## Why

Unary RPC is how hivemind does everything that matters — `DHTProtocol.rpc_{ping,store,find}`, and KwaaiNet's own `hello` / ollama-proxy / shard-proxy handlers. It is the first capability where the native stack has to be indistinguishable from p2pd on the wire, in both directions. The migration plan called for `request_response<HivemindUnaryCodec>`, but two of that behaviour's design constants are structurally wrong for hivemind, so this keeps its architecture (pending queues, handler preloading, dial-on-demand, worker futures) while fixing both. This is also the first PR where PR0's slash-less patch is exercised against a real Go peer end to end.

## What's in it

**`core/crates/kwaai-p2p/src/unary.rs` — the behaviour**

One request and one uvarint-framed `PersistentConnectionRequest` each way, then close. Two departures from `request_response`:

- **Per-request protocol.** `request_response` negotiates from one list fixed at construction and `send_request` cannot pick per request. A hivemind call *is* its protocol, so each outbound stream negotiates exactly the requested handler name — via `UnaryProtocol`, an `Arc<str>` that permits the slash-less names `StreamProtocol` refuses.
- **Dynamic inbound registration.** The advertised protocol set lives behind a shared `Arc<RwLock<…>>` read per `listen_protocol()` call, so a handler registered at runtime is negotiable on connections that already exist.

Call outcomes travel through the oneshot each request carries — dial failure, clean protocol refusal (`UnsupportedProtocol`, which the health probe depends on), timeout, remote error arm, wire failure — so no request-ID bookkeeping reaches the service loop. Inbound requests surface as `Event::InboundRequest` carrying the caller's **connection-derived** `PeerId` (`callUnary.peer` is untrusted, per PR1's finding) plus a responder oneshot; a dropped or timed-out responder becomes an error frame, never a hang.

**Service and handle integration** (`behaviour.rs`, `service.rs`, `handle.rs`)

- `KwaaiBehaviour` gains a `unary` field, built from `NetworkConfig`'s `request_timeout`. `max_concurrent_streams` keeps unary's own default, being a per-connection resource guard rather than anything config expresses.
- `Command::{CallUnary, AddUnaryHandler, RemoveUnaryHandler}`. `CallUnary` needs no pending map — `send_request` owns the oneshot and resolves it on every path.
- The service keeps a protocol → unbounded-sender dispatch map in lockstep with the behaviour's inbound protocol set. `InboundRequest` is routed by the *negotiated* protocol; a dead handler task deregisters the protocol so later calls get a clean refusal rather than a black hole.
- `NetworkHandle::{call_unary_handler, add_unary_handler, add_unary_handler_boxed, remove_unary_handler}`, named after their `P2PClient` counterparts. `add_unary_handler` spawns one dispatch task per protocol and one task per call, so a slow handler delays only its own caller. The `UnaryError` → `P2PError` mapping is documented on `call_unary_handler`.

**Three correctness fixes made during review of the above** — each worth reading, since they are the concurrency hazards that would otherwise surface only under production load:

1. **Outbound streams correlated by protocol, not position.** `requested_outbound` was a FIFO matched by position, on the assumption that `FullyNegotiatedOutbound` arrives in request order. It does not — each negotiation is an independent round trip — so two concurrent calls to *different* protocols on one connection could deliver one caller the other's response. This is exactly the deployment shape where `shard serve` and `storage serve` each serve a different protocol through one node.
2. **Then correlated by exact request id.** Protocol matching fixed the successes but left `DialUpgradeError` positional, so a refusal arriving before a success could still misattribute. Correlation now rides `OutboundOpenInfo` (a `u64` the swarm echoes back verbatim on *both* the success and failure events), which is exact on every path. Same commit: worker pools split into inbound vs outbound (both bounded by `max_concurrent_streams`) so an outbound burst cannot starve inbound serving; inbound overflow answers with the "node at capacity" error arm rather than resetting an already-negotiated stream; and dropping the handler resolves every queued request with a definite `Wire("connection closed")` instead of a silently dropped channel.
3. **Waking the throttle when a negotiation failure frees a slot.** The emission throttle counted `requested_outbound` in its gate, but nothing re-entered `poll` when `DialUpgradeError` shrank that map — worker completion wakes via `FuturesUnordered`, map mutation does not. With the cap saturated, a routine negotiation refusal freed the last slot silently and the next queued call hung forever. The handler now stores the poll waker and wakes it whenever a slot frees outside the worker pools. Refusal-frame writers also move to their own pool, since they previously counted against the same allowance that admits them.

## Risk

**No default runtime behaviour changes.** Nothing calls `call_unary_handler` or `add_unary_handler` outside tests at this commit — the node still runs on p2pd, and the `native_p2p` flag that selects this path does not exist until PR6. All changes are additive to `kwaai-p2p`; no existing module's behaviour is altered.

Worth looking at closely:

- `unary.rs`'s `ConnectionHandler::poll` — the waker discipline in fix (3) and the split worker pools in fix (2). These are the paths where a mistake shows up as a hang rather than an error, and the three fixes above are all in this one function's neighbourhood.
- The service's dispatch map staying in lockstep with the behaviour's protocol set; a divergence would surface as a black hole rather than a refusal.

## Test coverage

**In-process, `kwaai-p2p/tests/unary.rs`** — round trip via dial-on-demand, error arm, refusal-then-reuse of the connection, dropped responder, unreachable peer, 8 concurrent calls, late registration reaching an existing connection. Plus targeted regressions for each fix above: two concurrent calls on *different* protocols (fix 1 — the original test used one protocol and missed it), an outbound burst through a single-slot cap all completing, the at-capacity refusal arriving as a clean error arm while the occupied slot completes after release, and a queued call behind a refused one through a single-slot cap (fix 3).

**In-process, `kwaai-p2p/tests/service_unary.rs`** (9 tests) — echo round trip, dial-on-demand, remote error arm, unserved-protocol refusal, unreachable peer, removal and re-registration, per-call concurrency, post-shutdown behaviour.

**Tier 08 — `08_unary_swarm_interop.rs`** (3 tests): `kwaai_p2p::unary::Behaviour` in a real swarm against a real go-libp2p daemon — the first time the native stack negotiates TCP + noise + yamux + slash-less multistream-select against Go end to end. Swarm calls a daemon-served handler (proves PR0's patch dialer-side against Go's listener); daemon calls a swarm-served handler (listener-side, plus the responder frame end to end); an unserved protocol comes back as a clean `UnsupportedProtocol` refusal with the connection surviving for a follow-up call.

**Tier 09 — `09_service_unary_interop.rs`** (4 tests): raises the left-hand side to what production will run — a full `NetworkService` behind a `NetworkHandle`, exchanging unary calls with a real p2pd both ways. Over and above tier 08 this covers the handle's `UnaryError` → `P2PError` mapping, the service's inbound dispatch map, and the fact that identify/kad sharing the connection does not disturb unary negotiation against go-libp2p. Includes `remove_unary_handler` being visible to a foreign stack — the deregistration path PR5 depends on.

Tiers 08/09 are gated behind `KWAAI_INTEGRATION_TESTS=1` and require a real p2pd binary; daemons are owned by `TestNode` and killed by PID on drop.

## Stacked on

`native-p2p-pr2-swarm`

---

- `cargo test --workspace`: **697 passing, 0 failures.**

---

### DIFF
--- a/core/Cargo.lock
+++ b/core/Cargo.lock
@@ -3123,6 +3123,7 @@ dependencies = [
  "async-trait",
  "bincode",
  "futures",
+ "kwaai-hivemind-dht",
  "libp2p",
  "serde",
  "serde_json",
@@ -3132,6 +3133,7 @@ dependencies = [
  "tokio-test",
  "tracing",
  "tracing-subscriber",
+ "uuid",
 ]
 
 [[package]]
--- a/core/crates/kwaai-network-tests/tests/08_unary_swarm_interop.rs
+++ b/core/crates/kwaai-network-tests/tests/08_unary_swarm_interop.rs
@@ -1,0 +1,249 @@
+//! `unary::Behaviour` ↔ real p2pd interop — the Phase 2 swarm gate.
+//!
+//! Where `07_wire_interop` proved the *frame bytes* against a Go daemon over
+//! p2pd-managed streams, this tier proves the whole native stack: a rust-libp2p
+//! swarm running `kwaai_p2p::unary::Behaviour` negotiates TCP + noise + yamux +
+//! slash-less multistream-select against go-libp2p and exchanges hivemind
+//! unary calls in both directions.
+//!
+//! This is also the live check on the vendored `multistream-select` patch
+//! (`core/patches/multistream-select`): a Go peer must accept our bare-name
+//! proposal, and our listener must parse Go's.
+//!
+//! | test | caller | responder |
+//! | --- | --- | --- |
+//! | `swarm_calls_daemon_handler` | **swarm** | p2pd |
+//! | `daemon_calls_swarm_handler` | p2pd | **swarm** |
+//! | `swarm_gets_clean_refusal_from_daemon` | **swarm** | p2pd (no handler) |
+//!
+//! Gate: `KWAAI_INTEGRATION_TESTS=1`, like the other integration tiers.
+
+use std::time::Duration;
+
+use kwaai_network_tests::{harness::TestNode, metrics::MetricsRecorder, require_integration};
+use kwaai_p2p::unary::{self, UnaryError, UnaryProtocol, UnaryResult};
+use libp2p::futures::StreamExt;
+use libp2p::{swarm::SwarmEvent, Multiaddr, PeerId, Swarm, SwarmBuilder};
+use tokio::sync::{mpsc, oneshot};
+
+const PROTO: &str = "DHTProtocol.rpc_ping";
+const CALL_TIMEOUT: Duration = Duration::from_secs(20);
+
+// ============================================================================
+// Swarm harness
+// ============================================================================
+
+/// A driven swarm node: commands in via channels, its event loop on a task.
+struct SwarmNode {
+    peer_id: PeerId,
+    /// Listen address including `/p2p/<id>`, dialable by a p2pd.
+    addr: String,
+    calls: mpsc::UnboundedSender<(PeerId, String, Vec<u8>, oneshot::Sender<UnaryResult>)>,
+}
+
+impl SwarmNode {
+    /// Spawn a swarm serving `PROTO` with `handler`, and knowing `known_peer`'s
+    /// address for dial-on-demand.
+    async fn spawn(
+        known_peer: Option<(PeerId, Multiaddr)>,
+        handler: impl Fn(Vec<u8>) -> Result<Vec<u8>, String> + Send + 'static,
+    ) -> Self {
+        let mut swarm: Swarm<unary::Behaviour> = SwarmBuilder::with_new_identity()
+            .with_tokio()
+            .with_tcp(
+                libp2p::tcp::Config::default().nodelay(true),
+                libp2p::noise::Config::new,
+                libp2p::yamux::Config::default,
+            )
+            .expect("tcp transport")
+            .with_behaviour(|_| unary::Behaviour::new(unary::Config::default()))
+            .expect("behaviour")
+            .with_swarm_config(|c| c.with_idle_connection_timeout(Duration::from_secs(60)))
+            .build();
+
+        swarm
+            .behaviour_mut()
+            .register_protocol(UnaryProtocol::new(PROTO));
+        if let Some((peer, addr)) = known_peer {
+            swarm.add_peer_address(peer, addr);
+        }
+
+        let peer_id = *swarm.local_peer_id();
+        swarm
+            .listen_on("/ip4/127.0.0.1/tcp/0".parse().expect("valid multiaddr"))
+            .expect("listen");
+        let listen_addr = loop {
+            if let SwarmEvent::NewListenAddr { address, .. } = swarm.select_next_some().await {
+                break address;
+            }
+        };
+
+        let (calls_tx, mut calls_rx) =
+            mpsc::unbounded_channel::<(PeerId, String, Vec<u8>, oneshot::Sender<UnaryResult>)>();
+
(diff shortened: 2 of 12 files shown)


RezaRassool (approved):
Approve. 849 lines of behaviour against **1568 lines of tests** — and the tests are the good kind, covering refusal paths and handler lifecycle rather than just the happy round trip.

## Backpressure is the part I most wanted to see, and it is layered

A custom `NetworkBehaviour` accepting inbound streams from arbitrary peers is where this stack could most easily grow a memory-exhaustion bug. Instead:

- `max_concurrent_streams` bounds inbound workers
- past that, a **polite refusal** with its own bounded allowance (`REFUSAL_SLOTS`) — reads the request for its `callId`, answers with the error arm, closes
- past the allowance, drops outright
- the inbound channel is zero-capacity, so a worker parks until the handler actually takes the request

The refusal tier is the thoughtful bit: it converts "at capacity" from a caller-side hang into an immediate error, so a loaded responder does not stall every peer for their full timeout. `swarm_gets_clean_refusal_from_daemon` covers it.

Outbound is bounded too — `request_timeout` defaults to 30s, so a hung remote cannot leak a pending entry forever.

## The documented trap is handled

> The caller, derived from the connection — never from the frame's `peer`

That is exactly right, and it is the one thing Phase 0 flagged as a security-relevant footgun: on a raw stream the `callUnary.peer` field arrives as the *caller* sent it, so trusting it would let a peer claim any identity. Good that it is asserted in a comment at the point of use rather than left implicit.

## Verification

Unit and integration: **40 pass** across `unary.rs`, `service_unary.rs` and the swarm suites.

The interop tiers are gated behind `KWAAI_INTEGRATION_TESTS=1`, so CI runs none of them. Ran locally against real p2pd:

```
08_unary_swarm_interop
  swarm_calls_daemon_handler ............... ok
  daemon_calls_swarm_handler ............... ok
  swarm_gets_clean_refusal_from_daemon ..... ok

09_service_unary_interop
  service_calls_daemon_handler ............. ok
  daemon_calls_service_handler ............. ok
  removing_a_handler_makes_the_daemon_call_fail  ok
  service_maps_daemon_refusal_to_protocol_error  ok

7 passed, 0 failed
```

Both directions against a real Go daemon, plus refusal and handler-removal semantics. That is the interop claim actually demonstrated rather than asserted.

## One characteristic worth recording, not a defect

The protocol-per-request design here — a hivemind call *is* its protocol, so each call negotiates a fresh stream — costs roughly **one extra network round trip per call** versus p2pd's persistent pre-negotiated connection. Measured on the live fleet against a peer where both stacks held a *direct* connection: native 397ms vs p2pd 208ms, a 1.91x ratio, and halving each native figure recovers a plausible direct RTT.

It is inherent to the wire shape, not a flaw in this implementation, and native's much better path selection more than compensates on nearby peers. It is invisible on multi-second inference calls and visible on chatty low-latency RPC. Worth knowing before anyone benchmarks the cutover and is surprised. Amortising it would mean caching negotiated streams per (peer, protocol) — a design change, not a tweak, and not something to hold this PR for.

Merging.

PR #83, opened 2026-08-05
feat(hivemind-dht): serve the hivemind DHT natively (native-p2p 4/7)

dazwin:
| # | PR | What |
|---|----|------|

---

## What this adds

A bootstrap-grade native implementation of hivemind's DHT — storage with real subkey/dictionary semantics, XOR-distance nearest-peer selection, and a service that serves `DHTProtocol.rpc_{ping,store,find}` over the native swarm — plus the record builders that let a node announce itself without the Go daemon.

## Why

This is the capability that makes a native node a real participant rather than a client: it can hold records for other peers, answer lookups, and route iterative queries. It also fixes the single most consequential bug in the existing storage layer — `handle_store` ignored `request.subkeys` entirely, so N petals servers announcing under a shared `{prefix}.{block}` key overwrote each other and a block ended up with exactly one visible server. The semantics are ported against a pinned hivemind revision (`213bff98a62accb91f254e2afdccbf1d69ebdea9`, the revision petals depends on) with each rule cited to the Python source, because several of them are counter-intuitive and one contradicts hivemind's own proto comments.

## What's in it

**`kwaai-hivemind-dht/src/storage.rs` (new) — the pure storage model**

- `Stored::{Regular, Dictionary(BTreeMap<subkey, (value, expiration)>)}`, mirroring `DHTLocalStorage` (`dht/storage.py:35-69`). This is the subkey fix.
- Subkey sentinels (`dht/protocol.py:33`): `IS_REGULAR_VALUE` is msgpack nil (`0xC0`); `IS_DICTIONARY` is EMPTY bytes and means "the value is a whole serialized `DictionaryDHTValue`" — the opposite of the natural reading, and of the misleading comment at `proto/dht.proto:30`. Any other tag is a real subkey, kept opaque rather than decoded.
- `DictionaryDHTValue` as msgpack **Ext 0x50** (`storage.py:10,20-23`), verified byte-for-byte against Python msgpack output. Ext 0x40 is the *tuple* code (`utils/serializer.py:27`) used by petals `ServerInfo` values and passes through untouched.
- Strictly-greater freshness on absolute expiration (`timed_storage.py:80`), applied per-subkey inside dictionaries, so an equal expiration is rejected; expired stores refused outright (`:75-76`).
- Regular→dictionary promotion and outer-expiration refresh to the max of entries (`storage.py:51-69`).
- Two storage tiers selected by `in_cache` (`protocol.py:242`); find prefers whichever holds the later expiration (`protocol.py:342-347`).
- `FOUND_DICTIONARY` results carry the Ext(0x50) serialization and the **outer** entry expiration (`protocol.py:355`) — `proto/dht.proto:54` claims `latest_expiration_time`, which the code contradicts; the code wins.
- Nearest peers by XOR distance over 20-byte big-endian DHTIDs, k=20, excluding the requester (`protocol.py:362-364`, `routing.py:108-128`). Previously an insertion-ordered `take(20)` that ignored the key entirely.
- Bootstrap-grade extras: the cache tier is capacity-bounded with earliest-expiration eviction (`timed_storage.py:60-68`), and `cleanup_expired` sweeps both tiers so a key nobody queries is not pinned forever.

`server.rs` becomes the request/response layer over this.

**`NodeInfo::from_peer_id` emits a 20-byte DHTID.** `NodeInfo.node_id` is documented in `dht.proto` as `DHTID.to_bytes()` — a fixed 20-byte big-endian SHA1 (`routing.py:252,288-290`) derived as `SHA1(peer_id.to_bytes())`. We were sending the raw 34-38 byte multihash. Added as a reusable `protocol::dht_id_from_peer_id`; `map-server/src/crawler.rs:74-77` already derived this correctly for its own ID, so the two implementations now agree. **This one does change bytes on the default p2pd path — see Risk.**

**Protocol ID constants are bare, not slashed.** Hivemind derives IDs as `{ClassName}.rpc_{method}` with no leading slash, proven by PR1's `slashless_protocol_negotiates`. The slashed spellings would fail to negotiate against every Python peer. The only consumer was `client.rs`, a superseded `request_response` prototype with zero workspace consumers, which now spells its slashed names inline with a doc comment saying it cannot negotiate against a real peer.

**`kwaai-p2p/src/dht_service.rs` (new) — connecting storage to the swarm**

- `spawn_dht_service()` registers `DHTProtocol.rpc_{ping,store,find}` as unary handlers, each decoding a protobuf request, applying the matching `DHTStorage` method, and encoding the response.
- A maintenance task pulls kad's k-buckets every 60 s into `update_peer_ids()` and runs `cleanup_expired()`. Nothing called either before; without the maintenance half, storage would answer every `rpc_find` with an empty neighbour list (breaking iterative lookups through us) and pin every record forever.
- `NetworkHandle::routing_peers()` reads the kad routing table by walking in-memory k-buckets — no network I/O, so it is safe inside the event loop.
- `rpc_ping` answers with our `NodeInfo` and a live `dht_time` but `available=false`: hivemind only sets true after dialing the caller back. Reporting false is the safe direction — the field feeds the *caller's* confidence in its own reachability, not whether our records are stored or found.

**`kwaai-cli/src/announce.rs` (new) — record builders**

- `VpkInfo`, `DHTServerInfo`, `ModelInfo` and `dht_id()` extracted **verbatim** from `node.rs`. Their msgpack encodings are a published wire format that `map.kwaai.ai`'s crawler and Python hivemind both decode, so they move byte-for-byte; `node.rs` uses them from the new module and its p2pd delivery path is otherwise untouched.
- `build_announce_records()` / `build_unannounce_records()` produce exactly today's four record kinds (`{prefix}.{block}`, `_petals.models`, `_kwaai.inference.nodes`, `_kwaai.vpk.nodes`), all 360 s TTL, `in_cache` false, subkey `msgpack(peer_b58)` — a real subkey, never a sentinel.
- `send_records_via_handle()` pushes them through `NetworkHandle` rather than the p2pd control socket, keeping the `(any_success, timings)` shape the reputation store is fed from.
- Corrects the unannounce docstring, which claimed the tombstone uses "already-expired timestamps". It does not and must not: hivemind rejects both already-expired stores and any expiration not strictly greater than the record it replaces, so an expired tombstone would leave the live `state=2` record standing. The tombstone carries a normal now+360 s expiration and works through `state=-1`.

`run_node()` still calls its existing p2pd functions; the rewire is PR6.

## Risk

**No behavioural change to the node's runtime path — with one narrow, deliberate exception a reviewer should confirm.**

- `DHTStorage`, `dht_service` and the new `announce.rs` builders have no callers outside tests at this commit; `run_node` still uses its existing p2pd announce path.
- The extracted record types moved verbatim, so the announced msgpack bytes are unchanged.
- **The exception: `NodeInfo::from_peer_id` is called on the live p2pd announce and unannounce paths** (`node.rs:1255` inside `announce()`, `node.rs:1179` inside `unannounce()`), so after this PR the running node emits a 20-byte SHA1 `node_id` where it previously emitted its raw multihash. Verified by inspection at this commit, and by a daemon run with `native_p2p` off reaching those functions. This is a correctness fix — the previous value was malformed against the documented schema. It is believed harmless in practice because bootstraps only feed this field into routing-table scheduling (`rpc_store`/`rpc_find` kick off an async `update_routing_table` via `rpc_ping`, `protocol.py:232-234,337-338`); a wrong-length ID lands in an arbitrary k-bucket and is never used to address us, and nothing routes *to* us by `node_id` today. **This is the claim most worth a second opinion**, since it is the only place in PRs 0-5 where the default path emits different bytes.
- Three tests in `01_unit_dht.rs` encoded the old broken behaviour and are consciously updated: `storage_store_and_find` and `storage_multi_key_batch` passed an *empty* subkey tag (which is `IS_DICTIONARY`) and only passed because subkeys were ignored — they now use `IS_REGULAR_VALUE`. `node_info_peer_id_roundtrip` asserted a `PeerId` round trip that only held while `from_peer_id` emitted the raw multihash; it is renamed `node_info_carries_a_20_byte_dht_id` and asserts the correct property.
- **Out of scope, tracked as follow-up:** record validators and signatures (`_validate_record` / `_validate_dictionary` are absent here). The module docs say so explicitly. A reviewer should confirm they are comfortable deferring this for a serving node.

## Test coverage

**Unit (`storage.rs`, `server.rs`)** — golden-byte coverage of both sentinels and the Ext(0x50) layout (cross-checked against Python msgpack *and* against the decoder in `map-server/src/crawler.rs:378-413`, which reads live bootstrap traffic), subkey accumulation, per-subkey freshness, promotion, whole-dictionary merge with `all()` semantics, tier preference, XOR ordering with hand-computed distances, requester exclusion, and NOT_FOUND still carrying nearest peers. The storage semantics were additionally differential-tested against a faithful Python transcription of `TimedStorage`/`DictionaryDHTValue`/`DHTLocalStorage`.

**Unit (`protocol.rs`)** — 20-byte width, SHA1 equality, and a regression guard against re-emitting the raw multihash. Plus a test in `announce.rs` storing an announce then a tombstone into a real `DHTStorage` to pin the `state=-1` tombstone behaviour.

**In-process (`kwaai-p2p/tests/dht_service.rs`)** — two `NetworkService`s, one serving: a subkeyed store comes back as rt=2 `FOUND_DICTIONARY` with the subkey intact, NOT_FOUND still carries neighbours, ping round trips, the routing snapshot reaches served neighbour lists, removal surfaces as a negotiation refusal, and garbage gets the handler's error arm rather than a hang.

**Tier 10 — `10_native_dht_interop.rs`** — a **Go daemon** stores to our native DHT node. The in-process test above cannot catch a foreign-stack disagreement, and this is exactly the direction PR1's `stream.rs` bug used to break. Three checks: a subkeyed record stored by the daemon comes back as rt=2 `FOUND_DICTIONARY` with subkey and value intact (the exact shape every petals block announcement uses); `rpc_ping` round trips, covering the previously-broken inbound path in isolation; an unheld key answers NOT_FOUND with index-aligned nearest peers, so a Go caller's iterative lookup can route through us. Note the p2pd is a plain unary caller and never parses the DHT payloads, so this is a wire-and-dispatch proof rather than a claim about Go's own DHT client semantics.

**Live (`kwaai-p2p/tests/live_dht_announce.rs`)** — gated **twice**, `#[ignore]` *and* `KWAAI_LIVE_DHT_WRITE=1`, because it writes to the real DHT (the `live_bootstrap.rs` precedent is that `--ignored` alone means read-only). Safe against production: ephemeral Ed25519 identity, a per-run key under `_kwaai.test.ephemeral.{peer}` the health-map crawler provably cannot reach, a plain msgpack value deliberately not Ext(64) so even a hypothetical crawl would reject it, and a 60 s TTL so records self-clean.

Ran green against bootstrap-1/2.kwaai.ai:

- The RSA dial path works — both Python bootstraps (RSA-2048) complete the noise handshake from rust-libp2p in ~200 ms. The biggest Phase 1 unknown, closed empirically.
- A Python bootstrap accepts our native `rpc_store`: `store_ok=[true]`, ~170 ms.
- The subkeyed record reads back as rt=2 `FOUND_DICTIONARY` with subkey and value byte-identical, ~167 ms.
- A tombstone with a strictly-greater expiration replaces it. The key stays FOUND — a store can never delete — so the assertion is that the value changed, not that the key vanished.

**Two findings that contradict prior assumptions**, recorded in the migration doc:

- `rpc_store` is **single-hop**. A record stored on bootstrap #1 is not visible on bootstrap #2; hivemind's `DHTNode.store` gets redundancy by calling `rpc_store` on each of the k nearest nodes from the client side. Today's fan-out to every bootstrap is therefore load-bearing, not belt-and-braces.
- The live bootstraps answer NOT_FOUND with an **empty** nearest-peers list, though hivemind's `rpc_find` attaches them and our `DHTStorage` does too. A bootstrap will not route an iterative lookup for us.

## Stacked on

`native-p2p-pr3-unary`

---

- `cargo test --workspace`: **758 passing, 0 failures.**

---


RezaRassool (approved):
Approve. The biggest thing in the stack so far (+4566/−474), and it touches `kwaai-cli/src/announce.rs` — live code on today's p2pd path — so that is where I looked first.

## The live path is untouched, by construction

`announce.rs` has **zero lines removed**. Every addition is new functions carrying `#[allow(dead_code)] // used from node_native in the NAT-traversal layer (pr7)`. The running fleet's announce behaviour cannot change from this merge, because none of the new code is reachable until pr7 wires it up. That is the right shape for a change this size and it made the review much easier.

## The two tracked gaps, checked against what actually landed

**Record validators — absent, and honestly so.** Hivemind gates writes behind `RecordValidator`; this port does not. The code says so in the module docs rather than leaving it implicit, and ties the capacity bound to it: *"there for why it must not be unbounded while validators are missing."* Acknowledging a gap at the point where it matters is the right call.

**Storage is bounded, so there is no growth vector.** `DhtServer` constructs `LocalStorage::with_maxsize(DEFAULT_STORAGE_SIZE)`; the unbounded `LocalStorage::new()` appears only in tests. Combined with `MAX_STORE_KEYS_PER_REQUEST = 1024`, an unauthenticated writer cannot grow memory without limit. That satisfies the "capacity-bounded as an interim abuse guard" condition the migration doc set.

**But the O(n) eviction is worse than "add a heap eventually" suggests.** `enforce_capacity` picks its victim with `.iter().min_by(…)` while holding the write lock, and `DEFAULT_STORAGE_SIZE` is **1,048,576**. Below capacity this never runs and costs nothing. At capacity, every insert triggers a ~1M-entry scan under the write lock, which serialises the whole DHT service — so the behaviour is not "gradually slower", it is fine and then a cliff, exactly when the node is busiest.

That is not a blocker here: a non-bootstrap node will not fill a million entries, which is the deployment this PR targets. It does mean the tracked follow-up should be treated as **a hard prerequisite for running a Rust bootstrap**, alongside validators, rather than a nice-to-have. Worth stating in the follow-up so nobody reads "O(n) eviction" as merely untidy.

## Verification

**114 tests pass** across `kwaai-hivemind-dht` and `kwaai-p2p`, plus 24 in tier 01.

Tier 10 is gated behind `KWAAI_INTEGRATION_TESTS=1`, so CI runs none of it. Locally, against a real p2pd:

```
daemon_pings_the_native_dht_node ......... ok
daemon_stores_and_finds_a_subkeyed_record  ok
daemon_gets_not_found_with_nearest_peers . ok
```

The middle one is the one that matters — subkeyed records are where the dictionary/DHTExpiration semantics have to match Python exactly, and it is proven against the real daemon rather than against our own encoder.

CI green: Clippy, Format, Test (ubuntu), WASM, plan.

Merging.

PR #84, opened 2026-08-05
feat(p2p-daemon): node-hosted p2pd control socket + pipe mode (native-p2p 5/7)

dazwin:
| # | PR | What |
|---|----|------|

---

## What this adds

A `ControlServer` that serves the go-libp2p-daemon control protocol from the node itself, so every existing client — the map-server crawler, `shard serve`, `storage serve`, rag, `kwaainet p2p`/`status`, the GUI — can attach to the same socket and act as the node's peer identity without a p2pd process behind it. Includes both unary RPC and raw pipe mode.

## Why

The whole point of the native migration is that external processes should not notice it. Roughly 15 call sites across the workspace speak the p2pd control protocol through `client.rs` / `persistent.rs` / `dht.rs`, and rewriting them all would turn a dark launch into a flag day. Serving the identical protobuf protocol from the node means those files are unchanged and the cutover is a config flag. This PR also fixes a real outage class on the way through: a crashed handler client currently leaves the daemon advertising a protocol nothing will answer, turning what should be a clean negotiation refusal into a hang for every remote caller.

## What's in it

**`kwaai-p2p-daemon/src/server.rs` (new, ~1,730 lines) — the control server**

Slice 1 (unary): IDENTIFY, CONNECT, DISCONNECT, LIST_PEERS, DHT FIND_PEER, and the full persistent-connection unary sub-protocol (`add`/`remove_unary_handler`, `call_unary_handler`, bidirectional callId correlation). Refused with the Go daemon's own "not supported" shape: the remaining DHT verbs, PUBSUB and CONNMANAGER.

- Response shapes follow the **Go source, not the proto file**, because several behaviours are load-bearing but invisible in the schema — notably that an add/remove-handler ACK is a `PersistentConnectionResponse` with a callId and *no* message arm, which is exactly what `persistent.rs` decodes as success.
- **Handler ownership is per socket connection and released on disconnect for any reason, including a client crash.** This is the stale-handler fix. A deliberate divergence from Go, whose handler map is process-global and outlives the client.
- Call-ID correlation state is per connection, never global, so two independent clients cannot collide on a UUID — a failure mode that would be silent and would deliver one process another's response. Inbound dispatch generates its own call IDs for the same reason.

Slice 2 (pipe mode): STREAM_OPEN / STREAM_HANDLER / REMOVE_STREAM_HANDLER, relaying raw bytes between a socket client and a libp2p stream in both directions — what inference-mux and `node.rs`'s `rpc_*` forwarding need.

- **The fd handoff.** A successful STREAM_OPEN is terminal for the connection, exactly as PERSISTENT_CONN_UPGRADE is: Go writes the response, calls `doStreamPipe(c, s)`, and returns (`conn.go:59-73`). Slice 1 split each socket at accept and boxed the write half behind `Arc<Mutex<Box<dyn AsyncWrite>>>` — and a boxed trait object cannot be rejoined, so the two halves could never become one duplex stream again. Fixed by keeping the socket's *concrete* type: a `ClientSocket` enum over Unix/TCP that splits inside `ConnState` and reunites on the way into pipe mode. The writer slot becomes an `Option` that pipe mode takes; a late frame write then finds `None` and errors, which is correct — the socket is no longer a frame channel. Framing, persistent mode and every existing response shape are untouched.
- **Backpressure** is `tokio::io::copy_bidirectional` in both directions, whose flow control awaits each write before reading more, so a slow consumer stops the producer at the socket instead of accumulating in this process. libp2p's yamux window applies on top. Half-closes propagate; either side ending drops both socket and stream, so no task or fd outlives the relay.
- **STREAM_HANDLER** dials the client's listener per inbound stream, writes the length-delimited `StreamInfo` prologue every consumer already parses (`stream.rs::parse_stream_info`), then relays. A dial-back failure resets the inbound stream, matching Go's `handleStream`, so the remote fails at once rather than waiting out its own timeout. Handler addresses are validated as dialable TCP at registration rather than per stream.

**`kwaai-p2p/src/raw_stream.rs` (new) — raw libp2p streams**

A sibling to `unary::Behaviour` that reuses `unary::Protocols` and `UnaryProtocol` for the upgrade and then hands the negotiated `Stream` out untouched. No framing, no per-call timeout, no callId correlation, no worker futures — a raw stream may live for a node's whole session (inference-mux does), which is precisely the lifetime model the unary path is built to prevent. Folding a "raw mode" flag into `unary::Handler` would have put two unrelated lifetime policies behind one set of queues. `libp2p_stream` remains out: `StreamProtocol::new` refuses slash-less names, for the same reason unary needed a hand-rolled behaviour.

- Outbound `open_stream` dials on demand and correlates upgrades by id on both success and failure paths, matching the fix already made in `unary::Handler`. Inbound protocols live in a shared set read per `listen_protocol()`, so a registration reaches connections established earlier. Every path resolves the caller's oneshot.
- `NetworkHandle::{open_raw_stream, accept_streams, remove_stream_handler}`, named after their `P2PClient` counterparts. The service keeps a `stream_handlers` dispatch map separate from the unary namespace, and **refuses a second registration rather than rebinding** — a raw handler is an accept loop owned by one process, and silently rebinding would strand the first owner with no way to notice.

**One fix during review:** `deregister_all` held the global `handler_owners` mutex across sequential `remove_unary_handler` awaits. `NetworkHandle` calls have no timeout, so one dead client's cleanup behind a busy swarm blocked every other client's handler registration — the stale-handler outage class this code exists to fix, relocated to all clients at once. Ownership is still decided under the lock; the swarm calls now happen after it drops.

## Risk

**No default runtime behaviour changes.** `ControlServer` is **not wired into `run_node`** — that is PR6. Nothing constructs one outside tests, and the p2pd child process still serves the node's control socket. `raw_stream` is a new module with no non-test consumers here.

The client side is untouched by design: `client.rs`, `persistent.rs` and `dht.rs` and all their call sites are unchanged, which is what makes the byte-compatibility claim testable rather than aspirational.

Worth looking at closely:

- **The `ClientSocket` restructure** is the one place this PR modifies existing connection-handling code rather than adding alongside it. The Unix/TCP enum and the `Option` writer slot touch the framing path that persistent mode also uses; the assertion that "framing, persistent mode and every existing response shape are untouched" is the thing to verify.
- **Handler ownership divergence from Go** is intentional but is a behaviour difference a reviewer should agree with, not just notice.
- **Windows TCP parity for pipe mode is recorded as untested.** The `ClientSocket` TCP arm exists and the relay is transport-agnostic, but every test binds a unix socket.

## Test coverage

**In-process, `kwaai-p2p-daemon/tests/control_server.rs`** — 11 conformance tests for slice 1 driving the **real, unmodified `P2PClient`** against a `ControlServer` over a real `NetworkService`, including handler deregistration and call-ID isolation asserted from a *remote* caller (the only place registration is observable). Plus 8 for pipe mode: end-to-end relay socket → libp2p → socket, 2 MiB each way with a deliberately slow reader (two `copy_bidirectional` hops in the path), caller half-close observed as EOF with the reply direction still live, a failed open leaving the connection in framing mode, handler release on client disconnect, ownership on removal, and dial-back reset.

**In-process, `kwaai-p2p/tests/raw_stream.rs`** — slash-less negotiation, protocol preference lists, 4 MiB in each direction concurrently (backpressure: a buffering or truncating implementation fails here rather than in production), half-close observed as EOF with the reverse direction still writable, concurrent streams on one connection, dial-on-demand, and clean refusal for unserved or removed protocols. The unary path is untouched and its tests are unchanged.

**Tier 11 — `11_control_server_interop.rs`** — attaches real `P2PClient`s to *both* implementations and drives unary calls across them, holding two interfaces at once: the IPC protocol (the same client code cannot tell a Go daemon from a native node) and the libp2p wire (a handler registered over one side's socket answers a caller that arrived through the other's). The matrix: (1) client via native ControlServer → handler on p2pd; (2) client via p2pd → handler on native ControlServer; (3) native → native; (4) p2pd → p2pd, already covered by tier 07 and not repeated. Cells 1 and 2 are the mixed-fleet cutover window; cell 3 is the post-cutover steady state. Plus IDENTIFY parity in the fields clients actually parse (raw peer-ID bytes, binary multiaddrs — what `node.rs` self-discovery reads), and the stale-handler fix observed from the *Go* stack, where the assertion is that the daemon gets a refusal rather than a timeout.

**Tier 12 — `12_pipe_mode_interop.rs`** — the same matrix for pipe mode, which has no framing of its own to check (the whole contract is "the bytes that go in come out"), so each cell opens a raw stream across the implementation boundary and asserts payload fidelity both ways. Both mixed-fleet directions: a native node's client streaming to a p2pd's registered stream handler (a migrated node's inference-mux reaching a GPU node that has not migrated), and the reverse, which exercises the whole inbound path (libp2p → `raw_stream::Behaviour` → service dispatch map → dial-back → `StreamInfo` prologue → relay). Two checks beyond round trips: **`StreamInfo` prologue parity**, because `inference_mux.rs` consumes the prologue *blind* (reads the varint, discards the body) so a framing divergence would desynchronise its frame loop rather than fail loudly — both prologues are read through the consumers' own `parse_stream_info`; and **1 MiB each way** across the boundary, the only place Go's two `io.Copy` goroutines meet `copy_bidirectional` under load, with the far end reading in small chunks so a buffering or truncating regression cannot pass.

Tiers 11/12 are gated behind `KWAAI_INTEGRATION_TESTS=1`. All 23 integration tests across tiers 07/09/10/11/12 reported green against a real p2pd.

## Stacked on

`native-p2p-pr4-dht`

---

- `cargo test --workspace`: **798 passing, 0 failures.**

---


RezaRassool (approved):
Approve. +5471/−15 and the deletions are trivial (1–6 lines, mostly re-exports), so this is additive: 2845 lines of tests against roughly 2400 of implementation, and the `daemon.rs` p2pd-spawn path is untouched.

## I closed the documented gap rather than repeating it

The migration doc says of this tier:

> **Windows TCP parity for pipe mode is untested.** … every test in this tier binds a unix socket. The Windows client path (`/ip4/127.0.0.1/tcp/5005`) needs a run on Windows CI before the cutover, or at minimum a TCP-bound variant of the `control_server.rs` pipe tests.

Still accurate as written — `TestNode::spawn()` binds `/unix/…` exclusively, so the TCP arm has never been executed by any test, on any platform. Interestingly `local_addr()` already carries the comment *"Tests bind `/ip4/127.0.0.1/tcp/0` and need the real port back"*, so the seam was built for exactly this and then not used.

So I wrote the missing variant and ran it. The **unmodified `P2PClient`** against a TCP-bound `ControlServer`:

```
TCP control socket OK: /ip4/127.0.0.1/tcp/61021 -> peer 12D3KooWEoNmuqujrpn21NSqgtHXi5pUYpA7SXdPqYrRVeEdnTjo
test control_socket_over_tcp_serves_the_unmodified_client ... ok
```

Bind, connect and `identify()` all work over TCP, and the peer id round-trips correctly. That is the Windows control-socket shape executing for the first time.

Two honest caveats: this ran on macOS, so it exercises the **TCP code path** but not Windows itself (no unix-socket fallback differences, no Windows networking stack, no Git-Bash/service-account specifics). And I ran it as a scratch test which I deleted rather than pushing to your branch — **worth adding as a permanent TCP-bound variant of the harness**, since it costs 0.01s and converts an open cutover risk into a standing assertion. Happy to send it as a follow-up PR if you'd rather not touch this one.

## The rest

- `kwaai-p2p-daemon` + `kwaai-p2p`: **91 tests pass**
- Gated interop against a real p2pd — tier 11 (control server) **5 pass**, tier 12 (pipe mode) **4 pass**, including the large-payload cross-implementation case
- CI green: Clippy, Format, Test (ubuntu), WASM, plan

The conformance suite's framing is the right one — it drives the *unmodified* client rather than hand-built protobuf frames, so passing means the ~15 external call sites work against a node with no Go daemon behind it, not merely that the server agrees with a re-reading of the protocol.

Merging.

PR #85, opened 2026-08-05
feat(cli): run_node takes the native path behind native_p2p (native-p2p 6/7)

dazwin:
| # | PR | What |
|---|----|------|

---

## What this adds

The `native_p2p` config flag (default off) and the `node_native` module behind it: setting it makes `kwaainet run-node` assemble PRs 1-5 into a working node — swarm, DHT service, its own handlers, and a control socket on the same path — with no Go `p2pd` child process at all.

## Why

PRs 0-5 built the parts and proved each against a real p2pd, but nothing assembled them, so none of it ran in a node. This is the assembly, put behind an opt-in flag so it can be exercised on real hardware without touching anyone's default. The flag deliberately adds no companion settings: the native path reuses `port`, `initial_peers`, `identity_key` and `KWAAINET_SOCKET`, so a node's PeerId and control socket are identical either way and migrating between paths is a single boolean rather than a reconfiguration.

## What's in it

**`kwaai-cli/src/config.rs` — the flag.** `native_p2p: bool`, `#[serde(default)]`, initialised `false` in `Default`, and settable through the existing `config set` string dispatcher. The doc comment records what it does *not* yet do: AutoNAT, circuit relay, DCUtR and UPnP are still p2pd-only, so `no_relay`, `force_private` and `trusted_relays` are ignored while it is set, and a node behind a NAT is only reachable on the p2pd path. (That gap closes in PR7.)

**`kwaai-cli/src/node_native.rs` (new, ~590 lines) — the native node.** `NativeNode` assembles what PRs 1-5 built into the node `run_node` otherwise gets from p2pd:

- the swarm on the configured port;
- the DHT service, bootstrap-grade, matching p2pd's `-b`;
- this node's own handlers — `hello` / ollama-proxy / shard-proxy as unary, inference-mux as a raw stream;
- a `ControlServer` on the same socket path p2pd would have used, so the GUI, `kwaainet p2p …` and `shard serve` are unchanged.

Identity comes from the same key file through the same libp2p protobuf encoding, so the peer ID is identical on both paths. Announce and unannounce go through PR4's `announce.rs`, byte-for-byte, on the same 300 s ± 30 s cadence against a 360 s TTL, and the tombstone still runs on clean shutdown.

**`kwaai-cli/src/node.rs` — one branch, then a module.** The two paths differ in their whole lifecycle rather than in a handful of branches — there is no child process to spawn, watch, restart or reap — so `run_node` takes a single early return into `node_native::run_native_node` rather than growing a hundred `if native_p2p` arms. Everything above the branch (PID file, SIGHUP registration, the gRPC IPC surface, identity, credentials, public name, bootstrap-peer list) is shared, and the tail (PID cleanup, deferred auto-update respawn) runs identically for both.

What they share is *shared*, not copied: SigHup, the Ollama watcher, the auto-update respawn and the announce-input helpers are all extracted.

**`refactor(cli): extract the announce inputs from run_node`** — throughput measurement/reporting, the repository URL, and the VPK health poll (both the startup form with its 5×1 s retry and the per-tick refresh) become free functions. Pure extraction; `run_node` calls them in the same order with the same arguments. These are what the native path needs to announce the identical records, and sharing them is what keeps the two paths from drifting on the values the map crawler decodes.

**`kwaai-cli/src/inference_mux.rs` — native mux serving.** The mux frame loop is extracted into `serve_mux_frames`, generic over the stream halves, with `start_native_inference_mux_server` built on `NetworkHandle::accept_streams`. The p2pd path binds a local TCP listener, has the daemon forward inbound streams to it, and consumes a `StreamInfo` prologue describing the forwarded stream. Natively there is no forwarding hop: `accept_streams` yields the negotiated libp2p stream itself, so there is no listener, no prologue, and one fewer place for backpressure to be lost. Both paths now run the identical frame loop, so the per-request concurrency model is defined once.

**`kwaai-p2p-daemon/src/lib.rs`** — re-exports `default_socket_addr` so the native node can bind the same control-socket path p2pd would have used.

## Risk

**No default behaviour change.** What makes it safe:

- `native_p2p` defaults to `false` (`#[serde(default)]` on a `bool`, and `native_p2p: false` in `Default for KwaaiNetConfig`), so an existing config file with no such key deserialises to the p2pd path.
- The native path is reached through **exactly one branch** in `run_node` (`node.rs:161`), which returns early. With the flag unset, control never enters `node_native` and the p2pd code below runs verbatim.
- The `inference_mux` and announce-input changes are pure extractions — the p2pd path calls the same code in the same order with the same arguments.

Worth looking at closely:

- **The `node.rs` refactor is the largest diff here** (~650 lines changed) and is the one place a "pure extraction" claim could hide a behaviour change on the default path. The extracted helpers (throughput/reporting, repository URL, VPK health poll with its 5×1 s startup retry) are where to check that call order and arguments really are unchanged.
- The `serve_mux_frames` generalisation now backs the p2pd path too; a reviewer should confirm the p2pd caller still supplies the prologue-consuming wrapper it did before.
- The flag's documented limitations are load-bearing: **a NATed node must not enable this yet**, since NAT traversal does not land until PR7.

## Test coverage

**Tier 13 — `core/crates/kwaai-network-tests/tests/13_native_node_assembly.rs`.** Tiers 09-12 each proved one component in isolation; none assembles them. This tier composes them the way `NativeNode::start` does — identity from a key file, swarm, DHT service, control socket, in that order — and asserts the properties a node has that its parts do not:

- two nodes: B bootstraps to A, announces a subkeyed block record, and A serves it back as `FOUND_DICTIONARY` with the value byte-identical;
- a node serves its own announced records to a remote peer (the local half of announce, which makes a node a replica of itself);
- an unchanged `P2PClient` on the node's own control socket does identify plus a unary round trip that crosses both the socket and the wire;
- the socket and the swarm are **one node** — a `ControlServer` bound to a different handle would pass every tier-11 test and still be broken;
- the peer ID comes from the key file, so migrating keeps the node's DID, credentials and map entry.

No p2pd is spawned and none is required, which is the point.

**Deliberately omitted:** a `kwaainet run-node` spawn test. It would write the real PID file, bind the real socket and take the fixed gRPC port — process-global state that collides with a developer's running node and with a second copy of itself — to add only argument parsing and the config round trip over what tier 13 already covers.

## Stacked on

`native-p2p-pr5-control-socket`

---

- `cargo test --workspace`: **803 passing, 0 failures.**
- Daemon run, `native_p2p: false`: starts the Go p2pd, connects to 2 bootstraps, no panic, clean shutdown.
- Daemon run, `native_p2p: true`: starts the native stack with **no p2pd child process**, registers the hivemind DHT service, bootstraps to 2 peers, and stores to a live bootstrap (`1/1 stored`).

---

### DIFF
--- a/core/Cargo.lock
+++ b/core/Cargo.lock
@@ -3292,6 +3292,7 @@ dependencies = [
  "tempfile",
  "tokio",
  "tokio-stream",
+ "tokio-util",
  "tonic",
  "tracing",
  "tracing-subscriber",
--- a/core/Cargo.lock
+++ b/core/Cargo.lock
@@ -3109,6 +3109,7 @@ dependencies = [
  "rmp-serde",
  "serde",
  "serde_json",
+ "sha1",
  "sha2",
  "tempfile",
  "tokio",
--- a/core/crates/kwaai-cli/Cargo.toml
+++ b/core/crates/kwaai-cli/Cargo.toml
@@ -34,6 +34,9 @@ sysinfo = { workspace = true }
 # Async
 tokio = { workspace = true }
 futures = { workspace = true }
+# `compat` bridges libp2p's `futures::io` raw streams onto tokio's traits, which
+# is what lets the native inference-mux path reuse the tokio-based frame loop.
+tokio-util = { version = "0.7", features = ["compat"] }
 
 # HTTP (IP detection, health check, update check)
 reqwest = { workspace = true }
--- a/core/crates/kwaai-cli/src/cli.rs
+++ b/core/crates/kwaai-cli/src/cli.rs
@@ -286,7 +286,7 @@ pub enum ConfigAction {
     ///
     /// Valid keys:
     ///   model, blocks, start_block, port, use_gpu, log_level,
-    ///   public_name, public_ip, announce_addr, no_relay,
+    ///   public_name, public_ip, announce_addr, no_relay, native_p2p,
     ///   vpk_enabled, vpk_mode, vpk_local_port,
     ///   auto_rebalance, rebalance_interval_secs, rebalance_min_redundancy
     ///
--- a/core/crates/kwaai-cli/src/config.rs
+++ b/core/crates/kwaai-cli/src/config.rs
@@ -126,6 +126,23 @@ pub struct KwaaiNetConfig {
     #[serde(default = "default_force_private")]
     pub force_private: bool,
 
+    /// Run the node on the in-process rust-libp2p stack instead of spawning the
+    /// Go `p2pd` child process.
+    ///
+    /// The native path reuses every other setting on this struct — `port`,
+    /// `initial_peers`, `identity_key` (so the PeerId is identical either way)
+    /// and `KWAAINET_SOCKET` — and serves the same p2pd control socket, so
+    /// external clients (the GUI, `kwaainet p2p …`, `shard serve`) cannot tell
+    /// the two apart. What it does **not** do yet is NAT traversal: AutoNAT,
+    /// circuit relay, DCUtR and UPnP are still p2pd-only, so a node behind a NAT
+    /// is only reachable on the p2pd path. `no_relay`, `force_private` and
+    /// `trusted_relays` are therefore ignored while this is true.
+    ///
+    /// Defaults to false — the p2pd path stays the default until the NAT slice
+    /// lands and the cutover in Phase 5 flips it.
+    #[serde(default)]
+    pub native_p2p: bool,
+
     #[serde(default)]
     pub health_monitoring: HealthConfig,
 
@@ -633,6 +650,7 @@ impl Default for KwaaiNetConfig {
             initial_peers: default_peers(),
             trusted_relays: default_trusted_relays(),
             force_private: default_force_private(),
+            native_p2p: false,
             health_monitoring: HealthConfig::default(),
             model_dht_prefix: None,
             model_repository: None,
@@ -851,6 +869,7 @@ impl KwaaiNetConfig {
             }
             "announce_addr" => self.announce_addr = Some(value.to_string()),
             "no_relay" => self.no_relay = parse_bool(value)?,
+            "native_p2p" => self.native_p2p = parse_bool(value)?,
             "start_block" => {
                 self.start_block = value
                     .parse()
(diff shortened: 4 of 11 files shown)


RezaRassool (approved):
Approve. This is the riskiest PR in the stack for the running fleet — `node.rs` +369/−252, the live node's main path — so that is where the review went.

## The p2pd path is preserved, and I checked the two ways it could break

**The branch is a clean early return.** `run_node` opens with

```rust
if config.native_p2p {
    info!("native_p2p enabled — running without the Go p2p daemon");
    …
    return Ok(());
}
```

so everything below is the existing p2pd path, unreached when the flag is off.

**The 252 "removed" lines are not removed.** `maybe_auto_update`, `shutdown_signal`, `jitter_secs` and `compute_effective_tps` all still live in `node.rs`; the diff shows their signature lines changing visibility so `node_native` can share them. Both paths call `maybe_auto_update` and `shutdown_signal`, so auto-update and shutdown behaviour cannot diverge between stacks — which is the failure mode I would most expect from a change shaped like this, and it is avoided by sharing rather than duplicating.

**An upgrading node cannot switch stacks by accident.** `pub native_p2p: bool` carries `#[serde(default)]` and `Default` sets `false`, so an existing `config.yaml` with no such key deserialises to the p2pd path. Opting in is explicit.

## Verification

Tier 13 gated against a real environment — **5 pass**, and the roster is well chosen:

```
the_peer_id_is_the_identity_key_file_not_a_fresh_one ... ok
the_control_socket_reaches_the_swarm_the_node_is_running  ok
an_external_client_drives_the_assembled_node ......... ok
announced_records_are_findable_on_the_bootstrap ...... ok
a_node_serves_its_own_announced_records .............. ok
```

The first is the one that protects the cutover: a node flipping stacks must derive the same PeerId from the same key file, or it silently becomes a different node to the fleet.

Full workspace builds. CI green. `kwaainet` bin: 84 pass, 2 fail — both `grpc_server` tests, which bind fixed port 8093 and fail for anyone with a node running locally; verified pre-existing on clean `main`, unrelated to this PR.

## Field evidence

Unusually for a review, I have run this code in anger. My macOS node was built from this stack and ran the native path for a full session: **60,945 unary RPCs over 60 minutes**, RSS plateauing (second-half slope −1.2 MB/hour), FDs and threads flat, relay reservations held on both bootstraps, DHT announces `1/1 stored` throughout. Both metro machines have run it for days on the live network, keeping their peer IDs and staying `online` on the map with blocks 0–32 — indistinguishable to the rest of the fleet.

Two real bugs surfaced from that soak, and both are fixed and already merged earlier in this stack. That is a stronger signal for this PR than the test suite alone.

Merging.

PR #86, opened 2026-08-05
feat(p2p): NAT traversal on the native path + libp2p 0.56 (native-p2p 7/7)

dazwin:
| # | PR | What |
|---|----|------|

---

## What this adds

NAT traversal on the native path — AutoNAT, circuit relay, DCUtR and UPnP — driven by a reachability state machine and a relay manager, so a native node behind a NAT becomes reachable and re-announces itself when its reachability changes. Also upgrades the workspace to libp2p 0.56.

## Why

PR6 shipped a native node that only worked if it was directly reachable; `no_relay`, `force_private` and `trusted_relays` were documented as inert. That is the last functional gap between the native path and p2pd, and it is the one that matters most in the field, where most nodes are behind something. Beyond the feature itself, this slice is where the native stack was first driven against production and the docker NAT test bed, so it also carries the fixes that only appear once real peers are involved: routed dials, kad address hygiene, and the libp2p upgrade that replaces a global port-reuse flag which turned out to break ordinary dialing.

## What's in it

**`kwaai-p2p/src/addresses.rs` (new) — multiaddr classifier.** Every part of this slice asks the same question — is this address any use to a peer that is not us? The reachability machine asks it of identify-observed addresses; the relay manager asks it of a candidate relay's listen addrs. The rules are a faithful port of the p2pd path's `is_announceable_addr` / `is_globally_routable_v4`, not an improved version, because two of their decisions look like bugs and are not: a `/p2p-circuit` address is announceable whatever IP fronts it (the relay's address is the part that must be routable, and we reached the relay to get the reservation); and RFC5737 documentation ranges and RFC2544 `198.18/15` are accepted, because they are IANA-reserved but not LAN-private and the docker nat-test topology is built on them. A golden test pins that, naming the topology, because if it flips the whole test bed goes dark. `peer_id_from_multiaddr` and `strip_p2p` move here from `service.rs` with their tests, no behaviour change.

**`kwaai-p2p/src/behaviour.rs` — composing the behaviours.** autonat, relay (client + toggled hop server), dcutr and toggled upnp join `KwaaiBehaviour`. `relay::client::Behaviour` cannot be constructed in `KwaaiBehaviour::new` — the client behaviour and the circuit transport are one thing split in two, and only `SwarmBuilder::with_relay_client` can make the pair — so it is passed in and the `with_behaviour` closure grows a second argument; `with_relay_client` goes *after* `with_dns` because the circuit transport wraps the resolved one. Four autonat knobs move off their defaults: `only_global_ips` follows a new `require_global_ips` config field defaulting false, because autonat's `is_benchmarking` check is the one place in rust-libp2p that rejects the nat-test bed's `198.18/15` addresses; boot_delay/retry/refresh shorten so a node settles its reachability before its first announce; `confidence_max` stays at 3 as the flapping damper. New config: `trusted_relays`, `relay_server`, `enable_upnp`, `force_private`, `external_addr`, `require_global_ips`, `max_relay_reservations`, `identify_min_confirmations`. `for_tests()` turns the hop server and upnp off (SSDP is a LAN broadcast with no business firing from CI).

**`kwaai-p2p/src/reachability.rs` (new) — the state machine.** Four sources can claim to know whether we are reachable — operator declaration, a UPnP mapping, an AutoNAT dialback, a crowd of identify observations — and they disagree, so this is a **priority ladder**: `Source`'s `Ord` derive *is* the ladder, and a weaker claim can never overwrite a stronger one. Two rules look like bugs and are not, so both are in the module docs and pinned by tests: a declared `external_addr` outranks `force_private` (with a warning), because an operator who typed in their public address knows something a probe cannot disprove; and UPnP expiry returns to **Unknown, not Private**, because a lapsed mapping means we stopped knowing, not that we learned we are unreachable. The machine does no I/O — it returns `Effect`s the service applies — which keeps every rule testable without a network and gives one place where external addresses enter and leave the swarm. `force_private` starts Private so reservations begin at t=0; everyone else starts Unknown, and after a 45 s grace timer identify consensus decides by highest distinct-observer count among announceable addresses, tie-broken by address so the verdict cannot differ between runs on identical evidence.

**`kwaai-p2p/src/relay_manager.rs` (new).** Holds circuit reservations while reachability says Private. A plain state machine polled from the swarm loop rather than a task or a behaviour, because it needs `&mut Swarm` *and* swarm-level listener events and neither of the other shapes sees both. It deliberately does **not** refresh — libp2p-relay already renews at 3/4 TTL — it does recovery, for a relay that restarts, vanishes or refuses. That signal is `SwarmEvent::ListenerClosed`, not a relay-client event (`relay::client::Event` has no failure variant at all). Slots are keyed by `ListenerId`, making a duplicate reservation on one relay structurally impossible. A failed relay gets 30 s doubling to a 15-minute ceiling with ±20% jitter, and the cursor advances at the same time, so refusal rotates to the next candidate instead of stalling — that is the production bootstraps' documented behaviour, not an edge case. Candidates come from identify hop advertisements, with `trusted_relays` as a pure operator override; explicitly *not* from the kad routing table, which is the `-relayDiscovery` behaviour p2pd always ran disabled.

**`kwaai-p2p` — announce watch channel.** `AnnounceState` carries reachability, `using_relay`, `announceable` and an epoch — and deliberately **no addresses**, because the DHT record has no multiaddr field. That absence is what makes the channel quiet: an address can change, a reservation can migrate, identify can push a dozen times, and none of it alters this struct. Sends are equality-gated on top, and `announce_state()` hands each caller a receiver with the current value already marked seen. `using_relay` tracks a **confirmed** reservation only. Also adds `NetworkHandle::peer_protocols` (identify now records a peer's protocol list, dropped when its last connection closes) and `NetworkHandle::reachability`.

**`kwaai-cli` — NAT traversal on the native run_node path.** Maps the config fields that drove p2pd's NAT flags onto `NetworkConfig` so a node migrating between paths behaves the same: `force_private` → `-forceReachabilityPrivate`, `no_relay` → `-relay`, `trusted_relays` → `-trustedRelays`, `announce_addr`/`public_ip` → `-announceAddrs`, UPnP for `-natPortMap`. Closes both gaps PR6 flagged: `using_relay` now comes from the swarm (a confirmed circuit reservation) rather than being degraded to "no announce address is configured"; and a reachability change re-announces in place, via the watch channel, with a 10 s settle window absorbing the startup burst. Two follow-up fixes: the settle window is armed as a **deadline** rather than an inline sleep (sleeping in the select arm held shutdown, SIGHUP and Ollama-recovery hostage for the full 10 s), and the settle arm now publishes any state whose epoch it has not successfully announced — gating on `using_relay` alone meant a Private-with-no-circuit → Public transition never re-announced — while an error on the watch shuts the node down instead of silently disabling the arm forever.

**Five fixes found by running this against production and the NAT test bed:**

1. **Routed dials.** Go's p2pd wraps its host in a routed host: `NewStream` to a peer with no known addresses transparently runs a DHT `FindPeer` first. rust-libp2p has no equivalent, so a `CallUnary`/`OpenRawStream` to a server discovered from a petals record (peer ID and nothing else) dialed with an empty address set and failed with `"Unsupported resolved address: /p2p/<id>"` — exactly what `kwaainet shard run` hit against live block servers. The service now parks such requests in `pending_routed`, runs one `get_closest_peers` walk per peer burst, and forwards on completion; a walk producing no addresses fails them with a clear "peer not found in DHT".
2. **Bare `/p2p/` connects, and empty-address kad poisoning.** `shard run`'s pre-connect pass sends CONNECT with a bare `/p2p/<id>`, which natively went straight to `dial()` and seeded Kademlia with an *empty* stripped multiaddr, making `dispatch_routed` believe addresses were known and skip its DHT walk. `dial()` no longer seeds kad with an empty address, `known_addresses` ignores empty entries, and a bare-`/p2p/` ConnectPeer becomes a routed request. **Verified live:** `kwaainet shard run` completes distributed inference against production block servers.
3. **Filtering kad addresses on the way out.** Filtering identify was not enough — `kad::Behaviour::discovered` also inserts whatever multiaddrs a remote reports during a query walk, verbatim, inside libp2p. Five minutes after start, node-f held 46 loopback and 16 foreign-LAN entries for node-g against 21 usable ones, and a Connect from the GUI dialed loopback and reached *itself*. `known_addresses` is where every consumer reads them back, so one check there covers every ingestion path. Live connections stay unfiltered: an address we are connected on is reachable by definition, and a relayed connection to a NATed peer is exactly the case that must not be dropped.
4. **Capping routing-table addresses learned from identify.** kad's `Addresses` is an unbounded `SmallVec` with only reactive removal. A symmetric-NAT peer supplies exactly the input that breaks this: measured in the NAT test bed, node-a accumulated 20+ addresses for node-h and re-added them ~7× faster than kad evicted them. Capped at 6 (the SmallVec inline size), evicting oldest-first so a peer that legitimately moves is not frozen. Only the identify path is capped — `AddKadAddress` and the dial path are operator intent, and a peer flooding identify must never evict a hand-configured bootstrap address.
5. **`fix(cli): standalone shard serve writes its own PID file.`** `shard_is_ready()` requires the ready sentinel *and* a live PID from `shard.pid`, but only the supervised launch path wrote it — so a hand-launched `kwaainet shard serve` left the node announcing state 0 (map: offline) forever despite a fully loaded shard. **This affects both paths.**

**libp2p 0.53 → 0.56.** The `port_reuse(true)` flag added to make DCUtR hole punching work broke ordinary dialing, and the fast unit tests could not see it — all 67 finish in 0.00 s without ever constructing a swarm; the integration suite caught it. Upstream hit the same wall: 0.54 deprecated the global flag ("the port reuse policy is now decided on a per-connection basis by the behaviour"), which is exactly the granularity DCUtR needs. So the flag is dropped and ordinary dials call `allocate_new_port()` at each of the four dial sites — `DialOpts` defaults to best-effort reuse, which is wrong for us because binding a dial socket to our own listen port fails `EADDRINUSE` outright (SO_REUSEPORT lets several *listeners* share a port, not a listener plus an outbound connect; measured with a socket2 probe on macOS errno 48 and Linux errno 98). libp2p-tcp 0.44 does re-dial on a fresh port but only for `AddrNotAvailable` from `connect`; ours is `EADDRINUSE` from `bind`, one step earlier where `?` returns before the fallback can run. API changes absorbed: `handle_established_outbound_connection` takes a sixth `PortUse` argument; `identify::Event::{Received,Error}` gained `connection_id`; `GetClosestPeers` yields `PeerInfo` not `PeerId`; `kad::Config::set_protocol_names` is gone (`Config::default()` already uses `/ipfs/kad/1.0.0`, so the example calls were no-ops); `request_response::Event::Message` gained `connection_id`. **multistream-select stays at 0.13.0, so PR0's slash-less patch is unaffected.** Also normalises `known_addresses` to strip `/p2p/<peer-id>` on the way out — kad has stored fully-qualified addresses since 0.54 while connection addresses never carry one, producing `/p2p/<id>/p2p/<id>`, which parses, resolves to the right peer, and then fails to dial.

**`refactor(announce)` — one ONLINE/JOINING gate.** `if ShardManager::shard_is_ready() { 2 } else { 0 }` appeared verbatim at four announce sites (twice in `node.rs`, once each in `node_native.rs` and `announce.rs`). Collapsed into `KwaaiNetConfig::announce_state()`, which returns exactly what the inline conditional did.

## Risk

**The `native_p2p` flag still defaults to off, and the native path is still reached through the single early branch in `run_node`. But unlike PRs 0-6, this PR is *not* purely additive to the default path — three changes reach it, and all three deserve review.**

1. **`default_trusted_relays()` is now empty, and this changes the p2pd path.** It previously named the two production bootstraps, which have a documented `RESERVATION_REFUSED` history — with an active relay manager that is a visible refuse-and-back-off loop ending in no relay. Empty is also what the surrounding p2pd code always said it wanted: `node.rs` explains twice that it deliberately does not auto-promote bootstraps to relays "because bootstraps may not run a hop relay service", while the default did exactly that behind its back. **The consequence to watch:** `node.rs` skips its IDENTIFY-discover-and-restart cycle when `trusted_relays` is non-empty, so with the old default *every* node skipped it. That skip is written for deliberately NATed nodes whose only address is a circuit, and it now applies to those nodes only — which is what the comment describes. Nodes that were silently relying on the accident will start discovering and announcing their observed addresses. Anyone with `trusted_relays` set explicitly in their config file is unaffected.
2. **The `shard serve` PID-file fix affects both paths.** A hand-launched `shard serve` will now correctly announce state 2 where it previously announced 0.
3. **The libp2p 0.53 → 0.56 upgrade is workspace-wide.** Seven crates depend on libp2p (`kwaai-cli`, `kwaai-hivemind-dht`, `kwaai-network-tests`, `kwaai-p2p-daemon`, `kwaai-p2p`, `kwaai-trust`, `map-server`) and all are recompiled against the new version. Only `core/examples/*` needed source changes, but a reviewer should not treat "it compiles" as sufficient for `map-server` and `kwaai-p2p-daemon`, which run in production on the default path.

Also worth noting: `KwaaiNetConfig::announce_state()` is a pure refactor of a predicate the p2pd path already evaluated, so behaviour there is unchanged; and `refactor` is the right word for it, but it does mean `config.rs` now imports `daemon::ShardManager`.

**Observed difference, `native_p2p` off, same config, isolated HOME:** on the PR0 base the node reaches `✅ Announced to 2 of 2 bootstrap peers`; on this branch the same run stops at `No explicit announce address — discovering addresses via IDENTIFY...` and does not announce. That is change (1) above, reproduced end-to-end rather than argued from the diff — it is the single most important thing to review in this PR.

## Test coverage

**In-process — `kwaai-p2p/tests/reachability.rs`** — the priority ladder, both counter-intuitive rules (declared address outranking `force_private`; UPnP expiry → Unknown), the 45 s grace timer and the identify-consensus verdict with its deterministic tie-break. Includes the R3 fix: `observed_addrs` never dropped anything, so the consensus count could only rise and a node that changed networks would keep being promoted by peers that saw it at its old address — observers are now pruned when their last connection closes. Also the "re-confirming an address we already stand on emits nothing" case, found while testing: it used to retract and re-add, churning identify pushes and briefly dropping kad out of server mode, and AutoNAT re-probes every refresh interval so that was the common path.

**In-process — `kwaai-p2p/tests/relay.rs`** — the suite runs in ~3 s and passed three consecutive runs; before the dial/identify ordering fixes the same tests took 32 s and failed intermittently, every failure hidden behind a first-attempt failure and a backoff retry. Three things the tests forced out that no amount of reading would have: `listen_on` does not wait for a dial (hence `Dial` and `Listen` as separate actions); nor does it wait for identify (hop is not in the connection's protocol set until identify lands, so the trigger is the identify event, not `ConnectionEstablished`); and a relay with no confirmed external address is not a relay (its hop server accepts and returns an empty address list). A pending-dial timeout was added because a dial that resolved neither way would hold its slot forever.

**In-process — `kwaai-p2p/tests/dcutr.rs`** — with an explicit caveat stated in the test header: hole punching is a property of NATs, there is no NAT on loopback, and **none of these tests claims to demonstrate one**. Real success rates against cone and symmetric NATs are docker nat-test topology work. What *is* provable is the wiring: Bob, given nothing but Alice's circuit address, connects through the relay and ends up with a connection whose address carries no `/p2p-circuit` — a direct address he was never told, derived by dcutr from the relayed connection. Plus two regression guards: an upgrade attempt must not tear down the connection it was improving, and dcutr must leave an ordinary direct dial alone.

**In-process — `kwaai-p2p/tests/swarm.rs` and `service_unary.rs` additions** — the announce watch channel tested for the regression it exists to prevent (three peers connecting, identifying and disconnecting produce no wake-up and no epoch bump) and for the transition that must fire (a reservation acquired then lost). Three `service_unary` tests for routed dials: a three-node resolve-through-the-DHT round trip, a clean failure for an unknown peer after the walk, and a clean failure from a fully isolated node. Also an assertion that `/libp2p/dcutr` is *absent* from identify over a direct connection — rust-libp2p installs the dcutr handler only on relayed connections, unlike go-libp2p which advertises it statically — with a note attached so we hear about it if upstream changes.

**Live, `#[ignore]`d, read-only against production:**

- `live_bootstrap.rs` gains a snapshot of the bootstraps' identify protocol lists. This closed the design's one empirical unknown — whether the production bootstraps answer AutoNAT dialbacks — since they run hivemind's own go-libp2p daemon and our p2pd spawn flags say nothing about it. Measured: both advertise exactly eight protocols, identically, including `/libp2p/autonat/1.0.0` and `/libp2p/circuit/relay/0.2.0/{hop,stop}`. So AutoNAT is there and relay hop is there — a native node has two AutoNAT servers from its first dial, identify consensus is a genuine fallback rather than the primary source, and relay discovery finds hop candidates immediately. A hop advertisement is not a promise, which is why refusal has to rotate rather than fail.
- `live_map_probe.rs` and `live_circuit_reachability.rs` — the diagnostic pair that cracked the map freeze: dump the dictionaries the health crawler reads (with structural decode of each record) and `rpc_ping` a NATed node from outside through its relay circuit.
- `tests/kwaai-network/results/map-freeze-probe/` — a python-hivemind crawl harness reproducing the health-map crawler against production with the exact hivemind version the service pins (docker/amd64 and native-macOS variants). This is the harness that reproduced the dictionary-subkey freeze and verified the fix, and it is the seed for the tracked python-golden interop follow-up.

**One deviation from the plan, stated so it is not mistaken for an oversight:** `rpc_ping`'s hardcoded `available = false` stays. Sourcing it from our own reachability is the wrong quantity — `available` means "I dialed you back and you answered", a statement about the *caller* — so a public node would end up asserting the reachability of every peer that pings it. Closing it properly needs the actual `validate` dial-back and remains a tracked follow-up; the doc comment now says so.

**Docs:** `docs/NATIVE_P2P_MIGRATION.md` reaches 767 lines with the Phase 4 design notes and verification record. `tests/kwaai-network/results/metrics.jsonl` gains 954 lines of append-only tier-gate and live-run metrics across phases 0-4 (squashed from 15 commits; strictly appended, so the collapsed form is byte-identical).

## Stacked on

`native-p2p-pr6-cli-integration`

---

- `cargo test --workspace`: **890 passing, 0 failures.**
- Daemon run, `native_p2p: false`: starts the Go p2pd, connects to 2 bootstraps, no panic, clean shutdown.
- Daemon run, `native_p2p: true`: starts the native stack with **no p2pd child process**, registers the hivemind DHT service, bootstraps to 2 peers, and stores to a live bootstrap (`1/1 stored`).
- Relay path exercised live: both production relays accepted reservations, yielding real `/p2p-circuit/` addresses, and the announce state machine advanced `using_relay=false -> true` before storing.
- Live suite green: `live_bootstrap` (2), `live_map_probe` (1), `live_dht_announce` (1, ephemeral identity in the `_kwaai.test.ephemeral.*` namespace, 60s TTL). `live_circuit_reachability` self-skipped: it needs a genuinely NATed target and the test host was directly reachable.

---


core/crates/kwaai-p2p/src/relay_manager.rs
```diff
@@ -0,0 +1,948 @@
...
+        // add, but this identify may be the one that makes a pending
+        // reservation negotiable, so hand it to `on_relay_ready`.
+        if self.configured.iter().any(|(p, _)| *p == peer)
+            || self.discovered.iter().any(|(p, _)| *p == peer)
+        {
+            return self.on_relay_ready(peer, now);
+        }
+        // A relay we can only reach at a LAN address is no use to peers who are
+        // not on that LAN.
+        let Some(addr) = listen_addrs.iter().find(|a| is_announceable(a)) else {
```
RezaRassool: @dazwin This is the line behind the metro-win outage your session diagnosed on Aug 11 — flagging it here since #86 is the last of the stack and this is the one finding not yet addressed in it.

`is_announceable()` returns `true` for **any** address containing `/p2p-circuit`, unconditionally. That is correct for the question it was written to answer ("is my own address worth advertising"), and wrong for this one ("can I dial this peer and reserve on it"). A peer whose own advertised address is itself a circuit — i.e. relay-only, which on this fleet is most of them — is accepted as a candidate here, and `circuit_listen_addr` then appends onto it:

```
/ip4/18.219.43.67/tcp/8000/p2p-circuit/p2p/12D3KooWAUGTd8Gs…/p2p-circuit
```

`Swarm::listen_on` rejects that, the candidate goes on backoff, the cursor rotates to the next one — usually also relay-only — and it never recovers. The only candidates this path handles correctly are peers with clean non-circuit addresses, i.e. the two bootstraps, which is why the one reservation metro-win ever held was on a bootstrap and why nothing came back after it lapsed.

Corroborated from an unrelated macOS node, 26h of logs, split by `why=`:

| failure | count | which relays |
|---|---|---|
| `listen_on refused the circuit address` | **2350** | only fleet peers, never a bootstrap |
| `the dial to the relay failed` | 355 | fleet peers |
| `the relay connection closed` | 87 | only bootstraps |
| `reservation ended` | 20 | only bootstraps |

The 2350 spread ~200 apiece across ~12 relay-only fleet peers. The clean split — nesting never touching a bootstrap, relay-side churn never touching a fleet peer — is this bug's signature rather than flaky relays. The two interact badly: bootstrap churn is ordinary and survivable, but this removes the fallback, so ordinary churn becomes a permanent loss of inbound reachability.

There is a tested patch on `proposal/relay-candidate-fix` (`c425e83`) if it is useful: a separate `is_relay_candidate_addr` predicate (a relay must be *directly* dialable), `circuit_listen_addr` returning `Option` so the nested address cannot be constructed, and `?e` on the `listen_on` failure — it currently logs empty, which is why the address in the message was the only clue. Four regression tests, one verified to fail without the fix. Deliberately **not** changing `is_announceable` itself: it is right as written, the bug is the call site.

Entirely your call whether to take that, write your own, or land #86 and follow up — I have not pushed anything to this branch. Worth deciding before merge though, since #86 is the last one and a node in this state reports completely healthy: status green, announce succeeding, nothing surfacing "I have no inbound path". That is why it went unnoticed for hours.

dazwin:
Taken — thanks for the diagnosis, and for the corroborating log split. The clean separation (nesting never touching a bootstrap, relay-side churn never touching a fleet peer) is what makes this conclusive rather than circumstantial.

Pushed as `83e084af`. It is the same three changes you proposed, arrived at independently on `fix/relay-candidate-not-circuit` before I saw this comment, so the shape matches `c425e83` closely:

- `is_relay_candidate_addr` (announceable **and** not a circuit) is what `note_identify` now filters candidates on.
- `circuit_listen_addr` returns `Option` and refuses to build the nested address at all.
- `fill_slots` fails the relay properly if one slips through, freeing the slot instead of retrying forever.
- `error = ?e` on the `listen_on` failure — agreed that the `Display` form omitting the inner cause is what made 2350 identical lines unreadable.

I deliberately left `is_announceable` alone, same reasoning as yours: it is right for the question it answers, and a reserved circuit address is exactly what an unreachable node has to advertise. The bug is the call site.

What I did **not** take from `proposal/relay-candidate-fix` is the two scratch artefacts (`stress/BUG-relay-candidate-nesting.md` and `stress/relay-candidate-fix.patch`, ~400 lines) — the investigation is captured in the commit message and the doc comment on the new predicate, so the patch file would just be a stale copy of the diff beside it.

Verified: `cargo test -p kwaai-p2p` is green (78 unit + all integration). I confirmed `a_circuit_address_is_never_a_relay_candidate` genuinely catches this by reverting the predicate to plain `is_announceable` — it fails, and passes again with the fix in place.

The branch is also rebased onto current `main`, so it now carries #75 and #93.

On the wider point — a node in this state reporting completely healthy is the part that worries me more than the bug. Announce succeeding is not evidence of inbound reachability, and nothing currently surfaces the difference. Worth a separate issue for a "no inbound path" signal rather than folding it into this one.

core/crates/kwaai-cli/src/node_native.rs
```diff
@@ -127,6 +140,29 @@ impl NativeNode {
...
+            // `-forceReachabilityPrivate`. Defaults true, so relay reservations
+            // start immediately rather than after an AutoNAT round.
+            force_private: config.force_private,
+            // `-announceAddrs`. An operator declaration, so it outranks
+            // `force_private` and AutoNAT cannot demote it.
+            external_addr: configured_announce_addr(config),
+            // Deliberately permissive: this is the only address-class filter in
+            // rust-libp2p 0.53, and turning it on would classify the docker
+            // nat-test topology's `198.18/15` addresses unreachable.
+            require_global_ips: false,
```
dazwin: The global IPs filter can be turned on - this is a leak from my test environment and I'll patch there.
Remove these comments and keep the initialization concise.

core/crates/kwaai-p2p/src/config.rs
```diff
@@ -83,6 +83,97 @@ pub struct NetworkConfig {
...
+    ///
+    /// A pure **operator override**. The normal supply of relay candidates is
+    /// identify: any peer that advertises `/libp2p/circuit/relay/0.2.0/hop` is
+    /// a candidate, which on the live network includes both bootstraps. Set
+    /// this only to pin specific relays — an isolated test topology, or a
+    /// deployment where a known-good relay must be preferred. Configured
+    /// relays are tried first and dialed alongside the bootstrap peers so the
+    /// reservation does not wait on a lazy dial.
+    #[serde(default)]
+    pub trusted_relays: Vec<String>,
```
dazwin: Does this provide a tangible benefit over using trusted_peers? Was it added to provide feature parity with p2pd, or invented here? This is just a question, so push back if warranted.

core/crates/kwaai-p2p/src/reachability.rs
```diff
@@ -0,0 +1,826 @@
...
+}
+
+/// What we currently believe about our own reachability.
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub enum Reachability {
+    /// No verdict yet. Announce is deferred; kad stays in client mode.
+    Unknown,
+    /// Reachable at `addr`, on the strength of `source`.
+    Public { addr: Multiaddr, source: Source },
+    /// Not reachable directly. Relay reservations are wanted.
```
dazwin: Remove these three obvious comments

core/crates/kwaai-p2p/src/reachability.rs
```diff
@@ -0,0 +1,826 @@
...
+/// identify push, entirely separate from announcing. The only address-derived
+/// fact the record contains is the `using_relay` boolean.
+///
+/// That is what makes this channel quiet: an address can change, a reservation
+/// can move from one relay to another, identify can push a dozen times, and
+/// none of it alters this struct. Only a genuine change in *how reachable the
+/// node is* does.
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub struct AnnounceState {
+    /// Public, Private, or not yet known.
```
dazwin: Remove this comment - no value over reading ReachabilityKind

core/crates/kwaai-p2p/src/reachability.rs
```diff
@@ -0,0 +1,826 @@
...
+    /// is Unknown — a node that does not know where it stands should not be
+    /// telling the network it is Direct.
+    pub announceable: bool,
+    /// Increments on every published change, so a consumer can tell "no change"
+    /// from "changed back to what it was".
+    pub epoch: u64,
+}
+
+impl AnnounceState {
+    /// The state before anything is known.
```
dazwin: Remove obvious comment

core/crates/kwaai-p2p/src/service.rs
```diff
@@ -114,6 +205,28 @@ impl NetworkService {
...
+                //
+                // Forcing reuse globally (which is what the old flag did) made
+                // every dial bind the listen port, so a second dial to an
+                // endpoint already connected collided on the 4-tuple and failed
+                // `AddrNotAvailable`. Upstream hit the same wall and added a
+                // fallback in libp2p-tcp 0.44: on that error it re-dials from a
+                // fresh port rather than failing the connection.
+                //
+                // Neither mechanism rescues a symmetric NAT, which re-maps the
+                // port per destination — that is what the relay fallback is for.
```
dazwin: These comments are redundant (they're documenting behavior that doesn't exist and was removed in legacy code) - remove them and learn from this.

RezaRassool (approved):
Approve on the merits — needs a mechanical rebase before it can merge (recipe at the bottom, verified).

**Disclosure first:** `33ed7ee` is my proposal branch applied nearly verbatim — same predicate, same `Option` return, same test names. I cannot claim to have independently design-reviewed my own code, so I verified it *functionally* instead: reverting only the `note_identify` line on **your** tree makes `a_relay_that_is_itself_relay_only_is_not_a_candidate` fail, restoring it passes. The regression test is load-bearing on your implementation, not just mine. Everything else below is your work and reviewed normally.

## The libp2p 0.53 → 0.56 bump — the risk nobody had tested

Three minor versions across **7 crates**, including `kwaai-cli`, `kwaai-p2p-daemon`, `kwaai-trust` and `map-server`. Every test so far — yours and mine — has exercised the *native* path. Nothing had checked whether the Go p2pd path still works against the bumped libp2p types, which is what every node in the fleet runs today.

So I built this branch and ran a node with `native_p2p: false`:

```
✅ Connected to 2 bootstrap peer(s) in 0.0s
STORE response from QmQhRuhee…: 1/1 stored
STORE response from Qmd3A8N5aQ…: 1/1 stored
✅ Announced to 2 of 2 bootstrap peers
✅ Announced model to _petals.models registry
Total 4 connection(s): 3 direct, 1 via relay
```

It spawns its own Go daemon, bootstraps, announces and serves its control socket. **The bump does not break the p2pd path.** That was the single largest unverified risk in this PR and it is now measured rather than assumed.

## Tests

**149 pass** in `kwaai-p2p`: 78 unit plus dcutr 3, dht_service 6, raw_stream 12, reachability 12, relay 10, service_unary 14, swarm 14.

All **seven gated interop tiers** run against real p2pd — 07 (7), 08 (3), 09 (4), 10 (3), 11 (5), 12 (4), 13 (5). CI runs none of these, so this is the interop claim demonstrated rather than asserted.

## Field evidence

I have run this stack continuously for a session: 60,945 unary RPCs over an hour with RSS plateauing (second-half slope −1.2 MB/hour), FDs and threads flat, relay reservations held on both bootstraps. Both metro machines ran it for days on the live network keeping their peer IDs. Three bugs came out of that soak; all three are now in this branch.

## Needs a rebase — mechanical, and I verified the result

`main` moved under you tonight (#75, #76, #77, #78, #97 landed). Two trivial consequences:

1. `f2e5abc` (lease_v1 fixture) is now redundant — #97 landed the same fix.
2. One conflict, a **single comment line**, where your version of that fix carried an explanatory comment and #97's did not.

```bash
git rebase origin/main --empty=drop
# resolve grpc_server.rs by keeping your comment, then:
git rebase --continue
```

I ran exactly that locally: 28 commits over `main`, and the result passes `cargo build`, `cargo test --no-run`, `cargo build --workspace` and all 78 `kwaai-p2p` unit tests. I deliberately have **not** pushed it — #94 is based on this branch and you are actively working on it, so the rebase is yours to do rather than something I should force under you.

## One heads-up

Local `cargo clippy --all-targets -- -D warnings` fails on this branch at `tests/unary.rs:409` ("very complex type"), while CI's clippy passes. That is Rust-toolchain drift — 1.93 flags it, CI's version does not — and it is pre-existing, not introduced here. Same family as your "rustfmt 1.93 drift" commit. Worth a `type` alias at some point so anyone on 1.93 gets a clean local clippy.

Rebase and it is good to go. This is the last of the seven, and it unblocks #94 and #96.

RezaRassool:
Re-verified the rebased head (`5168021`) rather than relying on the approval carrying over from `f2e5abc`, since it is different code.

Two deltas from the rebase I had verified locally, both improvements:

- `tests/unary.rs` — the `ParkedResponders` alias resolves the `type_complexity` failure I flagged, and the comment names the cause precisely (1.93's clippy flags it, CI's does not, so it only ever failed on a local `--all-targets` run). Local `clippy --all-targets -- -D warnings` now **passes**, which it did not before.
- `grpc_server.rs` — the one-line comment conflict resolved the other way. Cosmetic.

All three soak fixes still present on the head: `is_relay_candidate_addr`, `strip_dest_p2p`, and the `WrongPeerId` eviction.

Verified on `5168021`: `cargo build`, `cargo test --no-run`, `cargo build --workspace`, `clippy --all-targets -- -D warnings`, and 78 `kwaai-p2p` unit tests — all pass.

Merging. That completes the seven-PR stack.

PR #87, opened 2026-08-09
fix(rag): upgrade pdf-extract to 0.12 to fix quadratic PDF parsing

dazwin:
pdf-extract 0.10 pulled in lopdf 0.38, whose parser recounts newlines from the start of the buffer (via nom_locate) for every object it reads — O(n²) over the file. A 105 MB PDF spent 10+ CPU-minutes in Document::load without finishing; lopdf 0.42 dropped nom_locate and the same file extracts in 1.2s in release mode. 0.12 also survives malformed PDFs that panicked 0.10's parser.

### DIFF
--- a/core/Cargo.lock
+++ b/core/Cargo.lock
@@ -35,7 +35,7 @@ checksum = "b169f7a6d4742236a0a00c541b845991d0ac43e546831af1249753ab4c3aa3a0"
 dependencies = [
  "cfg-if",
  "cipher",
- "cpufeatures",
+ "cpufeatures 0.2.17",
 ]
 
 [[package]]
@@ -585,12 +585,6 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "72f5acc6cb2ba439de613abc23857ec3d78374d8ed5ac84e9d11336e87da8649"
 
 [[package]]
-name = "bytecount"
-version = "0.6.9"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "175812e0be2bccb6abe50bb8d566126198344f707e304f45c648fd8f2cc0365e"
-
-[[package]]
 name = "bytemuck"
 version = "1.25.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -789,9 +783,9 @@ dependencies = [
 
 [[package]]
 name = "cff-parser"
-version = "0.1.0"
+version = "0.2.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "31f5b6e9141c036f3ff4ce7b2f7e432b0f00dee416ddcd4f17741d189ddc2e9d"
+checksum = "c5810ca1a2b5870df2aab1c03e11c40c361ba51d6e3e361e56310f1cb3b4e087"
 
 [[package]]
 name = "cfg-if"
@@ -813,7 +807,18 @@ checksum = "c3613f74bd2eac03dad61bd53dbe620703d4371614fe0bc3b9f04dd36fe4e818"
 dependencies = [
  "cfg-if",
  "cipher",
- "cpufeatures",
+ "cpufeatures 0.2.17",
+]
+
+[[package]]
+name = "chacha20"
+version = "0.10.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "d524456ba66e72eb8b115ff89e01e497f8e6d11d78b70b1aa13c0fbd97540a81"
+dependencies = [
+ "cfg-if",
+ "cpufeatures 0.3.0",
+ "rand_core 0.10.1",
 ]
 
 [[package]]
@@ -823,7 +828,7 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "10cd79432192d1c0f4e1a0fef9527696cc039165d729fb41b3f4f4f354c2dc35"
 dependencies = [
  "aead",
- "chacha20",
+ "chacha20 0.9.1",
  "cipher",
  "poly1305",
  "zeroize",
@@ -1050,6 +1055,15 @@ dependencies = [
 ]
 
 [[package]]
+name = "cpufeatures"
+version = "0.3.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "8b2a41393f66f16b0823bb79094d54ac5fbd34ab292ddafb9a0456ac9f87d201"
+dependencies = [
+ "libc",
+]
+
+[[package]]
 name = "crc32fast"
 version = "1.5.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -1201,7 +1215,7 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "97fb8b7c4503de7d6ae7b42ab72a5a59857b4c937ec27a3d4539dba95b5ab2be"
 dependencies = [
  "cfg-if",
- "cpufeatures",
+ "cpufeatures 0.2.17",
  "curve25519-dalek-derive",
  "digest",
  "fiat-crypto",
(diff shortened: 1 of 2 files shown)


RezaRassool (changes requested):
The speedup is real and I can confirm it — but I found a text-extraction regression that I think blocks this in its current form. Flagging rather than dismissing, because the problem you're solving is real and 0.10 clearly isn't the answer either.

## What I verified

Benchmarked 0.10 against 0.12 on five PDFs, release mode, same machine:

| file | 0.10 | 0.12 |
|---|---|---|
| LEST WE FORGET rev25 (68 MB) | 1.41s | **0.50s** |
| scaling-trust-ebook | 0.18s | **0.05s** |
| CT-1-Form-and-Instructions | 0.15s | **0.02s** |
| KwaaiNet_The_Linux_of_AI | 0.14s | **0.09s** |
| RAGPerformanceReport-charts | 0.01s | 0.01s |

Consistently 3–7× faster. No argument there.

I could **not** reproduce the O(n²) blow-up, though — the 68 MB file took 1.5s on 0.10, not minutes. That's consistent with your diagnosis rather than against it, since `nom_locate`'s cost scales with object count rather than file size, and that file is image-heavy. I just don't have a file that triggers it, so the 105 MB → 10+ minutes figure is unconfirmed on my side.

## The blocker: 0.12 corrupts text on some fonts

Four of the five files extracted byte-identically. `scaling-trust-ebook.pdf` did not — 96,459 → 94,935 chars, and the difference isn't whitespace cleanup:

```
0.10: Organizations are under pressure to digitize both customer-facing and internal pro-
0.12: Svkermexmsrw evi yrhiv tviwwyvi xs hmkmxmi fsxl gywxsiv1egmrk erh mrxivrep tvs-
```

That's a consistent glyph-mapping offset — a broken font encoding, not a parse failure. Separately, ligatures are being dropped: `Efficiency` → `Eciency`.

Scale of it, using "does this line contain common English words" as a rough readability proxy over substantial (5+ word) lines:

- 0.10 — **92%** readable (1041/1128)
- 0.12 — **80%** readable (908/1129)

So roughly an eighth of that document turns into gibberish.

## Why I think this blocks rather than just warrants a note

Extraction *succeeds*, with a plausible character count and no error. A KB ingested under 0.12 would take the garbage silently, and our own D6 corpus is PDF-sourced. A quadratic parse is at least loud — it hangs and you notice. Silent corpus corruption is discovered much later, if ever.

## Options, roughly in order of preference

1. **Check whether 0.11 has the speedup without the regression** — cheapest path if it does.
2. **Report upstream** — this looks like a lopdf 0.38 → 0.42 font/CMap change, and it's a clean, small reproduction.
3. **Take 0.12 with a guard** in `document.rs` — a readability sanity check that rejects or warns on extraction that comes out mostly non-lexical. Useful independent of this upgrade, honestly.

## Reproduction

```rust
// pdf-extract = "0.10" vs "0.12", release mode
let s = pdf_extract::extract_text("scaling-trust-ebook.pdf").unwrap();
println!("{} chars", s.len());   // 96459 on 0.10, 94935 on 0.12
```
The divergence starts around section "1.4 Efficiency and Client Process Transformation". I can share the file if it isn't one you have.

RezaRassool:
One thing that has changed since my review, which I want on the record here rather than discovered later: **the version this PR replaces carries a high-severity advisory.**

I ran `cargo audit` across the workspace today (it had never been run — it was listed as a gap in the native-p2p security doc). Among the findings:

```
Crate:     lopdf
Version:   0.38.0
Title:     Stack overflow in lopdf via deeply nested PDF objects
ID:        RUSTSEC-2026-0187
Severity:  7.5 (high)
Solution:  Upgrade to >=0.42.0
```

Reached as `lopdf 0.38.0 ← pdf-extract 0.10.0 ← kwaai-rag`, so it is exactly the dependency this branch moves. A malformed PDF with deeply nested objects can overflow the stack, and PDF ingestion takes untrusted input by definition — a document from a user is not a document you wrote.

**This does not unblock the PR.** The text-extraction regression I found stands: `scaling-trust-ebook.pdf` still loses ~1,500 characters on certain fonts under 0.12, and silently corrupting extracted text is not an acceptable trade for a fix to a crash, particularly in a RAG corpus where the corruption propagates into embeddings and is invisible downstream.

But it does change the shape of the decision. This is no longer "a performance improvement with a regression", which can sit indefinitely; it is "a security fix with a regression", which cannot. Staying on 0.10 is itself a choice with a cost now, and the status quo stops being the safe default.

Worth checking whether the two are even coupled: `pdf-extract 0.10 → lopdf 0.38` is the vulnerable pairing, and if a `pdf-extract` release exists that picks up `lopdf >= 0.42` **without** the font-handling change that causes the corruption, that would resolve both at once. If not, the options are narrower — pin `lopdf` directly, land 0.12 and fix the extraction separately, or accept the corruption knowingly on a documented subset. I have not investigated which is available.

No action needed from you on my account; I will pick this up if you would rather not. Flagging it because the priority moved, and because a reviewer weighing the trade a week from now should not have to rediscover this.

For reference, the other high-severity finding in the same area is `quick-xml 0.36.2` — two advisories at 7.5 (unbounded namespace allocation, and quadratic duplicate-attribute checking), a direct `kwaai-rag` dependency, fixed in >=0.41.0. Unrelated to this PR, same untrusted-document ingestion path.

RezaRassool:
Following up on my review, and on the security note above — I have opened #101, which takes this upgrade and closes the regression that was blocking it.

## Why I did not just merge this

I went looking for the middle ground I suggested in my review, and two of the three options turned out not to exist:

- **There is no `pdf-extract` 0.11.** Releases go 0.10.0 (2025-10-03) straight to 0.12.0 (2026-06-25). My "cheapest path" was not a path.
- **`lopdf` cannot be pinned under 0.10.** `pdf-extract` 0.10.0 requires `lopdf ^0.38`, which resolves `>=0.38.0, <0.39.0`, so 0.42 cannot satisfy it. Getting the advisory fix without 0.12 would take a `[patch]` override or a fork, not a version bump.

So the choice really was "corrupt some text" or "keep a high-severity advisory in an untrusted-input path". Both bad, which is what made this sit for a week.

## What #101 does

It carries **both** majors via a renamed dependency — the same approach `kwaai-network-tests` already uses to hold two `prost` versions — and picks per document. 0.12 runs first; the output is scored on whether it reads like language, using "does this line contain a common English word" as the proxy. A broken font map shifts every letter uniformly, so no common word survives and the score collapses. On the clean and garbled versions of the passage from my review the separation is total, 1.0 against 0.0, so this is not a delicate threshold. Only when the score drops does it pay for a second opinion from 0.10, and then it keeps whichever scored better rather than trusting either.

Clean documents never invoke 0.10 and never load the vulnerable parser at all. Your speedup is preserved on the happy path, which is the overwhelming majority of files.

Checked against real inputs: `scaling-trust-ebook.pdf` now gives 96,457 chars with `Organizations are under` present and `Efficiency` intact — 0.10's output rather than 0.12's — and the 66 MB `LEST WE FORGET` takes the fast path in 466 ms with no fallback.

The fallback is bounded three ways: skipped above 64 MB (0.10 is quadratic in object count, which is the hang you were fixing), skipped when there is too little prose to judge so slide decks and tables do not fall through, and refusable outright with `KWAAI_PDF_LEGACY_FALLBACK=0` for anyone ingesting genuinely untrusted PDFs. `cargo audit` will still flag RUSTSEC-2026-0187 while 0.10 is in the tree — deliberate, and called out in the PR.

## On this PR

Closing it in favour of #101, which contains your version bump plus the guard. The diagnosis was yours and it was correct: `nom_locate` was the culprit, 0.10 was not tenable, and the security finding made that sharper rather than softer. Sorry it took a week to get from "blocked" to "here is the way through" — the dead ends above are most of where that time went.

Worth flagging separately since it is the same ingestion path: `quick-xml` 0.36.2 is a direct `kwaai-rag` dependency with two 7.5 advisories, fixed in >=0.41.0. Unrelated to this change, but if you have appetite for another one, that is the next one I would take.

PR #88, opened 2026-08-09
perf(rag): maintain the persistent BM25 index instead of rebuilding per query

dazwin:
## Problem

`retrieve_hybrid` — and the graph-anchored and iterative variants — rebuilt the BM25 keyword index in RAM from every stored chunk on every query. That is O(corpus) work per query, which costs tens of seconds on a 2,800-chunk knowledge base.

A persistent tantivy index at `data_dir/tantivy/` already existed, with incremental add/delete support. Nothing ever wrote to or read from it.

## Change

Wires the existing persistent index up end to end.

**`kwaai-rag`**
- `BM25Index::open` is now tenant-scoped (`data_dir/tantivy/<tenant>/`). Knowledge bases can share a data dir, and a shared index would leak chunks across KBs.
- `BM25Index::open_backfilled` compares the index doc count against the new `MetaStore::chunk_count` and rebuilds from the metadata store on drift. This self-heals KBs ingested before this change, and any KB left inconsistent by a failed best-effort write.
- `IngestConfig` / `RetrieveConfig` gain an optional `Arc<BM25Index>`. Ingest adds new chunks incrementally; retrieval searches the prebuilt index and skips the full-corpus load entirely. Passing `None` preserves the previous in-RAM behaviour for one-shot callers, so this is additive for existing library users.

**`kwaai-cli`**
- `ingest` / `query` / `chat` / `eval` / `sync` and the HTTP API open the index once and pass it through; delete paths prune it alongside the metadata.

## Notes for review

Two deliberate trade-offs worth a look:

- **Ingest writes are best-effort.** A busy tantivy writer lock warns and lets the index drift rather than failing the ingest. The rationale is that a dropped keyword-index write should not lose a user's ingested document, and `open_backfilled` repairs the drift on next open. If you would rather ingest fail loudly, that is a one-line change.
- **Backfill triggers on doc-count drift alone**, not a content hash. Cheap to check on every open, but it will not detect a same-count divergence. That seemed an acceptable bound for a keyword index that is already a recall aid rather than a source of truth.

## Testing

- `cargo test -p kwaai-rag` — 210 passed, 0 failed.
- `cargo check -p kwaai-cli --features storage,rag` — clean.
- `cargo fmt --check -p kwaai-rag` — clean.

The performance claim above is from the originating profile on a 2,800-chunk KB; I have not re-run a formal before/after benchmark as part of this PR.

### DIFF
--- a/core/crates/kwaai-cli/src/rag_api.rs
+++ b/core/crates/kwaai-cli/src/rag_api.rs
@@ -12,6 +12,7 @@ use tokio::sync::Mutex;
 use uuid::Uuid;
 
 use kwaai_rag::{
+    bm25::BM25Index,
     embedder::EmbedClient,
     graph::GraphStore,
     ingestion::{ingest_text, IngestConfig},
@@ -44,6 +45,9 @@ struct RagState {
     client: Option<Arc<Mutex<P2PClient>>>,
     embed: EmbedClient,
     meta: Arc<MetaStore>,
+    /// Persistent keyword index; None means fall back to per-query in-RAM
+    /// builds (never wrong, just slower).
+    bm25: Option<Arc<BM25Index>>,
     inference_url: String,
     top_k: usize,
     http: reqwest::Client,
@@ -97,6 +101,14 @@ pub async fn run(port: u16, inference_url: String, top_k: usize, kb: String) -> 
         };
         let meta = Arc::new(MetaStore::open(&rag.data_dir(), tenant_id)?);
 
+        let bm25 = match BM25Index::open_backfilled(&rag.data_dir(), &meta) {
+            Ok(idx) => Some(Arc::new(idx)),
+            Err(e) => {
+                tracing::warn!("BM25 index unavailable ({e:#}); using in-RAM fallback");
+                None
+            }
+        };
+
         let doc_context: Option<String> = GraphStore::open(&rag.data_dir(), tenant_id)
             .ok()
             .and_then(|g| {
@@ -120,6 +132,7 @@ pub async fn run(port: u16, inference_url: String, top_k: usize, kb: String) -> 
             client: p2p_client,
             embed,
             meta,
+            bm25,
             inference_url: inference_url.clone(),
             top_k,
             http: reqwest::Client::new(),
@@ -237,6 +250,7 @@ async fn do_chat(state: &RagState, req: ChatRequest) -> Result<serde_json::Value
             hyde_inference_url: None,
             hyde_model: None,
             hyde_alpha: None,
+            bm25: state.bm25.clone(),
             ..Default::default()
         };
 
@@ -406,6 +420,7 @@ async fn do_search(
             hyde_inference_url: None,
             hyde_model: None,
             hyde_alpha: None,
+            bm25: state.bm25.clone(),
             ..Default::default()
         };
         let chunks = match state.storage_url.as_deref() {
@@ -507,6 +522,11 @@ async fn api_delete_doc(
         if ids.is_empty() {
             return (axum::http::StatusCode::NOT_FOUND, "document not found").into_response();
         }
+        // Best-effort: on failure the count drift makes the next
+        // open_backfilled rebuild the index without the deleted doc.
+        if let Some(ref idx) = state.bm25 {
+            let _ = idx.delete_doc(&name);
+        }
         match state.storage_url.as_deref() {
             Some("local") => {
                 if let Some(ref vs) = state.local_vs {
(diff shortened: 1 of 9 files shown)


dazwin:
## Benchmark results

Measured on this branch (`b71a88b`) against a KB built from a real 530k-char PDF technical manual (885 chunks), plus a scaled-up corpus (4,728 chunks) to check how the gap grows.

Both paths are exercised from a single binary via the `RetrieveConfig.bm25` `Some`/`None` split this PR introduces, so old and new are compared on identical code, identical data, and identical queries — 10 natural-language queries, `top_k=20`, 10 iterations each. Timing isolates the keyword-index half of `retrieve_hybrid`; no embedding or vector-search cost is included, since that half is untouched by this change.

### Per-query keyword retrieval

| Corpus | Old (rebuild in RAM) | New (persistent) | Speedup |
|---|---|---|---|
| 885 chunks | 47.19 ms mean / 45.20 ms median | 0.45 ms mean / 0.26 ms median | **106× mean, 175× median** |
| 4,728 chunks | 233.91 ms mean / 226.88 ms median | 0.71 ms mean / 0.64 ms median | **332× mean, 353× median** |

One-time index open on the new path: 2.4 ms (885 chunks), 10.1 ms (4,728 chunks) — paid once per process, not per query.

The old path scales with corpus size as expected (47 ms → 234 ms, ~5× cost for ~5.3× the chunks), while the new path stays essentially flat (0.45 ms → 0.71 ms). That is the O(corpus) → O(1)-per-query change the PR describes, and it is why the speedup ratio itself grows with corpus size.

### Correctness

A speedup from an empty or stale index would be meaningless, so I diffed the two paths' results on the 885-chunk KB:

- **Score sequences are identical** for all 10 queries — the returned BM25 scores match exactly, position for position.
- Set overlap on top-20 is **97.2%**; exact id-order match is 3/10.

The id differences are tie-breaking among chunks with equal scores (identical scores, different internal doc ordering between an in-RAM and an on-disk tantivy index), not a ranking change. This corpus has a lot of near-duplicate boilerplate, which produces many exact score ties. Worth a reviewer's eye if you consider tie-order part of the contract, but retrieval quality is unaffected.

### Caveats

- The "tens of seconds on a 2,800-chunk KB" figure in the original commit message is **not** what I measured — at 4,728 chunks I see ~234 ms per query, not tens of seconds. That original number likely included embedding and vector-search cost, or came from a debug build. The isolated keyword-index cost is the honest number for what this PR changes.
- Single machine (M-series macOS, release build), and another workload was active on the box during the runs — visible in the tail latencies (p-max 9.67 ms on the new path vs a 0.26 ms median). Medians are the more reliable figure here.
- The 4,728-chunk corpus is the same source text ingested four times under different doc names, since the only other PDFs I had available failed to parse. Chunk count and index size are realistic; lexical diversity is not, which if anything favours the *old* path (a smaller vocabulary makes the in-RAM rebuild cheaper than it would be on genuinely varied text).

Benchmark harness is scratch code, not included in this PR — happy to add it under `examples/` if useful.

RezaRassool (approved):
Approving — the design is sound and the write-up is honest about its own trade-offs.

I focused on the two things that could have made this quietly wrong, and both check out:

**Tenant scoping holds under the multi-KB path.** My first concern was `cmd_query`, which loops over `kb_names` — if it had reused a single `MetaStore`, every KB would have resolved to the same tenant's index and the leak this PR exists to prevent would still be there. It doesn't: a fresh `MetaStore::open(&rag_cfg.data_dir(), tenant_id)` is created per iteration and `open_backfilled` takes the tenant from `meta.tenant_id()`. Correct.

**`open` doesn't take a tantivy writer lock** — only the write paths do. That matters more than it might look, because a node running `rag_api` holds an index handle continuously; if `open` had acquired a writer, every concurrent CLI query would have silently dropped to the in-RAM fallback and the speedup would have evaporated in exactly the configuration people actually run. As written, readers don't contend.

On the two trade-offs you flagged:

- **Best-effort ingest writes** — right call. Worst case is a slow query, never a wrong answer, and `open_backfilled` repairs it. Failing an ingest because a keyword index was busy would be a worse bargain.
- **Count-drift backfill missing same-count divergence** — fair bound for a recall aid. Worth noting the narrow case where it bites: a failed `delete_doc` followed by an add of equal chunk count leaves a stale entry indefinitely. Not worth a content hash on every open, but perhaps worth a comment at the call site.

One note for the record: neither of us has benchmarked this. You say so explicitly in the PR, and I didn't re-run it either — so the "tens of seconds" figure remains the originating profile rather than a verified before/after.

PR #91, opened 2026-08-15
build: patched deps via checksum-pinned fetch script (native-p2p 0/7)

dazwin:
> [!NOTE]
> Replaces #79 — same branch content, moved into this repo so the chain can use GitHub stacked PRs (cross-fork stacks are unsupported).

| # | PR | What |
|---|----|------|

---

## What this adds

A build bootstrap step that fetches `multistream-select` 0.13.0 from crates.io and applies a one-rule patch allowing slash-less libp2p protocol IDs, wired in via `[patch.crates-io]`. This is the foundation the rest of the native-p2p stack builds on — without it a Rust node cannot speak hivemind's protocol names at all.

## Why

Hivemind negotiates bare handler names (`DHTProtocol.rpc_store`, `hello`) as libp2p protocol IDs. go-libp2p accepts them; upstream `multistream-select` rejects any name without a leading `/`, in both its `TryFrom` impls. That means a native rust-libp2p node can neither call a hivemind unary handler nor serve one to a Go peer. The restriction is local validation inside rust-libp2p, not part of the wire protocol — go-libp2p negotiates these names on the live network today. Every later PR in this stack depends on the relaxed rule.

## What's in it

- `core/patches/multistream-select.patch` — a 46-line patch relaxing the leading-slash requirement. Nothing is vendored: the repo carries only the patch, and the expanded source is gitignored.
- `core/patches/fetch-multistream-select.sh` — idempotent fetch/verify/apply. Pins version 0.13.0 and a SHA-256 of the crate tarball, extracts to `core/patches/multistream-select`, applies the patch, and writes a stamp file so re-runs are no-ops unless the patch changes.
- `core/patches/README.md` — documents the delta and the rationale.
- `core/Cargo.toml` — `[patch.crates-io] multistream-select = { path = "patches/multistream-select" }`, plus `exclude = ["patches/multistream-select"]` so cargo does not auto-adopt it as a workspace member and pull its dev-deps into the lock.
- **Build entry points, all for the same reason.** Because `[patch.crates-io]` points at a path, cargo cannot even *parse* the workspace until that path exists. So the fetch step is added to `setup.sh`, the eleven GitHub Actions workflows (`ci.yml`, the six `ci-kwaai-*.yml`, `publish-crates.yml`, `release.yml`, `release-riscv.yml`, `test-cuda-cache.yml`), `flake.nix`, and the nix crane/cross builders (`nix/crane.nix`, `nix/cross.nix`).

## Risk

**No default runtime behaviour changes.** This PR adds no application code — nothing in any crate is added, removed or rewired. The only functional delta is that `multistream-select` now permits protocol names it previously rejected locally; no code in the tree yet negotiates such a name (that starts in PR3).

What a reviewer should look at closely:

- **The bootstrap requirement is the real risk, and it is a build-time one.** Anyone running `cargo` in a checkout that has not run `core/patches/fetch-multistream-select.sh` will get a workspace-parse failure, not a helpful error. Please confirm your own build entry point is covered — if there is a CI workflow, container image, or local script not in the list above, it will break.
- The fetch script requires network access to `static.crates.io` at build time. Air-gapped or offline builds will need the directory pre-populated.
- The checksum pin is what makes the fetch trustworthy; verify `SHA256=ea0df8e5…` matches the published crate.

## Test coverage

No new tests. This PR is build plumbing; the patched behaviour is exercised by PR3's tier-08 interop test (`slash-less negotiation against a real go-libp2p daemon`) and by PR1's `slashless_protocol_negotiates`. The fetch script's own correctness is checked by the checksum verification and the idempotency stamp rather than by a test.

## Stacked on

`upstream/main`. This is the base of the 8-PR native-p2p stack.

---

- `cargo test --workspace`: **633 passing, 0 failures.**

---


RezaRassool (approved):
Approve — with one portability item worth closing before the next release tag.

The patch doesn't blanket-remove validation, it **replaces the leading-slash rule with the constraint the framing actually imposes** (non-empty, no embedded newline). That's the right shape. The third hunk goes further and documents a real behavioural trade-off out loud — relaxing proposal-recognition shadows the `ls`-response branch for single-name lists, justified by neither rust-libp2p nor go-libp2p sending `ls` during negotiation. Stating that rather than burying it is exactly right.

The apparatus is sound too: checksum verified *before* extraction, nothing vendored, expanded source gitignored, `exclude` keeping cargo from auto-adopting it as a workspace member, and a stamp keyed on version + patch hash so re-runs are genuinely no-ops.

## The one real risk: the Windows path has never executed

`release.yml` builds Windows, and this PR correctly adds the step there with `run: bash …` (which resolves to Git Bash). But **`release.yml`'s build jobs skip on PRs and run only on tag push** — on this PR's own checks, only `plan` ran. So the fetch script has never run on a Windows runner, and the first place a problem surfaces is the next release tag.

Not hypothetical: the script's first action is `shasum -a 256 "$PATCH"`, *before* any download, so a missing tool fails immediately on every Windows build.

There's a concrete smell here — **this repo's own workflows use `sha256sum`** (`release.yml:524`, `release-riscv.yml:51`), not `shasum`. This introduces a different tool into paths that include Windows.

Neither is universal, though: macOS has `shasum` but not `sha256sum`; Linux has `sha256sum`. So the fix is a fallback rather than a swap:

```bash
sha256() {
    if command -v shasum >/dev/null 2>&1; then shasum -a 256 "$@"
    else sha256sum "$@"; fi
}
```

Three lines and the question disappears. Field note: when metro-win built this branch, Git Bash had to be specified explicitly — this is a live concern, not a theoretical one.

## Minor

- **15 inline copies** of the step across 11 workflows. A composite action (`.github/actions/fetch-patched-deps`) would make it one definition — and more to the point, nothing today stops the *12th* entry point from forgetting it.
- **Network required at build time** on every fresh checkout. The checksum covers integrity, not availability — worth a README line for offline/air-gapped builds.
- **Confusing failure mode**: since `[patch.crates-io]` points at a path that must exist before cargo can parse the workspace, a fresh clone running `cargo build` gets a manifest error rather than "run the fetch script". Inherent to the approach, but a note in the root README would save someone an hour.

The README already covers rebasing on libp2p bumps and the intent to drop this if upstream relaxes the check, which is the maintenance question I'd otherwise have raised.

**Recommendation:** merge. Fold the `sha256` fallback in now or file it as a follow-up before the next tag — this is the foundation of the whole stack and holding it up over a three-line tweak would be the wrong trade.

PR #94, opened 2026-08-15
feat(cli): native DHT bootstrap seed

dazwin:
Phase 1 of bootstrap decentralisation: a Rust-native DHT bootstrap node, expressed as configuration rather than as a mode.

Rebased onto `main` now that #86 has landed — `git rebase --onto origin/main 5168021`, as suggested in review.

## What this adds

A bootstrap node is not a separate program, a subcommand or a mode. It is an ordinary node whose config says *publish nothing, dial nobody* — started the same way as any other node:

```bash
kwaainet start          # or `run-node` for a container entrypoint (foreground)
```

It replaces `python -m petals.cli.run_dht`: an in-process libp2p swarm that answers `rpc_ping` / `rpc_store` / `rpc_find`, runs Kademlia in server mode, and offers a circuit relay hop service so NATed nodes can reserve through it.

Two new config keys carry it, so one config shape describes both an ordinary node and a bootstrap:

| key | default | what it does |
|---|---|---|
| `announce_self` | `true` | whether the node publishes records *of its own* — and, symmetrically, whether it writes the `state = -1` tombstone on exit |
| `enable_upnp` | `true` | opt out of UPnP port mapping where there is no gateway to ask |

`enable_upnp` is independent of the other — nothing derives or overrides it, and a bootstrap behind a home router may well want it left on. It is defined here only so this branch builds standalone; **`feat/upnp-toggle` (#77) owns it**, and the overlap resolves in that branch's favour when both land.

[`docs/BOOTSTRAP.md`](../blob/feat/bootstrap-serve/docs/BOOTSTRAP.md) documents the full set.

## `dht_server` has been dropped

Earlier revisions of this branch added a third key, `dht_server`. **Review found it did not do what it claimed** — it documented itself as governing `rpc_ping`/`rpc_store`/`rpc_find` while reaching only libp2p-kad's server mode. That finding was correct.

I first made the claim true by gating the handlers. On working through the consequences, removing the key is the better answer, because neither half survives on its own terms:

- **Kademlia does not need a key.** The only case for declining to serve kad queries is a node too NATed to answer them — and auto-mode already reaches that state *from evidence*: client until the reachability machine confirms an external address, server after. The key let an operator assert what the swarm already observes, and get it wrong. Its original rationale (in the commit that introduced it) was guarding against a future change to the ordinary-node default; that default has since moved to `true`, so the guard now tracks the very thing it was guarding.

- **A serving switch would be half a control.** Unregistering the handlers is invisible to placement. With `decentralized_dht` on, candidates are drawn from the routing table with **no filter on who serves**, so a non-serving node is still selected among the *k* nearest, its store fails, and replication silently degrades toward the shortfall warning. Hivemind avoids exactly this — a `client_mode` node sends an empty `node_info` so peers never route to it — and we have no equivalent. Adding one is real work and not this branch's scope.

The underlying concern is real and unchanged: **there is no record validation**, so a served store accepts any key from any peer that can reach it. Worth being precise, since I was loose about this earlier — there is no authentication layer at all. `RequestAuthInfo` is a hivemind-compatible wire shape that nothing reads; no signature is checked because none is verified.

Record validators are the control for that. A config key is not a substitute, and shipping one as though it were is worse than its absence — *"a silently ineffective security control is worse than an absent one"* is the objection that started this, and it applies to a half-working switch too.

So: kad server mode stays hardcoded `true` on every node, matching p2pd's `-b` and what `main` already does. Serving is unconditional, as on `main`. `BOOTSTRAP.md` records why the key was dropped, so it is not re-proposed without the routing-table half.

## One change to existing behaviour

`run_node` fell back to the public Petals bootstraps whenever `initial_peers` was empty. That is right for an ordinary node ("not configured") and wrong for a bootstrap ("dials nobody") — a live run caught one opening connections to the public network on startup. The fallback is now gated on `announce_self`, so an empty list means what it says for a bootstrap and is unchanged for everything else.

Both keys default to today's behaviour, so no existing node changes. A config predating them still deserialises to ordinary-node defaults rather than `bool::default()`, which would silently mute every node on the network; there is a unit test for exactly that.

## Tests

Tier 14 (`14_bootstrap_seed.rs`) assembles the swarm inline and asserts the wire contract as a peer sees it — all five behaviours pass unchanged. 95 `kwaainet` unit tests green; `cargo fmt` and `clippy` clean.

## Follow-ups raised in review

Both are pre-existing on `main` rather than introduced by this branch, and neither blocks this merge:

- **Record validators.** No authentication on `rpc_store`. The exposure is not waiting on a bootstrap — it exists today for every native node, bounded only by `DEFAULT_STORAGE_SIZE` and `MAX_STORE_KEYS_PER_REQUEST`. With `dht_server` dropped, this is now the *only* answer to unvalidated writes, which is the right place for it to sit. It is also a stated prerequisite for enabling `decentralized_dht`.
- **The eviction cliff.** `LocalStorage::enforce_capacity` picks its victim with `.iter().min_by(…)` under the write lock against a `DEFAULT_STORAGE_SIZE` of 1,048,576. Below capacity it never runs; at capacity every insert scans ~1M entries holding the lock. An ordinary node will not fill a million entries — a bootstrap is exactly the deployment that will, so the expiration-ordered index should be a hard prerequisite for running one in anger, not a tidy-up.


core/crates/kwaai-cli/src/config.rs
```diff
@@ -148,6 +148,56 @@ pub struct KwaaiNetConfig {
...
+    /// startup, none on the timer, none on a reachability change or a SIGHUP,
+    /// and correspondingly no unannounce tombstone at shutdown, because there
+    /// is nothing to retract. This is what makes a bootstrap a bootstrap: it
+    /// stores and serves everyone else's records while appearing on the map as
+    /// nothing at all. A bootstrap that announced itself would show up as an
+    /// inference node offering zero blocks.
+    ///
+    /// This is about *publishing*, not about *serving*. The DHT server role —
+    /// answering rpc_ping/rpc_store/rpc_find and holding other peers' records —
+    /// is untouched by this flag; see `dht_server` for that.
```
dazwin: Comment block is too verbose. Shorten to a note about publishing, not serving. Do not comment the default (it's already stated in code). Reduce the bootstrap paragraph to two sentences - a boostrap node should set this false, otherwise it would show up as an inference node etc.

core/crates/kwaai-cli/src/config.rs
```diff
@@ -148,6 +148,56 @@ pub struct KwaaiNetConfig {
...
+    ///
+    /// Defaults to false, which is the ordinary node behaviour: `node_native`
+    /// already puts a regular node into server mode to match the p2pd path's
+    /// `-b` flag, and auto-detection governs the rest.
+    ///
+    /// Set true and the node answers Kademlia queries from t=0, before any
+    /// external address has been confirmed. A bootstrap node needs exactly
+    /// this: it exists to answer queries, and the auto-detected path would
+    /// leave it in client mode — refusing the very lookups it was deployed to
+    /// serve — until an AutoNAT round confirmed an address for it.
```
dazwin: Similarly, don't restate what's already in code. Reduce to a couple of sentences: what it does, and the recommendation for bootstrap nodes

core/crates/kwaai-cli/src/main.rs
```diff
@@ -118,7 +118,10 @@ async fn main() -> Result<()> {
 
     // Spawn a background update check that runs concurrently with the command.
     // Uses a 24-hour on-disk cache so it only hits the network once per day.
-    // Skipped for `update` (redundant) and `run-node` (internal daemon process).
+    // Skipped for `update` (redundant) and `run-node` (internal daemon process,
+    // and the entry a bootstrap node's container entrypoint uses — no operator
+    // watches its stdout for an upgrade hint, and its host may have no route to
+    // the update endpoint at all).
```
dazwin: Remove this change completely - there's no code change

core/crates/kwaai-cli/src/node_native.rs
```diff
@@ -245,6 +259,7 @@ impl NativeNode {
             handle,
             peer_id,
             storage,
+            announce_self: config.announce_self,
```
dazwin: Gate the call to announce here. Calling announce with a flag that no-ops is redundant.

core/crates/kwaai-cli/src/node_native.rs
```diff
@@ -278,12 +308,22 @@ impl NativeNode {
...
+    /// retracts an announce, so a node that published nothing has nothing to
+    /// retract. Without this, a bootstrap's one and only DHT record would be a
+    /// `state = -1` for a node the network never heard of.
     pub async fn unannounce(
         &self,
         ctx: &AnnounceContext<'_>,
         server_info: &DHTServerInfo,
         bootstrap_peers: &[String],
     ) {
+        if !self.announce_self {
```
dazwin: Similarly to the above comment, gate the call to unannounce()

core/crates/kwaai-cli/src/node_native.rs
```diff
@@ -346,7 +386,17 @@ pub async fn run_native_node(
...
 
     // ── Announce inputs ────────────────────────────────────────────────────
-    info!("[2/4] Preparing the DHT announcement...");
+    // Everything from here to the initial announce describes a record this node
+    // is about to publish. With `announce_self = false` no record is ever
+    // built, so the measurements behind it are skipped rather than computed and
+    // discarded — the bandwidth probe in particular is a real network round
+    // trip, and a bootstrap has no model to measure a throughput for.
+    let announce_self = config.announce_self;
+    if announce_self {
```
dazwin: Gate the entire announce setup in a single conditional rather than multiple announce_self blocks.

core/crates/kwaai-cli/src/node_native.rs
```diff
@@ -357,21 +407,42 @@ pub async fn run_native_node(
...
     // circuit, leaving the record stale until the 300 s tick.
     let mut last_announced_epoch = startup_state.epoch;
-    if let Some(addr) = configured_announce_addr(config) {
-        info!("  Announce addr: {addr} (declared)");
-    } else {
-        info!("  Announce addr: discovered (autonat, upnp or identify consensus)");
+    if announce_self {
+        if let Some(addr) = configured_announce_addr(config) {
+            info!("  Announce addr: {addr} (declared)");
+        } else {
```
dazwin: No need to log what isn't being done - remove

core/crates/kwaai-cli/src/node_native.rs
```diff
@@ -393,21 +464,36 @@ pub async fn run_native_node(
...
-    info!("[3/4] Announcing to DHT...");
-    if let Err(e) = node.announce(&ctx, &server_info, bootstrap_peers).await {
-        warn!("Initial announce failed: {e:#} — will retry at the 300 s tick");
+    if announce_self {
+        info!("[3/4] Announcing to DHT...");
+        if let Err(e) = node.announce(&ctx, &server_info, bootstrap_peers).await {
+            warn!("Initial announce failed: {e:#} — will retry at the 300 s tick");
+        }
+    } else {
+        info!("[3/4] Skipping the announcement — announce_self = false");
```
dazwin: No need to log what isn't being done - remove this line

core/crates/kwaai-cli/src/node_native.rs
```diff
@@ -393,21 +464,36 @@ pub async fn run_native_node(
...
+        info!("   Name    : {}", public_name);
+        info!("   Model   : {}", config.model);
+        info!(
+            "   Blocks  : {}–{}",
+            config.start_block,
+            config.effective_end_block()
+        );
+        info!("   Map     : https://map.kwaai.ai");
+    } else {
+        // A bootstrap's summary names what it does — serve — and omits the model and
```
dazwin: More spurious logging - remove except the peer id (unless this is already logged, then it can be removed as well)

core/crates/kwaai-cli/src/node_native.rs
```diff
@@ -393,21 +464,36 @@ pub async fn run_native_node(
...
+        // block range, which it has neither of. No model is loaded and no shard
+        // is spawned on this path: those are started by `shard serve`, never by
+        // the node runner, so the preset reaches no download code at all.
+        info!("[4/4] ✅ KwaaiNet bootstrap node running (native p2p)");
+        info!("   Peer ID : {}", peer_id.to_base58());
+        info!("   Serving : the hivemind DHT (rpc_ping/rpc_store/rpc_find)");
+        info!("   Announcing: nothing");
+    }
```
dazwin, line 444: Honestly, this entire block looks like it should be gated and in a select 'announce' function - something typically only non-bootstrap clients would run.

core/crates/kwaai-cli/src/node_native.rs
```diff
@@ -557,8 +655,12 @@ pub async fn run_native_node(
...
     // Tombstone before tearing down the swarm, so the map drops us immediately
-    // rather than waiting out the TTL.
-    info!("Unannouncing from DHT...");
+    // rather than waiting out the TTL. `unannounce` is itself gated on
+    // `announce_self`; the log line is gated here too so a bootstrap's shutdown does
+    // not claim a retraction it correctly never sends.
+    if announce_self {
+        info!("Unannouncing from DHT...");
+    }
     node.unannounce(&ctx, &server_info, bootstrap_peers).await;
```
dazwin: Here it anti-patterned call to unannounce. A simple gate on announce_self works better

core/crates/kwaai-cli/src/node_native.rs
```diff
@@ -187,8 +189,16 @@ impl NativeNode {
         // Dials the configured peers and seeds Kademlia. Non-fatal: a node with
         // no reachable bootstrap still serves its own DHT and control socket,
         // and the announce loop retries every 300 s.
+        // An empty list is normal for a bootstrap and a misconfiguration for a
+        // node, so the two say different things. Either way it is not an error
+        // and there is no retry loop: nothing is dialled, and `handle.bootstrap`
+        // is simply never called.
         if bootstrap_peers.is_empty() {
-            warn!("No bootstrap peers configured — this node will not join the network");
+            if config.announce_self {
```
dazwin, line 196: Combine these two conditionals to produce the warning. The info isn't needed

core/crates/kwaai-cli/src/node_native.rs
```diff
@@ -278,6 +289,10 @@ impl NativeNode {
 
     /// Write the `state = -1` tombstone so the map drops this node immediately
     /// rather than waiting out the 360 s TTL.
+    ///
+    /// Only called when the node announced in the first place: a tombstone
+    /// retracts an announce, so a node that published nothing has nothing to
+    /// retract.
```
dazwin: Avoid comments that presume usage - remove this one

core/crates/kwaai-cli/src/node_native.rs
dazwin: Remove the comment below as well (discovered etc.)

core/crates/kwaai-cli/src/node_native.rs
```diff
@@ -357,21 +373,30 @@ pub async fn run_native_node(
...
     let repository = crate::node::effective_repository(config);
-    info!("  DHT prefix:  {}", prefix);
-    info!("  Repository:  {}", repository);
-    info!("  Using relay: {}", using_relay);
 
-    let vpk_info = crate::node::initial_vpk_info(config, public_name).await;
+    // One gate for every announce input. A bootstrap computes none of them:
+    // `measure_download_bps_for` runs a live download, and `initial_vpk_info`
+    // polls a VPK endpoint — both pointless for a node that publishes nothing.
+    let (dl_bps, throughput, vpk_info) = if announce_self {
```
dazwin: These conditionals are more clearly expressed when the small block comes first (lines like 0.0, 0.0, None are left dangling, especially on larger conditionals. Invert it

core/crates/kwaai-cli/src/node_native.rs
```diff
@@ -393,21 +418,29 @@ pub async fn run_native_node(
...
+        info!("   Model   : {}", config.model);
+        info!(
+            "   Blocks  : {}–{}",
+            config.start_block,
+            config.effective_end_block()
+        );
+        info!("   Map     : https://map.kwaai.ai");
+    } else {
+        // A bootstrap has no model or block range to name.
+        info!("[4/4] ✅ KwaaiNet bootstrap node running (native p2p)");
```
dazwin: I'm pretty sure both these native p2p logs are an assumption and will print when running the p2pd daemon as well - remove them. You can also move the 4/4 block and peer id log outside the conditional, thus removing the smaller branch

RezaRassool:
Read through ahead of #86 landing. The shape is right — a bootstrap as *configuration* rather than a mode is a much better factoring than a separate subcommand, and "serve the DHT, announce nothing, dial nobody" is a clean way to say it.

Two observations, one of which raises the urgency of something I flagged on #83.

## `dht_server: true` is the correct default, and it made me check something

Defaulting to `true` preserves today's behaviour rather than changing it, and adds the off-switch that did not exist before — strictly an improvement, no note needed on the PR itself.

But checking *why* it preserves behaviour turned up something worth stating plainly: on `main` right now, `node_native.rs:144` calls `spawn_dht_service` **unconditionally**. Every native node already serves hivemind DHT writes.

When I reviewed #83 I framed the missing record validators as a bootstrap-only prerequisite — "a Rust bootstrap must not ship without validators", per the migration doc. That framing was too narrow. The exposure is not waiting on a bootstrap; it exists on `main` today for every native node, bounded only by `DEFAULT_STORAGE_SIZE` capacity and `MAX_STORE_KEYS_PER_REQUEST`. This PR does not introduce that and in fact gives operators the first way to opt out of it.

So: nothing to change here, but the validator work is more urgent than my #83 review implied, and this PR is the natural place for the argument to get made.

## The eviction cliff matters more for a bootstrap

`LocalStorage::enforce_capacity` picks its victim with `.iter().min_by(…)` under the write lock, against a `DEFAULT_STORAGE_SIZE` of **1,048,576**. Below capacity it never runs. At capacity, every insert scans ~1M entries holding the lock — not gradual degradation but a cliff, and precisely when the node is busiest.

An ordinary node will not fill a million entries. A bootstrap is exactly the deployment that will. I would treat the expiration-ordered index as a hard prerequisite for running one of these in anger, not a tidy-up — worth saying in the PR body so whoever deploys the first Rust bootstrap knows.

Neither point blocks the merge. Both are about what happens after someone turns it on.

Will review properly once #86 lands and this retargets to `main`.

RezaRassool (changes requested):
Retargeted to `main` now that #86 has landed. The design is right and most of this is ready — one finding I think should be fixed before merge, because it is a config key that does not do what its own doc comment says.

## `dht_server: false` does not stop the node serving DHT writes

`config.rs:173` documents the key as:

> rpc_ping/rpc_store/rpc_find is governed by `dht_server`.

It is not. `node_native.rs:182` calls `spawn_dht_service(handle, storage)` **unconditionally** — I checked the enclosing scope for an outer guard and there is none. `config.dht_server` flows only to `NetworkConfig.dht_server` (`node_native.rs:143`), which `behaviour.rs:131` uses to select libp2p-kad **server mode**. Those are two different things:

| | governed by `dht_server`? |
|---|---|
| libp2p-kad server mode | yes (`behaviour.rs:131`) |
| hivemind `rpc_store` / `rpc_find` / `rpc_ping` handlers | **no — always registered** |

So an operator who sets `dht_server: false` to mean "do not serve the DHT" still accepts `rpc_store` from any peer. Given record validators are not implemented, that key is the only lever anyone has for declining unauthenticated writes, and today it does not pull.

`043071b` ("dht_server actually controls Kademlia server mode") fixed the kad half. This is the other half, and it is the half the doc comment describes.

Fix is presumably a guard around the block at `node_native.rs:180-186` — worth also keeping the `DHTStorage::new` allocation inside it, since a node that serves nothing needs no store.

## The rest reads well

The evolution through the 14 commits is the good kind — it starts as `kwaainet bootstrap serve` and gets refactored into configuration (`292f13c`, `f3bd07a`), then the vocabulary is cleaned up from "seed" to "bootstrap node". A bootstrap being an ordinary node with `announce_self: false, dht_server: true, dial nobody` is a much better factoring than a mode or a subcommand, and it means the bootstrap path gets exercised by the same code everything else uses.

`7a14d4d` serialising the tests that drive `KWAAINET_HOME` is the right call — env-var tests that run concurrently are a classic source of flakes, and this repo has already been bitten by timing flakes this week.

Tier 14 covering "the bootstrap seed as a peer sees it" is the correct framing: a bootstrap's whole contract is what other nodes observe, not its internal state.

`enable_upnp` absorbed cleanly against #77 after the rebase — one declaration, one setter, no duplication.

## Rebase needed

Retargeting alone leaves the diff at +9403/−448 across 50 files, because the branch still carries #86's commits that landed under a different SHA in the squash. Dropping them gives the true change — **8 files, +990/−54**:

```bash
git rebase --onto origin/main 5168021
```

Clean, verified locally; that is what I reviewed above.

Happy to approve once `dht_server` gates the handlers it claims to.

dazwin:
Both addressed — thanks, the finding was correct.

**`dht_server` now gates the handlers** (`712fd12b`). You were right that there was no outer guard; `spawn_dht_service` ran unconditionally and the key reached only kad's server mode. The gate is one `if` at the call site in `node_native.rs`, so `kwaai-p2p` is unchanged.

I checked the Python before choosing where to cut, because "stop serving" could plausibly have meant several things. Hivemind draws exactly this seam with `client_mode`: `DHTProtocol.create` skips the handler registration *and* sets the daemon's `dht_mode` to client from the one flag — so the two halves have never been separable there either, which is a good sign the key should own both. What it pointedly does **not** do is touch the outbound path, and it keeps constructing local storage regardless.

So the fix follows that shape — turning off *serving* must not turn off *participating*. A `dht_server: false` node still dials `rpc_store`/`rpc_find` on other peers, still announces, still appears on the map, and still keeps a local store for its own records.

One place I went the other way from your suggestion: you proposed moving the `DHTStorage::new` allocation inside the guard, since a node that serves nothing needs no store. It turns out it does — `announce`/`unannounce` write this node's own records into that same storage, and `announce_self` is independent of `dht_server`, so `announce_self: true, dht_server: false` is a legal config that would have lost its own records. What I did move inside the guard is the **maintenance loop**, which is genuinely serving-only: its two jobs are the neighbour list in an `rpc_find` response and the periodic expiry sweep, and a node answering neither never reads the first. (I initially kept the sweep on the theory that a re-announcing node would grow its store unbounded — that was wrong. Re-announce rewrites the same three keys under the node's own subkey and `store_subkey` replaces in place, so it sits at three entries.)

**Test:** tier 14 gains `a_non_serving_node_refuses_store_and_find_at_negotiation`, the inverse of the five already there. It asserts on the error *arm* rather than on failure — a dial that never connected would also be an `Err` and would prove nothing about serving — so it dials first and checks the peer ID before calling. Verified it fails against the ungated code, where it returned a well-formed `StoreResponse` with `store_ok = [true]` rather than refusing.

**Rebase:** done as you suggested, `git rebase --onto origin/main 5168021`. Clean, and the diff is the 8 files / +990/−54 you reviewed. All 6 tier-14 tests and 95 unit tests green; fmt and clippy clean.

**On your two follow-ups** — I have written both into the PR body rather than fixing them here, since both are pre-existing on `main` rather than introduced by this branch, and you noted neither blocks the merge. Agreed on the framing for each: the validator exposure is every native node's today, not a bootstrap prerequisite, which is exactly why `dht_server` needed to actually work — it is the only lever anyone has in the meantime. And the eviction cliff is a hard prerequisite for the first real bootstrap deployment rather than a tidy-up, given a bootstrap is the one deployment that will reach 1M entries. Happy to take either as its own PR.

dazwin:
Update: I have gone the other way on this — **`dht_server` is dropped from the branch entirely** rather than fixed.

Your finding was right, and I first fixed it as you suggested, by gating the handlers. Working through what the key would then be for, neither half held up:

**The kad half does not need a key.** The only case for declining to serve kad queries is a node too NATed to answer them — and auto-mode already reaches that state from evidence rather than from a config guess: client until the reachability machine confirms an external address, server after (`apply_reachability_effects` in `service.rs` is the only thing that adds one). The key let an operator assert what the swarm already observes, and get it wrong. Its original rationale was guarding against a future change to the ordinary-node default; that default moved to `true` in `41dcf61b`, so the guard now tracks the thing it was guarding.

**The serving half would be half a control.** This is the part that decided it. Unregistering the handlers is invisible to placement: with `decentralized_dht` on, `gather_candidates` draws from the routing table with no filter on who serves, so a non-serving node is still selected among the `k` nearest, its store fails, and replication silently degrades into the shortfall warning. Hivemind avoids exactly this — `client_mode` sends an empty `node_info` so peers never add it to their routing tables — and we have no equivalent. Building one is real work, and not what this branch is for.

Which lands on your own objection: *a silently ineffective security control is worse than an absent one*. That applies to a switch that works locally but leaves the network still routing to you. Shipping it would invite operators to rely on it.

I should also correct my own language from earlier in this thread. I kept writing "unauthenticated writes," which implies a check being skipped. There is no authentication layer at all — `RequestAuthInfo` is a hivemind-compatible wire shape that nothing reads, and `server.rs` says so up front. The accurate statement is simply that any peer able to open a connection can write any key. That makes validators the control, and a config key not a substitute for one.

So the branch now adds **two** keys, not three: `announce_self` and `enable_upnp`. kad server mode stays hardcoded `true` on every node, matching p2pd’s `-b` and what `main` already does; serving is unconditional, also as on `main`. `BOOTSTRAP.md` gains a short section recording why the key was dropped, so it is not re-proposed without the routing-table half.

Tier 14 is back to its original five tests, all passing (the refusal test went with the gate). 95 unit tests green, fmt and clippy clean. Diff is 8 files, +1005/−53.

Net effect on your two open items: 2.2 is resolved by removal rather than by fix, and 2.1 gets slightly more weight — with no interim switch, validators are now the only answer to unvalidated writes. I think that is the honest place for it to sit rather than behind a key that half-works. Happy to be argued out of it if you would rather have the partial control in the meantime.

RezaRassool:
Agreed on the removal — and the second reason is the one that carries it, so I want to separate the two, because the first is stated in a way the code does not support and it is now written into `BOOTSTRAP.md` where it will outlive this thread.

## The serving-switch argument is correct, and I verified it

`gather_candidates` (`announce.rs:620`) merges known peers, configured peers and cached peers, and the "known" set is the kad routing table plus live connections, address-filtered — no filter anywhere on whether a peer serves the hivemind handlers. A non-serving node is still selected among the *k* nearest, its store fails, and the loss surfaces only as `total_shortfall` (`announce.rs:796`, computed at `placement.rs:250` as `k.saturating_sub(stored_on.len())`). Degradation without attribution: the warning does not say *which* peers refused or *why*, so the operator who flipped the switch and the operator seeing thin replication are not connected.

That is enough on its own. A control whose local effect is real but whose network effect is nil is exactly what I objected to on #83, and you are right that it applies to a switch as much as to a security control.

## The kad auto-mode argument does not apply to any kwaainet node

> kad's auto-mode reaches that state on its own — client until the reachability machine confirms an external address, server after

True of `kwaai-p2p` as a library; not true of anything this repo ships. `behaviour.rs:131` only leaves auto-mode on in the `else` branch, and `node_native.rs:149` hardcodes `dht_server: true` — as `main` already does at `node_native.rs:142`. So the auto-mode comment at `behaviour.rs:136-137` describes a path no kwaainet node takes, and the reachability machine never gets to demote anyone.

This does not change the outcome, but it inverts the reason. It is not "the key is unnecessary because auto-mode already does this from evidence" — it is "**every kwaainet node serves kad unconditionally, with no evidence-driven fallback at all**, and the key would have been the only way to opt out." Which strengthens your case rather than weakening it: the one lever that existed still wouldn't have worked, for the reason above.

Worth fixing in `BOOTSTRAP.md:105-109` specifically because that section exists to stop the key being re-proposed. As written, the next person to read it comes away believing there is an automatic client/server transition protecting NATed nodes, and there isn't one. Same correction applies to the `behaviour.rs` comment, though that is pre-existing and not this branch's problem.

## Rest of the branch checks out

- Two keys, not three — `announce_self` (`config.rs:178`) and `enable_upnp` (`:161`); no `dht_server` anywhere in `KwaaiNetConfig`.
- Behaviour identical to `main`: serving unconditional, kad server mode pinned `true`.
- Tier 14 back to five tests — `a_client_stores_and_finds_a_subkeyed_record_on_the_seed`, `the_seed_announces_nothing_of_its_own`, `a_later_expiration_replaces_and_a_store_never_deletes`, `the_seed_serves_many_peers_under_one_key`, `the_seed_keeps_the_identity_key_files_peer_id`.
- 8 files, +1005/−53, as described.

Thanks also for correcting "unauthenticated writes" — you are right that it implies a check being skipped, and "any peer able to open a connection can write any key" is the accurate statement. I will carry that wording into the security doc, which has the same slip.

## On 2.1

Agreed, and agreed it is the honest place for it. Validators being the only answer is a clearer position than a key that half-works, and I would rather the gap be visible than papered over. Not a blocker for this branch.

Fix the `BOOTSTRAP.md` paragraph and I am happy to see this land.

core/crates/kwaai-cli/src/node_native.rs
```diff
@@ -138,7 +138,14 @@ impl NativeNode {
...
             // what makes a node answer DHT queries rather than only issuing
-            // them. `dht_server` is the native equivalent.
+            // them. True on every node here for the same reason.
+            //
+            // Not a config key. The only case for turning it off is a node too
+            // NATed to serve queries, and kad's auto-mode already reaches that
+            // state on its own — it stays a client until the reachability
+            // machine confirms an external address (see `apply_reachability_
+            // effects` in kwaai-p2p's service.rs). A key would let an operator
+            // assert what the swarm already observes, and get it wrong.
```
dazwin: This comment is no longer needed - no change to dht_server now

core/crates/kwaai-cli/src/node_native.rs
```diff
@@ -346,7 +361,14 @@ pub async fn run_native_node(
     let peer_id = node.peer_id;
 
     // ── Announce inputs ────────────────────────────────────────────────────
-    info!("[2/4] Preparing the DHT announcement...");
+    // Everything from here to the initial announce describes a record this node
+    // is about to publish, so a bootstrap skips the lot: the bandwidth probe is
+    // a real network round trip, and it has no model to measure a throughput
+    // for.
+    let announce_self = config.announce_self;
+    if announce_self {
```
dazwin: If announce_self is not mutated, why copy? Looking at this variable later hides that it's configuration

dazwin:
All three addressed in `908e0139`.

**The kad auto-mode claim was wrong, and you traced it correctly.** `behaviour.rs:131` only leaves auto-mode on in the `else` branch, and both this branch and `main` pin `dht_server: true` — so the auto-mode comment at `behaviour.rs:136-137` describes a path no kwaainet node takes, and the reachability machine never gets to demote anyone. I had reasoned from what `kwaai-p2p` can do rather than from what this repo actually does.

You are right that it inverts the reason and strengthens the case: kad serving is unconditional too, with **no** evidence-driven fallback, so the dropped key would have been the only way to opt out of either half — and it still would not have worked, for the placement reason.

`BOOTSTRAP.md` now leads with the argument that holds and states the kad situation plainly, including your point about attribution — the shortfall names neither the peers that refused nor the reason, so the operator who flipped the switch and the operator seeing thin replication are not connected. That was the sharper version of what I had written and I have taken it.

The same wrong claim was in the `node_native.rs` comment. With the key gone there is no change to `dht_server` left to explain, so that reverts to `main`’s wording — the line is now byte-identical to `main`.

Agreed the `behaviour.rs` comment has the same problem and is pre-existing; leaving it for a separate change rather than widening this branch.

**Two of my own review notes, also fixed here:** `announce_self` was copied into a never-mutated local, which reads like state and hides that it is configuration — now read from `config` at each site.

165 unit tests green on the integration branch, tier 14 (5) and tier 15 (4) passing, fmt and clippy clean. Branch is 8 files, and the only `node_native.rs` change left is the announce gating, not the DHT setup.

RezaRassool:
@dazwin — nudging this one: it is 7 days old today, which is the point our own process calls overdue. Two things are holding it, and I have written the fix for the first so it is not just a restatement of the review.

## 1. The open change request — proposed patch

The finding was that `dht_server: false` does not stop the node serving DHT writes: `spawn_dht_service` runs unconditionally. Here is the fix, applied to your branch and compiled:

```rust
        // Gated on `dht_server` so the key means one coherent thing — "this
        // node is a DHT server" — covering both kad server mode
        // (`behaviour.rs:131`) and the hivemind record handlers. Registering
        // them unconditionally left `dht_server: false` still accepting
        // `rpc_store` from any peer, which is the opposite of what the key
        // says and, with record validators unimplemented, removed the only
        // lever for declining unauthenticated writes.
        let storage = DHTStorage::new(peer_id);
        if config.dht_server {
            tasks.push(
                kwaai_p2p::spawn_dht_service(handle.clone(), storage.clone())
                    .await
                    .context("starting the native DHT service")?,
            );
            info!("Hivemind DHT service registered (rpc_ping/rpc_store/rpc_find)");
        } else {
            info!(
                "DHT serving disabled (dht_server = false) — not answering \
                 rpc_ping/rpc_store/rpc_find"
            );
        }
```

**One correction to the review.** It suggested moving the `DHTStorage::new` allocation inside the guard too. That does not work as written: `storage` is a struct field, and the announce path writes to it (`self.storage.handle_store`, so the node serves what it publishes). Moving it would make it `Option<DHTStorage>` and ripple to both call sites, for an empty map's worth of memory. Better left where it is.

**Worth confirming the semantics you intend**, because two doc comments in the PR disagree:

| Comment | Claims `dht_server` governs |
|---|---|
| `announce_self` doc | "answering rpc_ping/rpc_store/rpc_find is governed by `dht_server`" |
| `dht_server` doc | forcing kad **server mode** vs auto-mode |

The patch above resolves it in favour of the first, giving the key a single meaning: *this node is a DHT server*, covering both. Default is `true`, so no existing node changes behaviour — only someone who explicitly set `false`, and that is exactly the person expecting not to serve.

The alternative is to split them — keep `dht_server` for kad mode and add something like `dht_serve_records` for the handlers. That is more honest if a node ever legitimately wants auto-mode kad *and* record serving, but it is more config surface for a case nobody has asked for. Your call; either way the `announce_self` doc needs to match.

## 2. Conflicts, and they are not small

`mergeable=CONFLICTING`. **31 commits have landed on `main` since this branch's merge-base**, and every file the PR touches has moved — `config.rs` 8 commits, `node.rs` 5, `shard_cmd.rs` 5, `node_native.rs` 3.

Three specific collisions worth knowing before you start, all merged today:

- **#133** changed `contribute.shards` to `Option<bool>` in the same `set_key` match block where this PR adds `"announce_self"` and `"dht_server"` arms.
- **#132** removed the implicit `save()` from the tail of `set_key` — it now mutates only, and callers persist. If you add keys, nothing else is needed, but the function's contract changed.
- **#130** rewrote large parts of `node_native.rs`, which is the file the fix above has to touch.

Also #128 sandboxes `kwaainet_dir()` under `cfg(test)`, so tests can no longer write the real config — relevant if any of your 16 commits added config tests.

Happy to take the rebase off your hands if that is easier than context-switching back to it — say the word.

dazwin:
Rebased onto `main` at `9190d4a2` (#130) — head is now `8d4a5778`. Both points in the nudge were stale against the branch, so stating where each stands:

**1. The `dht_server` change request.** That review predates `9d173298` *(refactor(config): drop dht_server; serving is not a config key)*. There is no `dht_server` key in `KwaaiNetConfig` to gate anything with — `git grep dht_server` over the branch returns three hits, all byte-identical to `main` (`node.rs` comment, and `node_native.rs:141-142` hardcoding `true`). The proposed patch does not compile against the branch for that reason. Resolution as agreed on 16 Aug: serving is unconditional, as on `main`; `BOOTSTRAP.md` records why, and the auto-mode claim was corrected in `908e0139`. The CHANGES_REQUESTED state is left over from before the key was dropped.

**2. Conflicts.** The branch was already based on `4b758b41` (#132), so #128, #132 and #133 were underneath it, not ahead of it. Only #129 and #130 were new; the replay was clean.

**One fix of our own.** `the_new_keys_round_trip_through_set_and_save` wrote to `$KWAAINET_HOME` and read it back — the pattern `main`'s `test_isolation` note documents as "passes alone and fails in the suite", because `grpc_server`'s tests set that var under their own lock. Green 6/6 on the #86 base this was reviewed on; ~80% red on #132+. Now round-trips through `serde_yaml` in memory, which is `save()`/`load_or_create()` minus the filesystem. The env lock this branch had added goes with it — no process-global state left in the PR.

Still 8 files. `fmt`/`clippy` clean; 164 `kwaainet` unit tests ×5 and tier 14's 5 green.

RezaRassool:
Rebased onto `main` (`093eb34b`) and pushed — no conflicts, 17 commits replayed clean. On the rebased branch: **164 kwaainet tests pass**, clippy clean, real config untouched by the run.

Reza asked me to rebase and apply the outstanding review fix. The rebase is done; **the fix turned out to be already in**, so I have applied nothing and left your resolution alone.

For the record, since the PR still shows `CHANGES_REQUESTED` and that is now misleading: the finding was that `dht_server: false` did not stop the node serving DHT writes, because `spawn_dht_service` runs unconditionally while the key's doc claimed otherwise. `0ed7494d` resolves it by **removing the key** rather than making it gate the handlers, and `announce_self`'s doc now states the position plainly:

> Serving — answering rpc_ping/rpc_store/rpc_find for other peers — is not configurable: every native node does it, and there is no record validation, so a served store accepts any key from any peer that can reach it. Validators are the control for that, not a config key.

I think that is the better of the two fixes. I had drafted the guard version — gate `spawn_dht_service` on the key — and it would have made the doc true while leaving a control whose only real effect is to decline traffic the node has no way to validate anyway. Deleting a key that promised something the system cannot deliver is more honest than making the promise narrowly true.

Worth being explicit that this closes the *contradiction* and not the *concern* behind it. Reza's point was that with no record validators, that key was the only lever for declining unauthenticated writes. After this change there is no lever at all — which is the accurate description of where we are, now written down where someone will find it. Validators remain the actual work.

@dazwin — one thing I did not touch, from my earlier comment: `DHTStorage::new` stays outside any guard because `storage` is a struct field the announce path writes to (`self.storage.handle_store`, so the node serves what it publishes). Moving it would make it `Option<DHTStorage>` and ripple to both call sites, which is not worth an empty map. That suggestion of mine was wrong; your structure is right.

#96 is rebased on top of this and pushed too.

RezaRassool (approved):
Approving, and dismissing my own earlier position on `dht_server` — your resolution is better than the one I proposed.

**Disclosure:** I rebased this branch onto `main` (`093eb34b`) at Reza's request. Seventeen commits replayed with no conflicts; I changed no content. Reviewing a branch I have touched, so the rebase is worth stating even though it was mechanical.

## On dropping the key

I had drafted the guard version — gate `spawn_dht_service` on `dht_server` — and posted it here. Your second argument is the one I missed, and it is decisive:

> Unregistering the handlers is invisible to placement. With `decentralized_dht` on, candidates are drawn from the routing table with **no filter on who serves**, so a non-serving node is still selected among the *k* nearest, its store fails, and replication silently degrades toward the shortfall warning.

My patch would have made the doc true and shipped a control that silently degrades replication when used. That is a worse outcome than the contradiction it fixed, and it is the same failure shape as the objection that started this thread — a switch that appears to do something while the system routes around it.

Recording the reasoning in `BOOTSTRAP.md` so it is not re-proposed without the routing-table half is the right call.

**One thing to keep visible:** this closes the contradiction, not the concern. There is now no lever at all for declining unauthenticated writes, and the `announce_self` doc says so plainly — *"there is no record validation, so a served store accepts any key from any peer that can reach it. Validators are the control for that, not a config key."* That is honest, and it means validators are load-bearing future work rather than a nice-to-have. Your correction that `RequestAuthInfo` is a wire shape nothing reads — no signature checked because none is verified — is worth having stated somewhere permanent.

## Verified

On the rebased branch: **164 kwaainet tests pass**, clippy clean, real config untouched by the run. CI green on the new head.

Traced the parts most likely to break:

- **The tombstone gate.** `if config.announce_self { node.unannounce(...) }` with the reason on the line above — a bootstrap published nothing, so it has nothing to retract. Correct, and the failure mode if it were wrong (a bootstrap writing `state = -1` for records it never made) would be quiet.
- **Call-site gating over silent no-ops.** `announce`/`unannounce` no longer check internally. That is the right direction and the same lesson as #140: a function that silently does nothing when asked is indistinguishable from one that had nothing to do.
- **The serde default.** `a_config_without_the_new_keys_deserialises_to_node_defaults` pins exactly the failure that would have been catastrophic — `bool::default()` on `announce_self` is `false`, which would have muted every existing node on the network at once. Good test to have written before it was needed.
- **The bootstrap fallback.** Gating `with_petals_bootstrap()` on `announce_self` is right: "empty" genuinely means two different things, and a live run catching a bootstrap dialling the public network is the kind of evidence that is hard to argue with.

## One observation, not blocking

`announce_self` now appears at seventeen points in `node_native.rs`. Each is individually justified and the call-site direction is correct, but that is a lot of independent places to remember. If a future announce path is added, nothing makes its author aware the gate exists.

Not worth restructuring now — the alternative (gating inside, which is what you moved away from) is worse. Perhaps a line in `BOOTSTRAP.md` under a heading like *adding a new announce path* saying the gate is the caller's responsibility, so the next person finds it before shipping.

## #96

Rebased on top of this and green. It is stacked, so this merges first.

PR #95, opened 2026-08-15
feat(dht): decentralized placement — any reachable node is a join point

dazwin:
> [!NOTE]
> **Based on `feat/bootstrap-serve` (#94), not `main`** — this is phase 2 of the same work and builds directly on that branch, which in turn builds on the native-p2p stack. Basing it here keeps the diff to this branch's single commit. It is deliberately **not** part of GitHub stack #92. Retarget once the branches below it land.

Phase 2 of bootstrap decentralisation. Behind `decentralized_dht` (default **false**), a node places and finds DHT records on the *k* peers nearest each record key, instead of fanning out to a hardcoded bootstrap list. The fixed bootstraps become ordinary candidates — any reachable node is a join point.

## The keyspace is hivemind's, not libp2p-kad's

This is mandatory rather than a preference, and it is the part most worth reviewing:

- node ID = `SHA1(peer_id.to_bytes())`
- record ID = `SHA1(msgpack(key))`
- distance = XOR over the 20-byte IDs

The Python health service beam-searches in that space via `DHTProtocol.rpc_find` nearest lists, which `DHTStorage` answers XOR-sorted. **A record placed by kad's SHA256 distance would store successfully and be findable by nobody** — a silent failure, so `placement` asserts the distinction directly and the mistake cannot land unnoticed.

## Placement is attempt-and-advance

`rank_candidates` returns the whole ranking, and a failed store advances to the next-nearest rather than consuming a replica slot — two dead peers at the front cost latency instead of durability. Records under different keys land on different peer sets; that is the point, and no single peer holds everything.

Per-key placement also means the block record's packed parallel arrays must be split before delivery (`split_by_key`), since one request carrying eight block keys has eight different nearest sets.

## New

- **`placement`** — the pure selection math (20 unit tests)
- **`peer_cache`** — `<config_dir>/peer-cache.json`; 60 s timer plus a shutdown write, bounded to the 100 most recent, atomic via temp+rename, malformed files tolerated as empty (13 unit tests)
- **`NetworkHandle::known_peers`** — routing table ∪ connected set, address-filtered. (`routing_peers` returns bare IDs, which cannot be stored to.)
- **tier 15** — three native nodes, no bootstrap: B joins via A, C via B, C publishes, A reads it back, and the record survives B being killed

## Config and blast radius

`decentralized_dht: bool = false` and `dht_replication: usize = 3`, both settable via `kwaainet config set`. Native path only; p2pd keeps its existing behaviour. With the flag off, `NativeNode::deliver` delegates straight to `send_records_via_handle`, so nothing changes.

## Tests

128 kwaainet unit, 74 kwaai-p2p, full integration suite (tiers 01–15) green; release build clean.

### DIFF
--- a/core/crates/kwaai-cli/src/node_native.rs
+++ b/core/crates/kwaai-cli/src/node_native.rs
@@ -459,14 +459,10 @@ pub async fn run_native_node(
         crate::node::jitter_secs(300, 30),
     )));
     // The Ollama watcher exists to trigger a re-announce when inference comes
-    // back. A bootstrap announces nothing, so the watcher would poll a port it does
-    // not serve every 15 s in order to fire a signal whose handler is a no-op.
-    //
-    // A dangling receiver stands in so the `select!` arm below keeps its type.
-    // The sender is *held* rather than dropped: dropping it would close the
-    // channel, and while `Some(()) = …recv()` would still disable the branch on
-    // `None`, holding the sender keeps the branch simply-never-ready, which
-    // does not depend on that subtlety.
+    // back, so a bootstrap does not start one: it would poll a port it does not
+    // serve every 15 s to fire a signal for an arm that is disabled anyway.
+    // A dangling receiver stands in so the `select!` arm keeps its type, and the
+    // sender is held rather than dropped so the branch is simply never ready.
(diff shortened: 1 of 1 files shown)


dazwin:
Superseded by #96 — this PR was auto-closed when a bad rebase on my side briefly pushed the branch without its commit. The branch is restored; #96 carries the same single commit against the same base.

PR #96, opened 2026-08-15
feat(dht): decentralized placement — any reachable node is a join point

dazwin:
> [!NOTE]
> **Based on `feat/bootstrap-serve` (#94), not `main`** — this is phase 2 of the same work and builds directly on that branch, which in turn builds on the native-p2p stack. Basing it here keeps the diff to this branch's single commit. It is deliberately **not** part of GitHub stack #92. Retarget once the branches below it land.

Phase 2 of bootstrap decentralisation. Behind `decentralized_dht` (default **false**), a node places and finds DHT records on the *k* peers nearest each record key, instead of fanning out to a hardcoded bootstrap list. The fixed bootstraps become ordinary candidates — any reachable node is a join point.

## The keyspace is hivemind's, not libp2p-kad's

This is mandatory rather than a preference, and it is the part most worth reviewing:

- node ID = `SHA1(peer_id.to_bytes())`
- record ID = `SHA1(msgpack(key))`
- distance = XOR over the 20-byte IDs

The Python health service beam-searches in that space via `DHTProtocol.rpc_find` nearest lists, which `DHTStorage` answers XOR-sorted. **A record placed by kad's SHA256 distance would store successfully and be findable by nobody** — a silent failure, so `placement` asserts the distinction directly and the mistake cannot land unnoticed.

## Placement is attempt-and-advance

`rank_candidates` returns the whole ranking, and a failed store advances to the next-nearest rather than consuming a replica slot — two dead peers at the front cost latency instead of durability. Records under different keys land on different peer sets; that is the point, and no single peer holds everything.

Per-key placement also means the block record's packed parallel arrays must be split before delivery (`split_by_key`), since one request carrying eight block keys has eight different nearest sets.

## New

- **`placement`** — the pure selection math (20 unit tests)
- **`peer_cache`** — `<config_dir>/peer-cache.json`; 60 s timer plus a shutdown write, bounded to the 100 most recent, atomic via temp+rename, malformed files tolerated as empty (13 unit tests)
- **`NetworkHandle::known_peers`** — routing table ∪ connected set, address-filtered. (`routing_peers` returns bare IDs, which cannot be stored to.)
- **tier 15** — three native nodes, no bootstrap: B joins via A, C via B, C publishes, A reads it back, and the record survives B being killed

## Config and blast radius

`decentralized_dht: bool = false` and `dht_replication: usize = 3`, both settable via `kwaainet config set`. Native path only; p2pd keeps its existing behaviour. With the flag off, `NativeNode::deliver` delegates straight to `send_records_via_handle`, so nothing changes.

## Tests

128 kwaainet unit, 74 kwaai-p2p, full integration suite (tiers 01–15) green; release build clean.


RezaRassool:
Read through ahead of the branches below this landing. Defaulting `decentralized_dht` to **false** is the right call for a change this structural, and keeping hivemind's keyspace rather than libp2p-kad's is the detail that makes it interoperable at all rather than a parallel network.

One thing worth deciding before the flag is ever flipped, rather than after.

## Placement on *k* nearest peers changes who the validator gap applies to

Today every native node already serves DHT writes — `node_native.rs:144` calls `spawn_dht_service` unconditionally — but in practice the fixed bootstrap list means almost all records land on the two AWS boxes. Ordinary nodes hold little because nobody routes to them.

Decentralised placement removes exactly that accident of topology. Once records go to the *k* peers nearest each key, every reachable node becomes a storage target for keys it did not choose, written by peers it cannot authenticate — because record validators are still not implemented (`kwaai-hivemind-dht/src/server.rs` module docs are honest about this). The interim guard is capacity-bounding, which prevents unbounded growth but not unauthenticated *content*.

That is a different risk profile from "two bootstraps we operate hold everything". It is not an argument against this design — decentralised placement is clearly where this should go — but it does mean **validators stop being a bootstrap prerequisite and become a precondition for enabling this flag**.

Same for the eviction cliff: `enforce_capacity` does an O(n) `.iter().min_by(…)` under the write lock against a 1,048,576-entry bound. Spreading records across the fleet makes it more likely some node somewhere reaches capacity, and that node serialises its DHT service when it does.

## Suggestion

Neither blocks merging, precisely because the default is `false`. What I would want is the precondition written down where someone about to flip it will see it — a line in the `decentralized_dht` doc comment along the lines of *"requires record validators; see server.rs module docs"* — so the flag cannot be turned on in good faith without meeting the person who knows why not.

Cheap to add, and it is the difference between a documented staged rollout and a footgun with a default that happens to be safe.

Will review the implementation properly once #94 and #86 land and this retargets.

RezaRassool:
Full review now that #86 has landed. This is strong work — the hardest part is identified and handled correctly, and the gating is conservative. It cannot merge until #94 does (which has changes requested), so this is advisory.

## The keyspace trap is the thing that mattered, and you got it

The module docs in `placement.rs` are the best part of this PR:

| | peer id | key |
|---|---|---|
| libp2p-kad | `SHA256(peer_id)` | `SHA256(key)` |
| **hivemind** | `SHA1(peer_id.to_bytes())` | `SHA1(msgpack(key))` |

> A record placed by SHA256 distance would sit on peers that no SHA1 beam search ever walks to.

That is exactly the failure that would have been near-impossible to debug in the field: placement "succeeds", every node reports healthy, and records are simply unfindable by anyone using hivemind's own lookup. Naming it in the module docs, deriving from `NodeInfo::from_peer_id`, and justifying why lexicographic XOR over big-endian 20-byte IDs equals numeric order — that is the level of care this needed.

## Bounds are right

- `MAX_CACHED_PEERS = 100` — the peer cache cannot grow without limit
- `CACHE_WRITE_INTERVAL_SECS = 60` — no write amplification from churn
- A malformed cache file is ignored with a debug line and the node starts fresh, rather than failing to boot on corrupt state. That is the correct failure direction for a cache.

## Gating is conservative and verified

`decentralized_dht` defaults to `false`, every use site is behind `if config.decentralized_dht` (`node_native.rs:204, 665, 678`), and it is native-path only. I confirmed `node.rs` — the p2pd path the whole fleet runs — is **not touched by this PR at all**. Nobody's node changes behaviour on upgrade.

## Verification

Build and `clippy --all-targets -- -D warnings` pass. Tests: **placement 20**, **peer_cache 13**, **tier 15 (decentralized DHT) 4**. 565 lines of tier-15 coverage against 1093 of implementation is a good ratio for something this structural.

## One ask, carried over

The `decentralized_dht` doc comment explains what the flag does but not what it presupposes. Record validators still are not implemented, and this flag is what turns "two bootstraps we operate hold everything" into "every reachable node stores keys it did not choose, written by peers it cannot authenticate". The capacity bound stops unbounded growth; it does not police content.

A line in the doc comment — *"requires record validators; see `kwaai-hivemind-dht/src/server.rs` module docs"* — is all I am asking for. The default being `false` is what makes this safe today; the doc comment is what keeps it safe the day someone decides to flip it, possibly not you and possibly without reading the DHT server's module docs first.

Same applies to `enforce_capacity`'s O(n) eviction: spreading records across the fleet makes it likelier that some node reaches the 1,048,576 bound, and that node serialises its DHT service when it does.

Approve once that line is in and #94's `dht_server` gating is fixed.

RezaRassool:
@dazwin — 7 days old today, same as #94.

This one reads `MERGEABLE / CLEAN`, but only because its base is `feat/bootstrap-serve` rather than `main`. It inherits #94's rebase the moment that lands: **31 commits** have gone into `main` since that branch's merge-base, including four today (#128, #130, #132, #133) that touch `node_native.rs`, `config.rs` and `set_key`.

At 15 files and +2474 lines this is the larger of the two, so it is worth knowing the conflicts are queued up behind #94 rather than absent.

I have posted a proposed patch for #94's outstanding change request there, which should unblock the ordering. Nothing needed on this PR until that moves.

dazwin:
Rebased onto the rebased #94 (`8d4a5778`) — head is now `12fc0a13`, and the PR is back to its own diff: 15 files, +2474/−13, `MERGEABLE`.

The only thing that changed in the commit itself is the import lists where #130's `NetworkSnapshot` / `RoutingEntry` exports landed in the same lines as this branch's `KnownPeer` (`kwaai-p2p/src/lib.rs` and the two call sites). No logic moved.

Re-verified on the new base:

| | |
|---|---|
| `cargo clippy -p kwaainet -p kwaai-p2p --all-targets -- -D warnings` | clean |
| `cargo test -p kwaainet` | 202 passed, 0 failed (incl. the 33 `placement::` / `peer_cache::` tests) |
| `cargo test -p kwaai-p2p` | 178 passed, 0 failed |
| tier 14 `14_bootstrap_seed` / tier 15 `15_decentralized_dht` | 5/5 and 4/4 |

Three small fixes that came out of exercising this stack on the NAT test bed are up as independent PRs against `main` rather than stacked here, because none of them depends on #94 or this: #137 (routed dials by `PeerId` — explicit candidate list, one DHT walk on failure), #138 (relay circuit limits sized for traffic), #139 (config reload before runtime saves). `git merge-tree` of each against both #94 and this branch is clean, so order of landing does not matter.

RezaRassool:
Rebased onto the updated `feat/bootstrap-serve` and pushed — one commit, replayed clean.

On the rebased branch: **202 kwaainet tests, 251 p2p+dht tests, 0 failures**, clippy clean.

This was previously blocked behind #94's outstanding review. That review is resolved — by `0ed7494d` dropping the `dht_server` key rather than gating on it; see the note on #94. Nothing in this PR needed changing.

Both branches now sit on `main` at `093eb34b`, which includes the four fixes merged today: #137 (routed dials fall back to a DHT walk), #138 (relay circuit sizing), #139 (config reload before save) and #140 (dream cycle guard). #137 and #139 in particular touch the same area as this stack, and the rebase surfaced no conflicts with either.

RezaRassool (approved):
Approving. The keyspace analysis is the heart of this and it is right — including the part that would have been easiest to get wrong quietly.

**Disclosure:** I rebased this branch onto `main` after #94 merged and retargeted the PR, at Reza's request. One commit replayed clean; no content changed.

## The keyspace

> A record placed by SHA256 distance would sit on peers that no SHA1 beam search ever walks to.

This is the claim the whole PR rests on, and it is the kind that is normally asserted in a comment and left to rot. Here it is pinned:

```rust
let expected = kwaai_hivemind_dht::protocol::NodeInfo::from_peer_id(peer).node_id;
assert_eq!(peer_dht_id(&peer), expected);
```

I went looking for the gap — `xor_distance` is a *copy*, since hivemind's is private, and a copy is exactly where a silent divergence would live. `xor_distance_matches_the_hivemind_server` covers it, and the doc comment on the function names that test as the guard. That is the right structure: the duplication is justified, the risk is named, and the named risk is tested.

The left-zero-extension on differing lengths is a good detail — a malformed ID orders deterministically instead of panicking, which matters for input that arrives off the wire.

## Blast radius

The flag-off path is a clean early return:

```rust
if !self.decentralized {
    return send_records_via_handle(&self.handle, bootstrap_peers, records).await;
}
```

No shared setup runs first, so with `decentralized_dht: false` — the default — an existing node takes exactly today's code path. The peer-cache timer is likewise gated at the `select!` arm rather than started and ignored.

## Verified

On the rebased branch, against `main` at `0f856c0e`: **202 kwaainet tests, 251 p2p+dht, 0 failures**, clippy clean, real config untouched. The new modules' own suites: **20 placement, 13 peer_cache**.

`peer_cache` does what it says — `with_extension("json.tmp")` then `rename`, bounded at `MAX_CACHED_PEERS = 100`, with a test asserting a crashed write leaves the real file untouched. Atomicity claimed and atomicity tested.

## Two observations, neither blocking

**Attempt-and-advance is the right default, and worth watching in the field.** Advancing past a dead peer rather than consuming a replica slot is clearly correct for durability. The cost is that a partition, or a run of stale cache entries at the front of the ranking, turns into latency rather than a visible failure — the same shape as the silent degradations we have spent this week fixing elsewhere. If `dht_replication: 3` is being met only after walking ten candidates, nothing currently says so. A count of attempts-per-successful-store in the store timings would make that visible cheaply.

**`split_by_key` for the packed block record** is the non-obvious consequence of per-key placement, and it is easy to imagine a future record type acquiring parallel arrays without anyone remembering it must be split. Worth a line in the placement module docs stating the invariant — one request, one key — so the next packed record type meets it rather than silently placing eight keys on one key's nearest set.

## On the argument this PR settles

You cited exactly this code when arguing against a `dht_server` switch on #94 — that candidates are drawn from the routing table with no filter on who serves, so a non-serving node would still be selected and its store would fail. Having now read `gather_candidates` and `rank_candidates`, that is plainly true, and it is a better reason to have dropped that key than the honesty argument I was making. Worth noting the converse for later: if a serving filter is ever added, this is the module that has to learn about it.

PR #99, opened 2026-08-17
feat(cli): `p2p probe` — measure a peer's round-trip latency

RezaRassool:
Asks a peer for the least work the fabric allows and times the reply.

```
$ kwaainet p2p probe --peer 12D3KooWCzuhpXrZ… --concurrency 4

  ✅ Round-trip time: 158.4 ms (median)
  min 156.5 ms   median 158.4 ms   max 162.9 ms   mean 159.4 ms   n=7
  💡 4 concurrent calls completed in 163.8 ms (4/4 round-tripped)
  💡 Run one at a time that would be ~633 ms — 3.9x overlap
```

## Why it works on any peer

The default protocol is one this project undertakes never to implement, so the
peer refuses it during multistream-select — and **a refusal still crosses the
network and comes back**. No cooperation required: bootstrap nodes, peers on
older versions, and peers not running kwaainet at all can all be probed. Pass
`--proto` to time a real handler instead; the difference is what that handler
costs.

## What it established

A minimal request costs **exactly one round trip and no more**:

| peer | bare TCP handshake (1 RTT) | `p2p probe` |
|---|---|---|
| bootstrap-1 (Ohio) | 75–80 ms | 75–83 ms |
| gandalf (direct, distant) | 186–187 ms | 184–187 ms |
| metro-linux (via relay) | — | 156–163 ms |

Stream setup, multistream-select, hivemind framing and the p2pd control socket
together add nothing measurable on top of the network. Two corollaries, both
verifiable with the flags this adds:

- **Payload is free to at least 256KB** (`--payload-bytes`) — latency-bound, not
  bandwidth-bound.
- **Calls overlap nearly perfectly** (`--concurrency`) — 4 concurrent cost the
  wall time of 1.

## Why it is a scheduling input, not just a diagnostic

The figure is the peer's network distance, and it is the entire per-call
transport cost of anything else you ask that peer to do. Which kills an
assumption worth killing: **a relayed peer is not inherently slower than a direct
one.** metro-linux through the Ohio relay (158 ms) beats directly-connected
gandalf (190 ms), because geographic distance dominates and the relay does not.

For context on where this came from: chasing why remote sharded inference ran at
~1 tok/s. The answer turned out to be `1 RTT + compute` per token — for
metro-linux, 158 ms transport + ~250 ms compute = the 411 ms/token measured. The
transport is not the problem, and this command is how that was settled.

## Also here

`persistent.rs` logged "protocols not supported" at `warn`, so probing emitted a
warning per call and drowned its own output. It now logs at `debug` alongside the
other negotiation-level outcomes: asking whether a peer speaks a protocol is a
normal thing to do, and "no" says something true about the peer rather than wrong
about us.

## Notes for review

- **What counts as a sample** is the part worth scrutinising. A call that never
  reached the peer is not a fast round trip — it is not a round trip at all.
  Dial failures, routing misses and timeouts are excluded from the statistics and
  reported separately, and a warm-up call runs first so a cold connection does
  not skew the median. Unit tests cover the classification and the order
  statistics.
- `PROBE_PROTO` must stay unimplemented. Implementing it anywhere would silently
  turn every probe into a measurement of the handler instead of the path; the
  constant's doc comment says so.
- 8 new tests pass; fmt and clippy clean. `cargo test -p kwaainet` is 96 passed /
  2 failed, the two being the pre-existing environmental `grpc_server::tests`
  failures from a running daemon holding port 8093.

https://claude.ai/code/session_01DTqHSBgjKCaKV3Tj6L4E7t

### DIFF
--- a/core/crates/kwaai-cli/src/cli.rs
+++ b/core/crates/kwaai-cli/src/cli.rs
@@ -2392,6 +2392,56 @@ pub enum P2pAction {
 
     /// Inspect active connections to other peers
     Peers(PeersArgs),
+
+    /// Measure round-trip latency to a peer.
+    ///
+    /// Sends the smallest request the fabric allows and times the reply. By
+    /// default it names a protocol deliberately left unimplemented, so the peer
+    /// refuses it during protocol negotiation — a refusal still completes a full
+    /// round trip, which means **this works against any libp2p peer without its
+    /// cooperation**, including bootstrap nodes and peers not running kwaainet.
+    ///
+    /// A minimal request costs one round trip and no more: payload size makes no
+    /// difference up to at least 256KB, and it does not matter whether a handler
+    /// runs. So the figure here is the peer's network distance, and it is the
+    /// entire per-call transport cost of anything else you ask that peer to do.
+    Probe(ProbeArgs),
+}
+
+#[derive(Args)]
+pub struct ProbeArgs {
+    /// Peer ID (base58) to probe. Must already be connected — check
+    /// `p2p peers list`. Probing an unconnected peer measures the dial as
+    /// well as the round trip, and is reported as such rather than averaged in.
+    #[arg(long)]
+    pub peer: String,
+
+    /// Number of timed round trips. One warm-up call runs first and is
+    /// excluded, so a cold connection does not skew the result.
+    #[arg(long, default_value_t = 7)]
+    pub count: usize,
+
+    /// Protocol to invoke instead of the unimplemented default. Use this to
+    /// time a real handler — the difference against the default is what that
+    /// handler costs. A protocol the peer does implement replies normally.
+    #[arg(long)]
+    pub proto: Option<String>,
+
+    /// Also measure how well calls overlap, by issuing this many at once.
+    /// Round trips to a peer are independent, so wall time close to a single
+    /// call means latency can be hidden by running work concurrently.
+    #[arg(long)]
+    pub concurrency: Option<usize>,
+
+    /// Payload size in bytes. Sizes up to at least 256KB cost the same as an
+    /// empty call, so raise this to confirm you are latency- and not
+    /// bandwidth-bound on a particular path.
+    #[arg(long, default_value_t = 1)]
+    pub payload_bytes: usize,
+
+    /// Per-call timeout in seconds.
+    #[arg(long, default_value_t = 10)]
+    pub timeout: u64,
 }
 
 #[derive(Args)]
--- a/core/crates/kwaai-cli/src/p2p_cmd.rs
+++ b/core/crates/kwaai-cli/src/p2p_cmd.rs
@@ -761,12 +761,24 @@ enum Sample {
 
 /// Classify a call outcome. Refusal is success for our purposes; anything that
 /// failed before reaching the peer is not a latency sample at all.
+///
+/// The two transports word the refusal differently, and matching only one of
+/// them makes the probe report "could not reach the peer" against a peer it
+/// reached perfectly well:
+///
+/// ```text
+/// p2pd:   failed to negotiate protocol: protocols not supported: [/kwaai/…]
+/// native: remote does not support protocol /kwaai/…
+/// ```
+///
+/// Matching on `not support` covers both, and any similar phrasing a future
+/// transport picks — the alternative is a list that silently rots the next time
+/// one of them rewords an error string.
 fn classify(result: Result<Vec<u8>, String>) -> Sample {
     match result {
         Ok(_) => Sample::RoundTrip,
         Err(e) => {
-            let low = e.to_lowercase();
-            if low.contains("protocols not supported") || low.contains("protocol not supported") {
+            if e.to_lowercase().contains("not support") {
                 Sample::RoundTrip
             } else {
                 Sample::NoRoundTrip(e)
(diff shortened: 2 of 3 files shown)


dazwin:
The command is worth having, and the "a refusal is a full round trip" idea is the right one — nothing else in the CLI can report latency today. Two things need fixing before it lands, one of which makes the command report nothing on a native node.

## Blocking: `classify` only recognises the p2pd refusal

`classify` matches on `"protocols not supported"`, which is the **Go daemon's** wording. The native path produces a different string:

```rust
// kwaai-p2p/src/handle.rs:584 (and :616)
P2PError::Protocol(format!("remote does not support protocol {p}"))
```

`"remote does not support protocol …"` contains neither `"protocols not supported"` nor `"protocol not supported"` — the words are there but the substring isn't. So on `native_p2p = true` every refusal classifies as `NoRoundTrip`, and the warm-up call bails out with:

```
❌ could not reach the peer: remote does not support protocol /kwaai/probe-unrouted/1.0.0
```

against a peer that answered correctly, in the expected way, on the first try.

Confirmed by adding the native string to the PR's own test:

```
left: NoRoundTrip("remote does not support protocol /kwaai/probe-unrouted/1.0.0")
right: RoundTrip
```

The eight passing tests don't catch it because every refusal fixture is a p2pd string. Suggested fix — match the negotiation outcome rather than one daemon's phrasing:

```rust
let refused = low.contains("protocols not supported")     // p2pd
    || low.contains("protocol not supported")
    || low.contains("does not support protocol");         // native
```

and add the native string as a test case alongside the p2pd one, so the next divergence fails here rather than in the field.

Worth noting the same asymmetry hits the `persistent.rs` log-level change: that fix is correct for p2pd, but the native path never goes through `persistent.rs`, so it needs its own treatment if the warning is noisy there too.

## Blocking-ish: `PROBE_PROTO` is guarded only by a comment

The doc comment is clear that implementing this protocol anywhere silently converts every probe into a handler measurement. That's exactly the kind of contract that gets broken years later by someone who never read the comment, and it fails **silently** — the numbers stay plausible and just stop meaning what they say.

Ten lines removes the concern:

```rust
#[test]
fn probe_proto_is_never_registered() {
    // Registering a handler for this name turns every probe into a
    // measurement of that handler instead of the network path.
    let hits = std::process::Command::new("grep")
        .args(["-rl", PROBE_PROTO, env!("CARGO_MANIFEST_DIR"), "../"])
        .output().expect("grep");
    // ... assert the only file naming it is this one
}
```

Either that, or register it in whatever central protocol list exists so a duplicate is a build error.

## Non-blocking notes

**`--payload-bytes 0`** produces an empty payload with no validation. Probably fine — a zero-length unary call should still negotiate and round-trip — but the flag's help says "up to at least 256KB" without saying whether 0 is legal. Worth either rejecting it or confirming in the help that it's the true minimum.

**The concurrency estimate is optimistic about what it proves.** `serial_estimate = median * n` compares against the *sequential* median, but the p2pd path funnels every call through one persistent connection (`persistent.rs:253`, single writer lock, responses demuxed by `call_id`). So the "3.9x overlap" figure is measuring the daemon's pipelining, not the network's — which is a genuinely useful thing to know, just not quite the thing the output claims. A word in the printed line would fix it.

**On the PR description**: the "works on any peer without cooperation" framing undersells this. `/ipfs/ping/1.0.0` is already advertised by every peer in the tree (see the captured protocol lists in `peers_view.rs:252` and `:269`) and would also work on any peer. The better argument — and the one that survives the "why not just use ping?" question — is that ping runs inside the swarm and skips the entire unary call path: stream setup, multistream-select, the control socket. `probe` measures the path a real request actually takes, which is why the number is a scheduling input rather than a network-distance trivia figure. Recommend leading with that.

**Naming**: agree with `probe` over `ping`. `/ipfs/ping` is real, is running, and already feeds the GUI's RTT column via `rtt_ms` — a CLI `ping` measuring something unrelated would be a genuinely nasty debugging trap.

## Adjacent gap this exposes

The CLI has no latency surface at all, and that's structural rather than an oversight: the swarm records a per-peer RTT from the ping behaviour (`service.rs:1492`), the gRPC `Network` op forwards it as `rtt_ms`, and the GUI renders it — but the p2pd control socket's `LIST_PEERS` reply has no RTT field, so `p2p peers list` can never show it. Adding that field would give the CLI a free RTT column and leave `probe` as the deliberate measurement rather than the only one. Follow-up, not this PR.

Separately and trivially: `peers_send` times its call for the timeout but never prints the elapsed time (`p2p_cmd.rs:571-587`). Two lines, unrelated to this change, and it'd make the existing one-shot command useful for exactly the question people will reach for `probe` to answer.

dazwin (approved):
Re-reviewed at `df39d740`. The blocking bug is fixed, and the fix is better than what I suggested.

**Verified locally on the updated branch:**

- 9/9 `p2p_cmd` tests pass, including the new `a_refused_protocol_is_a_valid_round_trip_on_native`.
- `cargo fmt --check` clean, `cargo clippy` clean.
- The native string that failed before now classifies as `RoundTrip`.

**On matching `not support` rather than a literal list** — I suggested enumerating both wordings; matching the shorter substring is the better call, and I checked it doesn't over-match. Every `not support` string in the p2p layer is a genuine protocol refusal:

```
kwaai-p2p/src/handle.rs:729   "remote does not support protocol {p}"    (raw stream)
kwaai-p2p/src/handle.rs:761   "remote does not support protocol {p}"    (unary)
kwaai-p2p/src/raw_stream.rs:79 "remote does not support any of the requested protocols: {0}"
kwaai-p2p/src/unary.rs:114     "remote does not support protocol {0}"
```

No dial failure, timeout, or transport error in either crate contains that substring, so the widened match can't silently promote a non-round-trip into a sample. That's the failure mode that would have mattered, and it isn't there.

The commit message is also right about *why* the original tests missed it: they asserted the wording that already worked. Pinning both transports is the correct regression shape.

**Still outstanding — the `PROBE_PROTO` guard.** Not added in this push. `/kwaai/probe-unrouted/1.0.0` is currently unregistered anywhere (I grepped the tree — the only hits are the constant and two test fixtures, all in `p2p_cmd.rs`), so the command is correct *today*. The concern is unchanged: the contract is enforced only by a doc comment, and breaking it fails silently with plausible-looking numbers.

I don't think that should block this. It's a guard against a future mistake rather than a present defect, and this PR has already earned its keep. Worth a follow-up issue so it doesn't evaporate.

The two non-blocking notes from my first pass (the concurrency figure measuring p2pd's pipelining rather than the network's, and `--payload-bytes 0` being unvalidated) also stand, and are also fine to leave.

Approving.

PR #100, opened 2026-08-17
docs: native pays an extra round trip on every accepted call

RezaRassool:
Found while verifying the native path ahead of the 0.6 cutover ("drop p2pd, go native"). **The native stack pays two network round trips for an accepted unary call where p2pd pays one**, which makes remote inference ~57% slower.

## The measurements

Same machine, same peer, same protocols, same relay. The only variable is `native_p2p`.

| call | p2pd | native |
|---|---|---|
| refused protocol | 158.9 ms | 157.7 ms |
| **accepted protocol** (`/kwaai/p2p/hello/1.0.0`) | **159.4 ms** | **320.1 ms** |
| 256KB payload, refused | 158 ms | 158.6 ms |
| local socket write | 0.0–0.1 ms | 0.0–0.1 ms |
| **inference, 14-token decode** | **478 ms/tok** | **750 ms/tok** |

Compute is ruled out (identical remote peer), bandwidth is ruled out (256KB costs the same as one byte), local IPC is ruled out.

## Why the refusal case is the diagnostic — and why this nearly shipped

A refused protocol costs the same on both stacks, because a refusal is answered on the first flight and there is never a second exchange to pay for. Only an *accepted* call reaches the step where native waits.

That is also how it almost got missed. `p2p probe` (#99) defaults to an unimplemented protocol so it needs no cooperation from the peer — and in that default mode it reports a clean, healthy **158 ms on native**. The probe's own documented assumption, that handler dispatch and payload size do not affect call cost, was verified true on p2pd and is **false on native**.

## Mechanism — hypothesis, not fact

Everything fits optimistic negotiation being absent on the native path: go-libp2p sends the multistream-select proposal *with* the payload; rust-libp2p appears to await confirmation first. That predicts the doubling on accepted calls, no change on refusals, and no payload sensitivity.

**Not confirmed in code.** Inferred from timings; nobody has read the native unary send path. The fix depends on which it actually is. Worth ruling in or out: this repo carries a patched `multistream-select` that sits directly on this path.

## What native did better

This is one fixable defect in an otherwise good cutover, and the document says so: **15 of 16 connections direct** (against p2pd's 3 direct / 3 relayed on the same machine minutes earlier), correct circuit addresses carrying both peer IDs, identity preserved across the switch, no Go daemon spawned, DHT reads fine.

## Reproducing

```bash
kwaainet p2p probe --peer <id>                                    # refused: 1 RTT
kwaainet p2p probe --peer <id> --proto /kwaai/p2p/hello/1.0.0     # accepted: 2 RTT on native
```

Equal medians = healthy. Requires #99.

## Recommendation

**Do not flip `native_p2p` to true until this is understood.** Otherwise 0.6 removes a Go dependency and halves remote inference throughput for every user in exchange.

Docs only — no code changes. Depends on #99 for the reproduction command.

https://claude.ai/code/session_01DTqHSBgjKCaKV3Tj6L4E7t

### DIFF
--- a/docs/NATIVE_P2P_PERFORMANCE.md
+++ b/docs/NATIVE_P2P_PERFORMANCE.md
@@ -1,0 +1,146 @@
+# Native p2p: an extra round trip on every accepted call
+
+Measured against `main` at `8ad4680` on 2026-08-16, from `rezas-mini-2` (macOS,
+relay-only inbound) to the live fleet. Every figure here is reproducible with
+`kwaainet p2p probe` in two commands; see §5.
+
+**The finding in one line: the native stack pays two network round trips for an
+accepted unary call where p2pd pays one, which makes remote inference ~57%
+slower.** Everything else about the native path measured better.
+
+---
+
+## 1. Why this matters now
+
+The 0.6 release is scoped as *"drop p2pd, go native"*, which means flipping
+`native_p2p` from `false` to `true` in `config.rs:671` and removing the Go
+daemon from the build. That flag is the difference between this being a defect
+some opted-in users could hit and a defect **every** user gets on auto-update.
+
+Remote inference is the visible casualty: **478 → 750 ms/token**. Nothing else in
+the cutover moves in the wrong direction.
+
+---
+
+## 2. The measurements
+
+Same machine, same peer (metro-win), same protocols, same relay. The **only**
+variable is `native_p2p`.
+
+| call | p2pd | native |
+|---|---|---|
+| refused protocol (unimplemented) | 158.9 ms | 157.7 ms |
+| **accepted protocol (`/kwaai/p2p/hello/1.0.0`)** | **159.4 ms** | **320.1 ms** |
+| 256KB payload, refused | 158 ms | 158.6 ms |
+| local socket write | 0.0-0.1 ms | 0.0-0.1 ms |
+| remote inference, 14-token decode | **478 ms/token** | **750 ms/token** |
+
+Medians, n=5-7 per cell, spreads tight (±5 ms).
+
+Three explanations are ruled out by the table itself:
+
+- **Not compute.** The remote peer is identical in both rows; only the local
+  stack changed.
+- **Not bandwidth.** A 256KB payload costs the same as one byte, on both stacks.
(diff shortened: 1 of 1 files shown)


dazwin:
Read the native unary send path to check §4's hypothesis. It holds, and two of the doc's open items can be closed:

**§4 confirmed.** In `kwaai-p2p/src/unary.rs`, `Protocols::upgrade_outbound` is `futures::future::ready` (`unary.rs:246-248`) — no I/O, so the stream only surfaces after multistream-select has completed its own round trip. The request is then written in the worker spawned by `on_fully_negotiated_outbound` (`unary.rs:459-471`), strictly after that. Nothing is sent optimistically, exactly as §4 predicted.

**§3's refusal asymmetry confirmed.** A refusal never reaches that write — it lands in `on_dial_upgrade_error` and maps `NegotiationFailed` → `UnsupportedProtocol` (`unary.rs:515-517`). That's the mechanism behind "refusals cost the same on both stacks."

**The patched `multistream-select` can be ruled out.** §4 flags it as worth ruling in or out. The ordering is imposed by rust-libp2p's `ConnectionHandler` API, which surfaces an outbound stream to a behaviour only via `FullyNegotiatedOutbound` — by construction after the negotiation response arrives. The patch isn't implicated.

**Question on the inference row.** Which transport produced 478 → 750 ms/token? There are two, and they should behave differently here:

- `/kwaai/inference/1.0.0` (`block_rpc.rs:44`) is unary — one `call_unary_handler` per call (`block_rpc.rs:191`), so it pays the extra round trip per token.
- `/kwaai/inference-mux/1.0.0` (`inference_mux.rs:43`) is a single persistent stream with requests correlated by `request_id`, selected by `mux://PEER_ID` in `--inference-urls`. Negotiation is paid once per stream, so it shouldn't be affected.

If it was the unary path, §4 explains the number fully. If it was `mux://` and still showed 750 ms/token, then §4 doesn't explain it and there's a second regression worth chasing separately.

**Minor.** Would it be worth stating the cost as "+1 RTT per accepted call" rather than "~2×"? `320 ≈ 2 × 159` because negotiation RTT ≈ exchange RTT over this relay; on a LAN or a slower relay the ratio moves but the added round trip doesn't. Might save someone concluding it didn't reproduce.

PR #107, opened 2026-08-18
perf(p2p): close the native extra round trip with V1Lazy negotiation

RezaRassool:
Closes the 0.6 cutover blocker written up in #100: an accepted unary call cost
two round trips on the native stack where p2pd costs one, so flipping
`native_p2p` on by default would have slowed every remote inference ~57%.

Found with `p2p probe` (#99).

## The change

**`V1Lazy` outbound substream negotiation** (commit 1). The default `V1`
upgrade writes the multistream-select proposal, waits for the peer to confirm
it, and only then sends the payload. `V1Lazy` flushes the proposal together
with the first application data — the optimistic negotiation go-libp2p has
always done. Measured on rezas-mini-2 against metro-win, same peer, same
protocols, switching only `native_p2p`:

|                | p2pd        | native V1   | native V1Lazy |
|----------------|-------------|-------------|---------------|
| accepted call  | 159.4 ms    | 320.1 ms    | **158.0 ms**  |
| inference      | 478 ms/tok  | 750 ms/tok  | **483 ms/tok**|

Interop-safe: the wire bytes are identical, and a *listener* behaves as V1
regardless, so only our dialer changes.

**Keeping refusals precise** (commit 2). Laziness moves *where* a refusal
shows up, and two call sites had to be taught the new shape to keep the
contracts they already had.

## Where the refusal actually surfaces

Worth reviewing carefully, because it is not where you would first look.
`poll_read` drives negotiation to completion; `poll_write` and `poll_flush`
do not. So for a unary call the upgrade **succeeds**, the write and flush
pass, and the *read* fails — the refusal never reaches
`on_dial_upgrade_error`, it lands in the exchange worker as a wire failure. A
peer that simply does not serve a handler therefore read as a broken stream,
and the health probe lost the clean refusal it depends on.

Nothing is actually lost: `NegotiationError` converts into an `io::Error`
that keeps the original enum as its inner error. One trap — the walk has to
step into `io::Error::get_ref()` explicitly, because `io::Error::source()`
reports the inner error's *own* source and skips straight past the variant we
want. There is a unit test pinning that specific trap.

Only `NegotiationError::Failed` is folded into `UnsupportedProtocol`;
`ProtocolError` still means the stream broke and still reads as a wire
failure.

## The sentinel, and why raw streams stay eager

`open_stream` has to decide `UnsupportedProtocol` *before* the caller is
handed a stream, or the control socket enters pipe mode for a protocol the
remote does not serve — a divergence from p2pd right where the cutover claims
parity, and one the socket client cannot recover from because the reply slot
is gone by then. `stream_open_for_an_unserved_protocol_errors_without_entering_pipe_mode`
and four sibling tests encode this.

The version override is swarm-global (no per-substream API in libp2p-swarm
0.47), so raw streams cannot simply opt out. But multistream-select only
takes the lazy shortcut for the **last** protocol it has to offer, so
outbound raw negotiation now appends a single sentinel nobody serves. Every
real protocol stays on the eager path at the same one round trip it already
paid — which a long-lived stream pays exactly once — and reaching the
sentinel means everything real was refused, so the handler turns it straight
back into a refusal. It is never advertised inbound, and never proposed on
the wire in the accepted case.

Alternatives considered and rejected: a zero-length read to force negotiation
**hangs** on success (yamux returns `Pending` when the buffer yields nothing);
vendoring libp2p-swarm to plumb a per-substream version is cleaner but pulls
a large core crate into `patches/`; and letting raw refusals go lazy breaks
the contract above.

## Verification

- `kwaai-p2p` 85 unit + all integration tests green; `kwaai-p2p-daemon` green
  (the 3 control-server tests that broke under plain V1Lazy now pass)
- **Interop tiers 07–13 run manually** with `KWAAI_INTEGRATION_TESTS=1` — all
  green, including `12_pipe_mode_interop`. CI never runs these.
- `clippy --all-targets -D warnings` clean on both crates
- `kwaai-rpc`, `kwaai-distributed`, `kwaainet`, `map-server` compile and test
- **No existing test needed changing** — the refusal signal callers see is
  identical to before. New coverage: the error walk (including the
  `io::Error::source()` trap) and a raw-stream candidate list with nothing
  served, the path the sentinel exists for. Both verified to fail without the
  fix.

## Not yet done

The live re-measurement has **not** been redone since the sentinel landed —
the table above is from V1Lazy applied to everything. It should hold: remote
inference goes through `ollama_proxy.rs` `call_unary_handler`, one unary call
per request, which stays fully lazy; the inference-mux raw stream is opened
once per session, so its one round trip does not touch per-token cost. Worth
confirming on real hardware before the cutover.

### DIFF
--- a/core/crates/kwaai-network-tests/tests/08_unary_swarm_interop.rs
+++ b/core/crates/kwaai-network-tests/tests/08_unary_swarm_interop.rs
@@ -58,7 +58,15 @@ impl SwarmNode {
             .expect("tcp transport")
             .with_behaviour(|_| unary::Behaviour::new(unary::Config::default()))
             .expect("behaviour")
-            .with_swarm_config(|c| c.with_idle_connection_timeout(Duration::from_secs(60)))
+            .with_swarm_config(|c| {
+                c.with_idle_connection_timeout(Duration::from_secs(60))
+                    // Must mirror `service.rs`, or these tests exercise the eager
+                    // V1 path and their refusal assertions prove nothing about
+                    // what production actually runs.
+                    .with_substream_upgrade_protocol_override(
+                        libp2p::core::upgrade::Version::V1Lazy,
+                    )
+            })
             .build();
 
         swarm
--- a/core/crates/kwaai-p2p/src/raw_stream.rs
+++ b/core/crates/kwaai-p2p/src/raw_stream.rs
@@ -87,6 +87,30 @@ pub enum RawStreamError {
     Io(String),
 }
 
+/// A protocol nobody serves, appended to every *outbound* raw-stream
+/// negotiation to keep refusals eager.
+///
+/// The swarm negotiates outbound substreams with `Version::V1Lazy` (see
+/// `service.rs`), which wins a round trip per unary call but hands back a
+/// stream before the peer has confirmed the protocol. For a raw stream that is
+/// the wrong trade: [`RawStreamError::UnsupportedProtocol`] has to be decided
+/// *before* the caller is handed a stream, or the control socket enters pipe
+/// mode for a protocol the remote does not serve — a divergence from p2pd, and
+/// one the socket client cannot recover from because the reply slot is gone.
+///
+/// multistream-select only takes the lazy shortcut for the **last** protocol it
+/// has to offer (`dialer_select.rs`, the `protocols.peek().is_some()` branch).
+/// Appending one more entry therefore keeps every real protocol on the eager
+/// path, negotiated exactly as it was before, at the same one round trip: a
+/// long-lived stream pays that once, so there is nothing to win here anyway.
+///
+/// It is a marker, not a protocol. Reaching it means every real protocol was
+/// already refused, so [`Handler`] turns it straight back into
+/// `UnsupportedProtocol` and drops the stream. It is never proposed on the wire
+/// in the accepted case, and never advertised inbound — `listen_protocol` is
+/// built from the registered set alone.
+const NEGOTIATION_SENTINEL: &str = "kwaai.__negotiation_probe__";
+
 /// The result of an [`Behaviour::open_stream`] request: the protocol
 /// multistream-select settled on, plus the stream itself.
 pub type OpenResult = Result<(UnaryProtocol, RawStream), RawStreamError>;
@@ -267,7 +291,12 @@ impl ConnectionHandler for Handler {
         // clients, not by remote behaviour. The unary cap exists because there
         // a remote can open streams faster than handlers retire them.
         if let Some(open) = self.pending_outbound.pop_front() {
-            let protocols = Protocols::new(open.protocols.clone());
+            // Real protocols first, then the sentinel — see
+            // `NEGOTIATION_SENTINEL`. `open.protocols` itself is left alone, so
+            // error messages and bookkeeping only ever mention real names.
+            let mut offered = open.protocols.clone();
+            offered.push(UnaryProtocol::new(NEGOTIATION_SENTINEL));
+            let protocols = Protocols::new(offered);
             let id = self.next_outbound_id;
             self.next_outbound_id += 1;
             self.requested_outbound.insert(id, open);
@@ -296,6 +325,16 @@ impl ConnectionHandler for Handler {
                 protocol: (stream, proto),
                 info,
             }) => match self.requested_outbound.remove(&info) {
+                Some(open) if proto.as_ref() == NEGOTIATION_SENTINEL => {
+                    // Nothing real survived negotiation. `stream` is dropped
+                    // with this arm, which resets it — the remote never sees a
+                    // sentinel it would have to answer.
+                    let names: Vec<&str> = open.protocols.iter().map(|p| p.as_ref()).collect();
+                    trace!(protocols = %names.join(", "), "raw stream refused by the remote");
+                    let _ = open
+                        .reply
+                        .send(Err(RawStreamError::UnsupportedProtocol(names.join(", "))));
+                }
                 Some(open) => {
                     trace!(%proto, "outbound raw stream negotiated");
                     // If the caller has gone away the stream is dropped here,
--- a/core/crates/kwaai-p2p/src/raw_stream.rs
+++ b/core/crates/kwaai-p2p/src/raw_stream.rs
@@ -109,7 +109,17 @@ pub enum RawStreamError {
 /// `UnsupportedProtocol` and drops the stream. It is never proposed on the wire
 /// in the accepted case, and never advertised inbound — `listen_protocol` is
 /// built from the registered set alone.
-const NEGOTIATION_SENTINEL: &str = "kwaai.__negotiation_probe__";
+pub(crate) const NEGOTIATION_SENTINEL: &str = "kwaai.__negotiation_probe__";
+
+/// True for the reserved sentinel name. Callers and remote registrations must
+/// never use it: [`Handler::on_connection_event`] identifies the sentinel by
+/// name, so a caller-supplied `kwaai.__negotiation_probe__` would make
+/// `offered` read `[sentinel, sentinel]`, negotiate the *first* one eagerly and
+/// successfully, then trip the guard on the name — resetting a live stream
+/// under the remote while telling the caller it was unsupported.
+pub(crate) fn is_reserved(proto: &UnaryProtocol) -> bool {
+    proto.as_ref() == NEGOTIATION_SENTINEL
+}
 
 /// The result of an [`Behaviour::open_stream`] request: the protocol
 /// multistream-select settled on, plus the stream itself.
@@ -382,6 +392,10 @@ impl Behaviour {
     /// Start advertising `proto` for inbound raw streams. Idempotent, and
     /// visible to connections that already exist.
     pub fn register_protocol(&mut self, proto: UnaryProtocol) {
+        if is_reserved(&proto) {
+            debug!(%proto, "refusing to register the reserved negotiation sentinel");
+            return;
+        }
         self.inbound_protocols
             .write()
             .expect("raw stream protocol set lock poisoned")
@@ -419,6 +433,12 @@ impl Behaviour {
         protocols: Vec<UnaryProtocol>,
         reply: oneshot::Sender<OpenResult>,
     ) {
+        if protocols.iter().any(is_reserved) {
+            let _ = reply.send(Err(RawStreamError::UnsupportedProtocol(format!(
+                "{NEGOTIATION_SENTINEL} is reserved"
+            ))));
+            return;
+        }
         if protocols.is_empty() {
             let _ = reply.send(Err(RawStreamError::UnsupportedProtocol(
                 "no protocols requested".to_string(),
--- a/core/crates/kwaai-p2p/src/service.rs
+++ b/core/crates/kwaai-p2p/src/service.rs
@@ -247,7 +247,25 @@ impl NetworkService {
                 KwaaiBehaviour::new(kp, &behaviour_config, relay_client)
             })
             .map_err(|e| anyhow::anyhow!("configuring behaviour: {e}"))?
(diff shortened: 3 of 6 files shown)


dazwin (changes requested):
Requesting changes — on four small, mechanical items. The approach is sound: the sentinel design rests on a verified property of the vendored multistream-select (the lazy shortcut fires only for the *last* offered protocol), it adds no wire bytes or round trips on the success path, and it actually fixes a pipe-mode hazard the p2pd path still has (on go's lazy branch with a stale peerstore, `STREAM_OPEN` returns OK before the peer confirmed anything, then dies with a bare socket EOF). The perf win is real. What blocks approval is about a day of contained fixes, none of which touches the architecture.

This review compared every finding against the p2pd baseline (go-libp2p-daemon `v0.5.0.hivemind1`, go-libp2p v0.32.1, go-multistream v0.5.0 — read from source, citations below), since parity with p2pd is the point of the cutover.

## Must fix before merge

**1. The `is_negotiation_failure` doc inverts the actual behavior, and the guard arm it justifies is dead code** (`unary.rs:520`, `:546`).

The doc's "one inherited wrinkle" paragraph claims a peer that hangs up mid-negotiation reads as `NegotiationError::Failed`, i.e. a refusal. That is true only of the *eager* path (`dialer_select.rs::AwaitProtocol`) — which V1Lazy no longer reaches for a single-protocol offer. On the lazy path, `negotiated.rs::State::Expecting` maps EOF to `ProtocolError::IoError(UnexpectedEof)`, and the `From<NegotiationError> for io::Error` impl strips the wrapper entirely (`negotiated.rs:153-157`, `:363-369`) — so a hang-up classifies as `UnaryError::Wire`.

Here's the good news: **that behavior is correct.** go-multistream returns its refusal error only for a literal `na` token; a reset/EOF surfaces on the p2pd path as `failed to negotiate protocol: stream reset` — a wire error, never a refusal (`go-multistream/client.go:137-151`, `basic_host.go:690`). It was the pre-PR native V1 path that was the outlier, conflating hang-up with refusal. The code now matches p2pd; only the comment says otherwise. Left as-is, the next person who "fixes" the code toward the doc breaks parity. Please rewrite the paragraph to state the real split (na → refusal, hang-up → wire, matching p2pd).

Relatedly, the new `StreamUpgradeError::Io(ref e) if is_negotiation_failure(e)` arm at `unary.rs:520` is provably unreachable: `to_stream_upgrade_error` (`libp2p-swarm-0.47.1/src/connection.rs:633-639`) maps `NegotiationError::Failed` to `NegotiationFailed` and only ever puts a bare io error or a `ProtocolError` inside `Io` — never a `NegotiationError`. Under V1Lazy a refusal never reaches `on_dial_upgrade_error` at all (the upgrade *succeeds*); every refusal lands in the exchange via `wire_or_refusal`. Delete the arm, or demote it to a comment pointing at the exchange path.

**2. Reserve the sentinel name** (`raw_stream.rs:328`).

The guard matches on the negotiated name alone, and nothing filters `kwaai.__negotiation_probe__` out of caller-supplied protocol lists (`parse_protocols` drops only empties; `register_protocol` accepts any string). If a remote registers the sentinel name and a caller opens it, `offered` becomes `[sentinel, sentinel]`; the first entry negotiates *eagerly and successfully*, the guard fires on the name, and a live, fully-negotiated stream is reset under the remote while the caller is told `UnsupportedProtocol("kwaai.__negotiation_probe__")` — leaking the sentinel the new test asserts can never reach a caller. One-line fix: reject the name in `register_protocol`/`open_stream` (or key the guard on offered-list position).

**3. Rewrite the `service.rs` comment to state the override's real scope** (`service.rs:265`).

"It applies when the dialer offers a single protocol, which every unary call here does" reads as scoped to unary. The override sits on the swarm pool config and applies to **every** outbound substream of **every** behaviour — ping, identify, kad, autonat, relay, dcutr, upnp, and the raw-stream handler (which this same PR makes offer 2+ protocols, which is why the sentinel exists three files away). Two concrete consequences should be named and accepted in the comment:

- **ping**: `libp2p-ping`'s only transition to `State::Inactive` is the `NegotiationFailed` arm (`ping/handler.rs:201-206`), now unreachable — a peer not serving `/ipfs/ping/1.0.0` is re-pinged every 30 s for the connection's lifetime, and in-flight ping substreams can keep the connection from idling out. Latent today (the go daemon serves ping unconditionally, and every native node composes `ping::Behaviour`), and there is no per-behaviour override in libp2p-swarm 0.47 to scope around it — so this is a documented acceptance, not new machinery.
- **kad**: `on_fully_negotiated_outbound` sets `protocol_status = supported: true` on the first "negotiated" substream (`kad/handler.rs:487-494`) — under V1Lazy that fires before the remote confirms anything, so a kad query to a connected non-kad peer inserts it into the routing table (`ProtocolConfirmed` → `behaviour.rs:2286`) until identify corrects it. Net new versus p2pd: go-libp2p-kad-dht admits routing-table peers only from identify-reported protocols plus a live FIND_NODE probe (`subscriber_notifee.go`, `dht.go:658-690`), never from a merely-opened substream.

**4. Enable the override in the test harnesses** (`tests/unary.rs:47`, `kwaai-network-tests/tests/08_unary_swarm_interop.rs`).

Both build their own swarms with only `with_idle_connection_timeout`, so their refusal assertions still exercise the pre-PR eager path. The lazy refusal chain is covered end-to-end only Rust↔Rust via `service_unary.rs`; nothing automated covers it against go-libp2p/p2pd (the manual tier runs do, but CI never runs those). Two lines each, and a libp2p bump that changes the error chain would otherwise break the classification while every suite stays green.

## Follow-ups (before the 0.6 cutover, not blocking this PR)

- **Identify-gate for lazy unary calls.** go only goes lazy when the peerstore (identify) already says the peer supports the protocol (`basic_host.go:662-677`); a genuinely-unserved protocol takes the eager `SelectOneOf` branch — clean refusal, zero payload bytes. Native V1Lazy is unconditionally lazy, so a refused call ships the full request frame first (up to `MAX_FRAME_LEN` = 10 MiB, ×3 retries on the block-forward path) — precisely the case go avoids. Consulting identify's known remote protocol set before going lazy is go's design and would also close the kad item above at the source. Also worth noting in the docs either way: `UnsupportedProtocol` no longer means "nothing was sent."
- **Stall bound divergence** (informational — parity or better): the 10 s substream-upgrade deadline no longer bounds unary negotiation; a black-holing peer is now bounded by the 30–60 s request timeout. p2pd is worse (no outbound negotiation deadline; `CALL_UNARY` has no deadline at all), so not a blocker — but raw streams keep the 10 s bound via the sentinel's eager path, so the two native paths' stall budgets now silently differ 3–6×.
- **`classify_hop_failure` wording** (`shard_cmd.rs:3502`): it greps for go's `"protocols not supported"`; the native `"remote does not support protocol …"` falls through to `Other`. Same routing outcome (both blacklist), but operator messages and `HopFailureKind` telemetry mislabel clean refusals — and this PR makes native produce that text more often. The same substring family is independently maintained in `block_rpc.rs:197` and `node.rs:1851`. Pre-existing, worth one alignment pass. (The `p2p probe` itself is fine: its `"not support"` substring matches both wordings and the PR's tests pin it.)
- Minor code-quality: the sentinel arm duplicates the `NegotiationFailed` refusal arm's formatting (`raw_stream.rs:231` vs `:328`, with `names.join` computed twice); `wire_or_refusal` on `write_all`/`flush` is unreachable for refusals (`negotiated.rs` `poll_write`/`poll_flush` in `Expecting` delegate straight to the inner io — only the read can surface one), which hides the single load-bearing site; and `a_broken_negotiation_is_not_a_refusal` is vacuous — `ProtocolError(...).into()` strips the `NegotiationError` wrapper (`protocol.rs:444-450`), so no implementation could fail it.
- The live re-measurement the PR body already flags as not yet redone since the sentinel landed.

## Net-new vs. p2pd summary

| Finding | vs. p2pd |
|---|---|
| Hang-up classifies as wire failure | parity (doc wrong, code right) |
| Stall bound 10 s → 30–60 s | parity or better (p2pd: 60 s / unbounded) |
| Pipe-mode refusal race on `STREAM_OPEN` | **better than p2pd** (sentinel fixes a hazard p2pd has) |
| Payload sent before refusal | net new in practice (go's identify gate avoids it) |
| Kad routing-table admission on unconfirmed upgrade | net new (go is identify-gated + live probe) |
| Ping never marks Unsupported | net new (latent — all current fleet peers serve ping) |

RezaRassool:
Thanks — all four must-fixes are in `bac0a09f`. The review was right on every count, including that the code was already correct where the comment said otherwise.

**1. Doc corrected, dead arm deleted.** You're right that the hang-up split matches p2pd and only the comment was wrong. The doc now states it plainly — `na` → refusal, reset/hang-up → wire — notes that the pre-V1Lazy native path was the outlier, and warns against "fixing" it back. The `StreamUpgradeError::Io(..) if is_negotiation_failure(..)` arm is gone; `to_stream_upgrade_error` never puts a `NegotiationError` inside `Io`, so it was unreachable. Replaced with a comment pointing at the exchange path where refusals actually land.

**2. Sentinel reserved.** Rejected in both `register_protocol` and `open_stream` via a single `raw_stream::is_reserved` predicate. Regression test added, and verified it fails without the guard.

One note: I first put the filter in `handle.rs::parse_protocols` and backed it out. That function runs *only* on the open path, so filtering there silently emptied the list and returned "no protocols requested" — hiding the reason rather than naming it. The explicit rejection in `open_stream` is the right place.

**3. `service.rs` scope stated.** The comment now says the override sits on the swarm pool config and applies to every outbound substream of every behaviour, and names both consequences as accepted: `ping` never reaching `State::Inactive`, and `kad` admitting peers to the routing table on an unconfirmed upgrade — flagged explicitly as net new versus p2pd's identify-gated admission. It records that gating laziness on identify closes both at the source.

**4. Both harnesses enable the override.** You were right that their refusal assertions were proving nothing about the production path.

Also took the vacuous-test item while I was in there: `a_broken_negotiation_is_not_a_refusal` now builds the wrapper explicitly, since `ProtocolError(..).into()` unwraps the `NegotiationError` and no implementation could have failed it.

Verified locally: kwaai-p2p 85 unit + all integration green, kwaai-p2p-daemon green, interop tiers 07–13 green under `KWAAI_INTEGRATION_TESTS=1`, clippy `--all-targets -D warnings` clean, fmt clean.

Follow-ups not taken here, as agreed: the identify gate (which would also close the kad item at source), the stall-bound divergence, `classify_hop_failure` wording, and the live re-measurement.

dazwin (approved):
Approving at bac0a09f. All four requested changes landed and each is correct against the sources I verified earlier:

- The dead `Io` guard arm is gone, and the `is_negotiation_failure` doc now states the real split — `na` → refusal, hang-up → wire failure — with the parity rationale and an explicit warning not to "fix" it back. Matches both the vendored `negotiated.rs` mechanics and go-multistream's behavior.
- The sentinel is reserved in both `open_stream` (typed error before any substream) and `register_protocol`, with `the_negotiation_sentinel_is_reserved` covering the exact collision path from the review.
- The `service.rs` comment now states the swarm-wide scope with the ping/kad consequences named and the identify-gate tracked as a follow-up.
- Both test harnesses mirror the production override — `unsupported_protocol_is_a_clean_refusal` and friends now genuinely exercise the lazy path.

Bonus: `a_broken_negotiation_is_not_a_refusal` was rebuilt with an explicit wrapper that keeps the `NegotiationError` in the chain, so it now actually pins the variant match instead of passing vacuously.

Verified locally at this head: `cargo test -p kwaai-p2p` fully green (unit + raw_stream + unary + service suites) and `cargo clippy -p kwaai-p2p --all-targets -- -D warnings` clean.

Follow-ups before the 0.6 cutover, as discussed: the identify gate for lazy unary calls (closes payload-before-refusal and the kad admission at the source), the `classify_hop_failure` wording alignment, and the live re-measurement.

Issue #116, opened 2026-08-20
shard serve: auto-assign ignores configured start_block, then overwrites it


RezaRassool:
## What happens

A node with a pinned range in `config.yaml` is reassigned anyway on restart, and
its configuration is overwritten with whatever the assigner picked.

Observed on `rezas-mini-2` today. Config before restart:

```yaml
blocks: 32
start_block: 0
```

After a plain `kwaainet shard serve --auto-rebalance`:

```
Querying DHT for gap in Llama-3-1-8B-Instruct (32 blocks)…
Auto-assigned blocks [8, 16)
```

```yaml
blocks: 8
start_block: 8
```

The node went from serving the full model to an eighth of it, and the config no
longer records what the operator asked for.

Note `--auto-rebalance` is **not** the cause — that ran separately and did
nothing (`Rebalance check: coverage OK, no move needed`). This is the
*auto-assign on startup* path, which is on by default.

## Why

`shard_cmd.rs:236`:

```rust
let (start_block, end_block, initial_client) = if !args.no_auto
    && (args.auto || args.start_block.is_none())
```

`args.start_block` is the **CLI** flag — `cli.rs:772` documents it as "Override
start_block from config.yaml". So `is_none()` is true whenever `--start-block`
is absent from the command line, **regardless of the config file**. The
configured value is never consulted in the decision, and is then clobbered by
the result.

The practical consequence is that **config-based pinning does not work at all**.
The only ways to hold a range are `--no-auto` or an explicit `--start-block` on
every invocation — neither of which is discoverable from the config file that
appears to control it.

## Suggested fix

Decide on the *effective* range, not the CLI arg: auto-assign only when neither
the CLI flag nor a configured `start_block` is present. Something like

```rust
let pinned = args.start_block.is_some() || cfg.start_block.is_some();
if !args.no_auto && (args.auto || !pinned) { … }
```

which needs `start_block` to become `Option<u32>` in config so "unset" is
distinguishable from "0" — the same three-state problem `native_p2p` just solved
in #114, and for the same reason. `0` is a legitimate value here and currently
indistinguishable from absent.

Worth deciding separately whether auto-assign should ever *write back* to
`config.yaml`. Overwriting an operator's stated intent is surprising even when
the assignment is right.

## Why this matters now

An auto-update restarts every node at roughly the same time. Every one of them
re-queries the DHT unless invoked with `--no-auto`, and none of them respect
[...]

PR #128, opened 2026-08-21
fix(config): stop cargo test from destroying the developer's real config

RezaRassool:
## What happened

`kwaainet rag list` on rezas-mini-2 reported **"No knowledge bases initialised"**. All 23 registrations — `D6`, `Legal`, `Astrophysics`, the lot — were gone from `config.yaml`, along with `inference_url`, `storage`, `vpk_local_port` and `vpk_mode`. The config had shrunk from 7253 to 1144 bytes. The 2.4 GB of KB data was intact on disk; nothing referenced it.

The culprit is `cargo test`.

`set_key` ends in `self.save()` (`config.rs:1030`), and several unit tests call it — two from #116, one added in #127. With no `KWAAINET_HOME` set, `kwaainet_dir()` resolved to the real `~/.kwaainet`, so those tests wrote a `Default`-derived config straight over the live one.

Demonstrated by checksumming the real config around a full suite run:

```
before: 634adb0a…  blocks: 32
test result: ok. 130 passed
after:  a38a5e38…
*** TEST SUITE REWROTE THE REAL CONFIG ***
< blocks: 32          > blocks: 8
< start_block: 0      > start_block: 8
< native_p2p: true    (deleted)
< inference_url: …    (deleted)
< rag_kbs: (23 KBs)   (deleted)
```

## Why it matters beyond the data loss

This is also what made the config look like it was being **rewritten periodically** — the symptom that sent #127 hunting through the rebalancer. Every `cargo test` run silently reset `blocks` and `start_block`. #127's three findings are real and worth keeping, but this was the reported symptom's actual cause.

## The fix

`kwaainet_dir()` resolves to a per-process temp sandbox under `cfg(test)`. `KWAAINET_HOME` still takes precedence, so integration tests are unaffected; non-test builds are unchanged.

The guard is in `kwaainet_dir()` rather than in each test deliberately — a guard every test author must remember is one that eventually gets forgotten, and this failure is silent and lands somewhere other than the test that caused it.

## Verification

```
test result: ok. 132 passed
config before=634adb0afa6e6d5e5d4438cf11186f4b
config after =634adb0afa6e6d5e5d4438cf11186f4b
KBs: 23 -> 23
✅ real config untouched by the full suite
```

Plus two regression tests: the resolved path is never the real home under test, and a `set_key` write demonstrably lands in the sandbox.

## Follow-up worth considering

`set_key` performing IO is surprising for a setter and is what made this reachable. Splitting the mutation from the persistence would remove the hazard at its source, but that is a wider refactor than this fix.

### DIFF
--- a/core/crates/kwaai-cli/src/config.rs
+++ b/core/crates/kwaai-cli/src/config.rs
@@ -17,7 +17,40 @@ pub fn kwaainet_dir() -> PathBuf {
     if let Ok(home) = std::env::var("KWAAINET_HOME") {
         return PathBuf::from(home);
     }
-    dirs_home().join(".kwaainet")
+    // Unit tests must never resolve to the developer's real ~/.kwaainet.
+    //
+    // `set_key` ends in `save()`, and several tests call it. With the real path
+    // that serialises a `Default`-derived config straight over the live one,
+    // silently deleting every key a default does not carry. On this machine
+    // `cargo test` destroyed 23 registered knowledge bases, `inference_url`,
+    // the storage block and the vpk settings — the KB data survived on disk,
+    // but nothing referenced it any more.
+    //
+    // The sandbox is deliberately here rather than in the tests: a guard each
+    // test has to remember is a guard that eventually gets forgotten, and the
+    // failure is silent and off-target. `KWAAINET_HOME` still wins, so
+    // integration tests that want a specific directory keep working.
+    #[cfg(test)]
+    {
+        test_sandbox_dir()
+    }
+    #[cfg(not(test))]
+    {
+        dirs_home().join(".kwaainet")
+    }
+}
+
+/// Per-process temp directory standing in for `~/.kwaainet` under `cargo test`.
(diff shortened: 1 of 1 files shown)


dazwin (approved):
Verified the `#[cfg(test)]` approach is airtight for this crate: `kwaai-cli` is a pure binary crate (no `[lib]` target, no `tests/` directory), so there are no integration tests or doctests that could compile the crate without `cfg(test)` and bypass the sandbox. `KWAAINET_HOME` precedence is preserved, and no existing test reads config back from disk, so the shared per-process sandbox can't leak state between tests. Putting the guard in `kwaainet_dir()` rather than per-test is the right call for exactly the reason the comment gives.

One take-it-or-leave-it nit: the sandbox dir is keyed by PID and never cleaned up, and PIDs get reused — a later run can inherit a stale `config.yaml`. Harmless today since nothing reads it, but a random suffix in the dir name would close that off permanently.

PR #129, opened 2026-08-21
docs(readme): recommend Ollama whole-model serving on Apple Silicon

RezaRassool:
The README told Apple Silicon users they get **30–36 tok/s from llama.cpp + Metal** in three separate places. No published binary can do that.

## Why the claim was wrong

- `llama-cpp` is not in `default = ["storage", "rag"]` (`kwaai-cli/Cargo.toml:122`)
- nothing in `.github/workflows/release.yml` enables it, so no release archive has it
- `kwaainet benchmark` therefore hits `#[cfg(not(feature = "llama-cpp"))] let ran_llama = false;` (`main.rs:1486`) and falls through to candle
- candle's Metal backend is then skipped at runtime — *"Metal GPU available but skipped (decode is 10x slower than CPU)"* (`kwaai-inference/src/lib.rs:175`)

So a stock Mac runs on CPU at single-digit tok/s, while the README promised 36+.

## What the hardware actually does

Measured on an Apple Silicon Mac mini, same model, same prompt:

| Path | Decode |
|---|---|
| candle, Metal GPU | 2.3 tok/s |
| candle, CPU | 4.2 tok/s |
| **Ollama (llama.cpp + Metal)** | **47.3 tok/s** |

A Mac contributing block shards makes the network **slower**, not faster.

## Changes

- New **Current status** section: the v0.6.0 native-libp2p cutover, and the Apple Silicon guidance with the measurements above
- Header bullet, Quickstart §4, benchmark instructions, project status and roadmap row all corrected to match what the code does
- Benchmark guidance now measures Ollama directly, since that is the path a Mac actually serves on

## Accuracy about what has shipped

The macOS auto-detection is #126 — **merged but not released**. Rather than claim v0.6.2 does something it doesn't, the guidance reads "from the next release" and gives the manual equivalent for current users:

```bash
kwaainet config set contribute.shards false
ollama pull llama3.1:8b
```

`contribute.shards` verified as a real key (`config.rs:1046`); `--force-blocks` verified against `shard serve --help`.

## Verification

The documented benchmark snippet was run against a live Ollama and returns **46.8 tok/s**, consistent with the table. It pins `num_ctx` on `/api/chat` deliberately — `/v1/chat/completions` ignores `options`, and leaving the context at Ollama's default pushes part of the model onto CPU and depresses throughput.

Code fences balanced; no stale `36+ tok/s` / `30+ tok/s` / `Dual backend` claims remain.

### DIFF
--- a/README.md
+++ b/README.md
@@ -17,7 +17,7 @@ From an app's point of view, KwaaiNet looks like a familiar chat-completion styl
 
 ## Current status
 
-**Latest release: v0.6.2** (August 2026). Two things are worth knowing before you set up a node.
+**As of v0.6.2.** Two things are worth knowing before you set up a node.
 
 **The networking stack is now native.** As of **v0.6.0** every node runs the in-process
 rust-libp2p stack by default; the Go `p2pd` child process is no longer required. Existing
@@ -198,7 +198,7 @@ curl -s http://localhost:11434/api/chat -d '{
   "model": "llama3.1:8b", "stream": false,
   "options": {"num_ctx": 8192, "num_predict": 80},
   "messages": [{"role": "user", "content": "Explain knowledge graphs in one paragraph."}]
-}' | python3 -c 'import json,sys; d=json.load(sys.stdin); print(f"{d["eval_count"]/(d["eval_duration"]/1e9):.1f} tok/s")'
+}' | python3 -c 'import json,sys; d=json.load(sys.stdin); print("%.1f tok/s" % (d["eval_count"]/(d["eval_duration"]/1e9)))'
 ```
 
 > Pin `num_ctx` explicitly. Ollama defaults to a very large context, which pushes part of the
@@ -235,15 +235,18 @@ The node will connect to bootstrap peers, announce itself on the DHT, auto-detec
 
 > **Pre-release note (< v1.0):** `kwaainet start --daemon` automatically starts shard serving (if a local model is present) and storage serving (if storage has been initialised). This opt-out default keeps the network dense during the insider phase. Run with `--no-contribute` to start the node without contributing, or permanently disable with `kwaainet config set contribute.shards false`.
 >
-> On macOS this contributes **whole-model inference via Ollama** rather than block shards — see [Current status](#current-status).
+> From the next release, on macOS this contributes **whole-model inference via Ollama** rather than block shards — on v0.6.2 see [Current status](#current-status) for the manual equivalent.
 >
-> Note that block sharding is **experimental** but currently opt-out. On Linux/CUDA, if you would
-> rather contribute only whole models through Ollama, turn shard serving off — the node keeps
-> serving over `/kwaai/ollama-proxy/1.0.0` either way:
+> Block sharding is **experimental**. On v0.6.2 it is opt-out — a node with a local model serves
+> blocks unless told otherwise — and it becomes opt-in from the next release. Whole-model serving
+> over `/kwaai/ollama-proxy/1.0.0` is unaffected either way. To be sure a node contributes only
+> whole models, whichever release it is on:
 >
 > ```bash
 > kwaainet config set contribute.shards false
 > ```
+>
+> To serve blocks once it is opt-in, set that to `true` or pass `--shard`.
 
 ### 3. Call the OpenAI-compatible API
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -138,13 +183,28 @@ kwaainet benchmark --gpu
...
 ```bash
-ollama pull llama3.1:8b    # download a GGUF model
-kwaainet benchmark         # auto-detects GGUF → 36+ tok/s via Metal
+ollama pull llama3.1:8b
+
+curl -s http://localhost:11434/api/chat -d '{
+  "model": "llama3.1:8b", "stream": false,
+  "options": {"num_ctx": 8192, "num_predict": 80},
+  "messages": [{"role": "user", "content": "Explain knowledge graphs in one paragraph."}]
+}' | python3 -c 'import json,sys; d=json.load(sys.stdin); print(f"{d["eval_count"]/(d["eval_duration"]/1e9):.1f} tok/s")'
```
dazwin: This is a `SyntaxError` on Apple's system Python (3.9.6) — nesting double quotes inside a double-quoted f-string needs Python 3.12+ (PEP 701), and stock macOS is exactly the audience for this section:

```
$ /usr/bin/python3 -c 'import json,sys; d=json.load(sys.stdin); print(f"{d["eval_count"]/(d["eval_duration"]/1e9):.1f} tok/s")'
  File "<string>", line 1
SyntaxError: f-string: unmatched '['
```

Version-proof alternative (works on 3.6+, keeps the outer single quotes):

```suggestion
}' | python3 -c 'import json,sys; d=json.load(sys.stdin); print("%.1f tok/s" % (d["eval_count"]/(d["eval_duration"]/1e9)))'
```

README.md
```diff
@@ -174,6 +234,16 @@ kwaainet start --daemon
 The node will connect to bootstrap peers, announce itself on the DHT, auto-detect available hardware, and appear on [map.kwaai.ai](https://map.kwaai.ai). No Python, no build tools, no manual configuration required.
 
 > **Pre-release note (< v1.0):** `kwaainet start --daemon` automatically starts shard serving (if a local model is present) and storage serving (if storage has been initialised). This opt-out default keeps the network dense during the insider phase. Run with `--no-contribute` to start the node without contributing, or permanently disable with `kwaainet config set contribute.shards false`.
+>
+> On macOS this contributes **whole-model inference via Ollama** rather than block shards — see [Current status](#current-status).
```
dazwin: Present tense overstates this: per the **Current status** section above, on v0.6.2 (the release this Quickstart installs) a Mac still serves blocks — the Ollama redirect is #126, merged but unreleased. Suggest qualifying it the way Current status does, e.g. "From the next release, on macOS this contributes whole-model inference via Ollama rather than block shards — on v0.6.2 see [Current status](#current-status) for the manual equivalent."

The "Per-platform backends" bullet under *Project status* has the same present-tense slip ("Apple Silicon serves whole models through Ollama").

dazwin (changes requested):
The rewrite is honest and well-sourced — I verified `/kwaai/ollama-proxy/1.0.0` exists (`ollama_proxy.rs:37`), `shard serve --force-blocks` exists, and the macOS auto-detect (#126) is merged-but-unreleased, so the "from the next release" framing in **Current status** is accurate.

Requesting changes for one thing only: the benchmark one-liner is a `SyntaxError` on stock macOS Python — details inline. Two consistency notes inline as well, plus one coordination point:

This PR states sharding is "experimental but currently opt-out" (Quickstart §2), while #131 makes it opt-in. Whichever merges second leaves the README wrong — probably worth a note on both PRs so the follow-up doc touch-up doesn't get lost.

Minor: "Latest release: v0.6.2 (August 2026)" is the kind of line that goes stale silently; consider "As of v0.6.2" phrasing that stays true after the next release.

RezaRassool:
Thanks — all points taken, pushed.

**The Python one-liner was the important catch.** You are right, and my verification was the problem: I ran that line against my own interpreter (3.13.9), it printed 46.8 tok/s, and I wrote it down as verified. Stock macOS is 3.9.6 and PEP 701 is 3.12+. Taken your `%`-formatting version and checked it on `/usr/bin/python3` — 47.3 tok/s.

Both present-tense slips fixed to match the qualified phrasing in Current status. `Latest release:` → `As of v0.6.2`.

**On the coordination point** — rather than leave a follow-up to be lost, I made the wording correct in both worlds: it now says sharding is opt-out on v0.6.2 and opt-in from the next release, and the `contribute.shards false` command is valid either way. Neither merge order leaves this wrong.

PR #130, opened 2026-08-21
feat(rpc): whole-network peer view and dial-a-known-peer over the Session stream

dazwin:
Adds the `Network` session op at **frame tag 17** and a `Connect` op at **tag 18**, giving a client a whole-network view of a node's p2p state and a way to dial a peer the node knows about but is not connected to.

### Why now

`Kwaai-AI-Lab/KwaaiNetGUI` `main` already ships the Peers tab (`lib/src/ui/pages/peers_tab.dart`), and its generated bindings reference tags 17 and 18. Released KwaaiNet **v0.6.2** stops at tag 16 (`storage`), so the GUI's Peers tab currently has no daemon that understands its frames. This PR closes that gap — it is the remaining blocker for shipping the GUI against an upstream KwaaiNet release.

The GUI's bindings and this proto were verified to match field-for-field and tag-for-tag across all seven new messages (`NetworkRequest`, `NetworkUpdate`, `ConnectRequest`, `ConnectReply`, `ConnectedPeer`, `RoutingPeer`, `SelfStatus`).

### What it adds

**`kwaai-p2p`** — the swarm exposes RTT, agent version, served protocols, DCUtR upgrades, the relay an inbound connection arrived through, routing-table addresses, and DHT client/server role, as one whole-network snapshot.

**`kwaai-rpc`** — `NetworkRequest`/`NetworkUpdate` at tag 17 (one-shot or subscribe with an interval), `ConnectRequest`/`ConnectReply` at tag 18.

**`kwaai-cli`** — serves both ops from the native swarm; a new `peers_view` module holds the classification logic (bootstrap/relay identification, connection kind, DHT role) shared with `kwaainet p2p peers`.

Two fixes fell out of the work: the swarm no longer re-dials peers it is already connected to, and a transportless connection is no longer reported as direct.

### Tag allocation

17 and 18 are unused on `main` (which allocates through 16). No existing tag changes meaning.

### Non-native path

`Network` returns `UNIMPLEMENTED` rather than `UNAVAILABLE` when the node is on the Go p2p path — the handle slot will never fill there, so a client retrying `UNAVAILABLE` would retry forever. During native startup, when the slot is about to fill, it returns `UNAVAILABLE` so retrying is correct. Read through `KwaaiNetConfig::native_p2p()` so the three-state flag from #114 resolves properly.

### Testing

Rebased onto `main` at `b359162f`. `cargo fmt --check` clean, `cargo clippy --workspace --all-targets -- -D warnings` clean, `cargo test --workspace` passes with 0 failures. Proto round-trip coverage extended for the new frames; `kwaai-p2p/tests/swarm.rs` covers the snapshot.

### DIFF
--- a/core/crates/kwaai-cli/src/grpc_server.rs
+++ b/core/crates/kwaai-cli/src/grpc_server.rs
@@ -1314,8 +1314,11 @@ fn network_identity(u: &NetworkUpdate) -> NetworkIdentity {
             // Every field a client renders per row *except* rtt_ms — see the
             // type comment. Protocols are joined rather than nested so the
             // whole row stays one comparable string.
+            // `dcutr` is in here deliberately: an upgrade flips a connection
+            // from relayed to direct, which is exactly the kind of change the
+            // user is watching this page for.
             format!(
-                "{}|{}|{}|{}|{}|{}|{}|{}",
+                "{}|{}|{}|{}|{}|{}|{}|{}|{}",
                 p.peer_id,
                 p.addr,
                 p.kind,
@@ -1324,13 +1327,14 @@ fn network_identity(u: &NetworkUpdate) -> NetworkIdentity {
                 p.is_trusted_relay,
                 p.protocols.join(","),
                 p.agent_version,
+                p.dcutr,
             )
         })
         .collect();
     let routing = u
         .routing
         .iter()
-        .map(|r| format!("{}|{}", r.peer_id, r.connected))
+        .map(|r| format!("{}|{}|{}", r.peer_id, r.connected, r.is_bootstrap))
         .collect();
 
     (
@@ -1564,6 +1568,7 @@ fn build_network_update(
                     .map(|d| d.as_millis().min(u32::MAX as u128) as u32)
                     .unwrap_or(0),
                 agent_version: p.agent_version.clone().unwrap_or_default(),
+                dcutr: p.dcutr,
             };
             (
                 group_index(is_bootstrap, is_trusted_relay, kind),
@@ -1583,6 +1588,7 @@ fn build_network_update(
         .map(|p| RoutingPeer {
             peer_id: p.to_base58(),
             connected: connected_ids.contains(p),
+            is_bootstrap: bootstraps.contains(p),
         })
         .collect();
     routing.sort_by(|a, b| a.peer_id.cmp(&b.peer_id));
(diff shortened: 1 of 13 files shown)


RezaRassool (approved):
Approving. I verified the claims rather than taking them on trust, and everything checked out — details below, plus two minor observations that are not blockers.

## Verified

**Tag allocation.** `main` allocates through 16 on *both* the request and reply oneofs; 17 and 18 are genuinely free and no existing tag changes meaning. All seven new messages are present.

**Rebase is stale but harmless.** The PR says it was rebased at `b359162f`; `main` has since taken four merges (#128, #133, #132, #129), two of which touch `kwaai-cli` config handling. I merged current `main` into the branch locally — clean, no conflicts, and:

```
cargo build -p kwaainet                   clean
cargo test -p kwaainet --bin kwaainet     159 passed
cargo test -p kwaai-p2p -p kwaai-rpc      all passed
cargo clippy (kwaainet, p2p, rpc)         clean
```

Worth noting #133 made `contribute.shards` an `Option<bool>` and #132 removed the implicit `save()` from `set_key`; neither interacts badly with this.

**No new panics in production code.** Every `unwrap`/`expect` the diff adds is inside a `#[test]`.

**`Connect` is not remotely reachable.** I checked specifically, since the op lets a caller make the node dial an arbitrary peer: the gRPC server binds `127.0.0.1` plus a Unix socket, so there is no remote dial-abuse surface.

## What I liked

The `UNIMPLEMENTED` vs `UNAVAILABLE` split is the right call and the comment explains *why* rather than *what* — a client retrying `UNAVAILABLE` on the Go path would spin forever, and the distinction is invisible without that reasoning.

`classify_addr` treating a transportless address as relayed rather than direct is the correct default, and the justification is the strongest part of the PR: calling an unclassifiable address direct claims the more surprising thing on the least evidence.

The `AlreadyConnected` relay-backoff interaction is a subtle catch — without the `continue`, a redundant dial would have backed off a perfectly good relay candidate, and the comment correctly identifies the identify-timing window that makes the arm reachable at all.

19 new tests, behaviour-named, covering the edge cases that actually bite: transportless addresses, named-network kad suffixes, malformed config entries, and identity churn suppression.

## Two minor observations, neither blocking

**Config is read once per subscription.** `bootstrap_peer_ids()` and `trusted_relay_peer_ids()` are resolved before the loop, so a long-lived GUI subscription will not notice edits to `initial_peers` or `trusted_relays`. That is arguably consistent — those need a node restart to take effect anyway — but if a subscription can outlive a config change, moving them inside the loop is cheap.

**The GUI compatibility claim is the one thing I cannot verify from this repo.** It is also the stated motivation, and the failure mode if the field-for-field check drifted is a silently broken Peers tab. Worth recording which `KwaaiNetGUI` commit the bindings were compared against, so a future reader can tell whether the check still holds.

Issue #117, opened 2026-08-20
shard serve: auto-assign hands Mac nodes partial shards Metal cannot serve


RezaRassool:
## What happens

The block assigner has no notion of platform. It handed a Mac a partial
`[8, 16)` range exactly as it would a CUDA box.

Sharded inference is not supported on Metal at driver level — Mac GPU nodes can
serve a **whole model** or fall back to CPU, and nothing else. So a partial
assignment is not merely suboptimal, it is unservable on the GPU the node has.

Observed today on `rezas-mini-2` (M-series, Metal available):

```
Auto-assigned blocks [8, 16)
WARN kwaai_inference: Metal GPU available but skipped (decode is 10x slower than CPU).
INFO kwaai_inference: Using CPU for inference
Device: CPU
```

The node advertised eight blocks of a model it could only serve from CPU. Per
earlier benchmarking the GPU/CPU gap on comparable hardware is roughly 29 tok/s
against 0.79 — so a peer routing to those blocks gets an order-of-magnitude
worse path than the chain implies, with nothing in the DHT record to say so.

## Why

Nothing in the assign path in `shard_cmd.rs` consults platform, accelerator, or
a whole-model-only capability. A search for `metal`, `whole model`, or
`target_os = "macos"` across that file finds only unrelated strings.

## Suggested fix

Two parts, and the second matters more than the first:

1. **Exclude Metal-only nodes from gap-filling**, or constrain them to the whole
   range. A Mac should auto-assign `[0, total)` or decline to auto-assign at all,
   not take a slice.
2. **Advertise the constraint.** Whatever a node can actually serve should be
   visible to peers choosing a route. Today the chain shows a Mac holding
   `[8, 16)` with no indication that path is CPU-bound.

Related: sharded inference on Metal is blocked on driver support and paused
since Feb 2026, so this is not a temporary state to wait out. The assigner needs
to know that some nodes are whole-model-or-nothing.

## Why this matters now

The 0.6 cutover restarts the fleet, and every node re-queries the DHT on restart
(see the companion issue on `start_block`). Mac nodes are a large share of the
network; a coordinated restart could quietly move several of them onto partial
shards they cannot GPU-serve, degrading routes that look healthy in
`shard chain`.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

dazwin:
[...]

PR #131, opened 2026-08-21
feat(shard): make experimental block-shard serving opt-in

RezaRassool:
> **Stacked on #128.** Base is `fix/tests-must-not-write-real-config`, so this diff shows only the opt-in change. GitHub will retarget to `main` when #128 merges. (I stacked rather than branching from `main` because without #128's sandbox guard, running this crate's tests destroys the developer's real config — it wiped 23 KB registrations on my machine twice today.)

## The problem

`kwaainet start --daemon` enrolled every node in block-shard serving unless it opted out. Sharding is experimental, and on Apple Silicon it does not work at all (#117):

| Path | Decode |
|---|---|
| candle, Metal GPU | 2.3 tok/s |
| candle, CPU | 4.2 tok/s |
| **Ollama (llama.cpp + Metal)** | **47.3 tok/s** |

So the opt-out default signed up every Mac for the one path that makes the network *worse* — a block server 20× slower than what the same machine could serve, advertising coverage it fills badly.

## The change

`contribute.shards` becomes `Option<bool>` defaulting to off, matching how `native_p2p` and `start_block` already resolve.

The three states are load-bearing. A plain `bool` flip cannot distinguish "never chosen" from "explicitly set", so it would silently override operators who deliberately asked for sharding. With `Option`, an existing `contribute.shards: true` keeps serving blocks across the upgrade.

```rust
pub const DEFAULT_CONTRIBUTE_SHARDS: bool = false;

pub fn shards(&self) -> bool { self.shards.unwrap_or(DEFAULT_CONTRIBUTE_SHARDS) }
pub fn shards_unset(&self) -> bool { self.shards.is_none() }
```

## This reduces block coverage, not contribution

Worth being explicit, because "contribution off by default" would be the wrong reading. Whole-model serving over `/kwaai/ollama-proxy/1.0.0` is registered by the node itself on every platform (`node.rs:340`, `node_native.rs:687`), independent of shard serving. A node with `shards` unset still serves inference — and on a Mac it serves it ~20× faster than the blocks it stopped advertising.

Expect `shard chain` block coverage to drop as nodes upgrade. That is the intended effect: the coverage being lost is largely coverage that was slowing chains down.

## Not silent

A node that has a local model and would have sharded under the old default says so once at startup:

```
Block-shard serving is now opt-in (experimental) — not starting it.
This node still contributes whole-model inference via Ollama.
Serve blocks anyway: kwaainet config set contribute.shards true
```

## Tests

Eight regression tests: the default; explicit opt-in and opt-out each staying distinguishable from unset; the pre-flip upgrade path (`contribute.shards: true` keeps working); `--no-contribute` still winning over an explicit opt-in; storage contribution untouched; and the unset value staying out of serialised YAML.

140 tests pass, clippy clean, and the real config is verified untouched by the run.

### DIFF
--- a/core/crates/kwaai-cli/src/config.rs
+++ b/core/crates/kwaai-cli/src/config.rs
@@ -342,27 +342,64 @@ pub struct ContributeConfig {
     #[serde(default = "default_true")]
     pub storage: bool,
 
-    /// Automatically start shard serving on daemon start (when a model is available).
-    #[serde(default = "default_true")]
-    pub shards: bool,
+    /// Automatically start **experimental** block-shard serving on daemon start.
+    ///
+    /// **Opt-in, and deliberately three-state.** `Some(true)` serves blocks,
+    /// `Some(false)` is an explicit refusal, and `None` means the operator has
+    /// never chosen — which now resolves to [`DEFAULT_CONTRIBUTE_SHARDS`],
+    /// i.e. off.
+    ///
+    /// This used to default to on. Block sharding is experimental: it does not
+    /// work on Apple Silicon at all (#117), where a node serving blocks runs
+    /// ~20x slower than the same machine serving whole models through Ollama —
+    /// so the opt-out default enrolled every Mac into the one path that makes
+    /// the network worse. Turning contribution *off* by default would be the
+    /// wrong reading of this change: whole-model serving over
+    /// `/kwaai/ollama-proxy/1.0.0` is registered by the node itself and is
+    /// unaffected, so a node with `shards: None` still contributes inference.
+    ///
+    /// It is an `Option` for the same reason `native_p2p` is: flipping a plain
+    /// `bool` default cannot distinguish "never chosen" from "explicitly set",
+    /// so it would silently override operators who deliberately asked for
+    /// sharding. Read it through [`ContributeConfig::shards`], never directly.
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    pub shards: Option<bool>,
 
     /// Automatically install updates when a new version is available (pre-v1.0 default: true).
     #[serde(default = "default_true")]
     pub auto_update: bool,
 }
 
+/// Block-shard serving is opt-in: experimental, and actively harmful on Metal.
+pub const DEFAULT_CONTRIBUTE_SHARDS: bool = false;
+
+impl ContributeConfig {
+    /// Whether to serve block shards — the explicit choice if there is one,
+    /// otherwise [`DEFAULT_CONTRIBUTE_SHARDS`].
+    pub fn shards(&self) -> bool {
+        self.shards.unwrap_or(DEFAULT_CONTRIBUTE_SHARDS)
+    }
+
+    /// True when the operator has said nothing about shard serving, so the
+    /// opt-in default applies. Used to explain the change once at startup
+    /// rather than silently dropping a node's block contribution.
+    pub fn shards_unset(&self) -> bool {
+        self.shards.is_none()
+    }
+}
+
 impl Default for ContributeConfig {
     fn default() -> Self {
         Self {
             storage: true,
-            shards: true,
+            shards: None,
             auto_update: true,
         }
     }
 }
 
 fn contribute_config_is_default(c: &ContributeConfig) -> bool {
-    c.storage && c.shards && c.auto_update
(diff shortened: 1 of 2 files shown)


core/crates/kwaai-cli/src/main.rs
```diff
@@ -322,6 +322,19 @@ async fn main() -> Result<()> {
...
+                } else if !policy.shards
+                    && !args.no_contribute
+                    && cfg.contribute.shards_unset()
+                    && shard_available
+                {
+                    // This node has a model and would have served blocks under
+                    // the old opt-out default. Say so once, rather than letting
+                    // its block contribution vanish silently on upgrade.
+                    print_info(
+                        "Block-shard serving is now opt-in (experimental) — not starting it.",
```
dazwin, line 342: This branch is reachable by a user who explicitly typed `--shard`. `args.shard` feeds `shard_available` and `enough_ram` (lines 304–306) but never `policy.shards`, which was fine while the config default was `true` — now, with `contribute.shards` unset, `kwaainet start --daemon --shard` lands here and prints "not starting it", telling the user to go set a config key for the thing they just asked for on the command line.

Suggest treating the flag as an explicit opt-in:

```rust
let serve_shards = policy.shards || (args.shard && !args.no_contribute);
```

and using `serve_shards` in place of `policy.shards` in this `if`/`else if` chain.

core/crates/kwaai-cli/src/config.rs
```diff
@@ -1443,3 +1480,87 @@ mod test_isolation {
...
+    fn no_contribute_still_overrides_an_explicit_opt_in() {
+        let mut cfg = KwaaiNetConfig::default();
+        cfg.set_key("contribute.shards", "true").expect("set");
+        let policy = cfg.contribute_policy(true);
+        assert!(!policy.shards, "--no-contribute wins over config");
+        assert!(!policy.storage);
+    }
+
+    #[test]
+    fn unset_shards_is_omitted_from_yaml() {
```
dazwin, line 1574: This passes trivially: a default config hits `contribute_config_is_default`, so the whole `contribute:` block is skipped and the assertion would hold even without the new `skip_serializing_if` on `shards`. Setting e.g. `storage: false` first would force the block to serialize and actually exercise the `Option::is_none` skip.

dazwin (changes requested):
The design is right — `Option<bool>` three-state matching the `native_p2p` precedent, the upgrade path for an explicit `contribute.shards: true` is preserved, and since the field type changed, the compiler guarantees no reader was missed (I grepped; `contribute_policy` and `set_key` are the only ones). Test coverage is good.

One real behavioral bug, inline below: `kwaainet start --daemon --shard` no longer starts shard serving when `contribute.shards` is unset, even though the flag's help text is "Also start the shard inference server".

Related but smaller: `kwaainet restart` (main.rs, `contribute_policy(false)`) will now drop shard serving for a node that was started with `--shard` on an unset config. Root cause is the same — `--shard` is never persisted — and it predates this PR in spirit, but the flip widens it from "low-RAM forced nodes" to every unset-config node, so it's worth a mention in the PR description at least.

Nit: the new startup message asserts "This node still contributes whole-model inference via Ollama" unconditionally, even when Ollama isn't installed or has no model pulled. "can still contribute … via Ollama" would be accurate in both worlds.

RezaRassool:
Thanks — all points taken, pushed.

**The `--shard` regression is real and was the right catch** — the flag fed `shard_available` and `enough_ram` but never the policy, so it landed in the new notice and advised setting a config key for the thing the user had just typed.

Taken your suggestion, with one addition: `--shard` now also beats an *explicitly* false config, on the reasoning that a flag typed now is a later decision than a config file. `--no-contribute` still outranks both. Extracted as `ContributePolicy::serve_shards` rather than left inline so it is testable — six tests cover the truth table.

**On `restart`:** confirmed, and I have added it to the PR description. `--shard` is never persisted, so `contribute_policy(false)` on restart drops shard serving for a node started with the flag. The flip widens it from low-RAM forced nodes to every unset-config node. I have left it unfixed here because the fix is a design choice — persist `--shard` as `contribute.shards: true`, or have `restart` carry the original argv — and both are bigger than this PR.

`unset_shards_is_omitted_from_yaml` was indeed passing trivially; it now sets `storage: false` to force the block to serialize, and asserts that as a precondition. Notice softened to "can still contribute".

PR #132, opened 2026-08-21
refactor(config): set_key mutates, callers persist

RezaRassool:
> **Stacked on #131** (which is stacked on #128). Base is `feat/shard-serving-opt-in` so this diff shows only the refactor. GitHub retargets as the stack merges.

## The hazard

`set_key` ended in `self.save()` — assigning one field wrote the whole config to disk as a side effect:

```rust
_ => anyhow::bail!("Unknown config key '{}'…", key),
}
self.save()      // ← a setter performing I/O
```

That is precisely what let unit tests destroy the real config. A test doing:

```rust
let mut cfg = KwaaiNetConfig::default();   // empty of everything you configured
cfg.set_key("start_block", "0")            // …saves it over ~/.kwaainet/config.yaml
```

serialises a *default* config over the live one, so it does not edit one key — it replaces the file. On my machine that deleted 23 KB registrations, `inference_url`, the storage block and the vpk settings, twice. Nine test call sites do this.

## Guard vs. cause

#128 sandboxes `kwaainet_dir()` under `cfg(test)`, so no test can reach the real path whatever it calls. **Keep that** — it cannot be forgotten by the next test author.

This PR removes the hazard itself. A setter should not do I/O.

## Blast radius: one caller

Exactly one production call site, `config set`, which now persists explicitly. Verified end to end against an isolated `KWAAINET_HOME`:

```
✅ Set blocks = 16
blocks: 16                    # persisted
🧱 blocks:         16         # reads back
Error: Unknown config key 'bogus_key'…
unchanged after error: YES
```

Separating mutation from persistence also makes it possible to validate several keys before committing any — impossible while every assignment wrote immediately.

## A pre-existing race this surfaced

I first added a companion test asserting `save()` lands in the sandbox. It passed alone and failed in the suite.

Cause: `grpc_server`'s tests set `KWAAINET_HOME` **process-wide** (`grpc_server.rs:1602`), serialised by a mutex private to that module — so it does not serialise against anything else. A concurrent `save()` gets redirected into their temp dir between the write and the assertion.

I removed the racy test rather than paper over it, and recorded why in a comment where it would have gone. **This is a live hazard beyond my test**: any test in this crate that touches `HOME` or `KWAAINET_HOME`-derived paths can be perturbed by the grpc module. Worth a crate-wide test lock, but that is a separate change.

## Verification

140 tests pass across **three consecutive full runs**, with the real config verified byte-identical after each.

### DIFF
--- a/core/crates/kwaai-cli/src/config.rs
+++ b/core/crates/kwaai-cli/src/config.rs
@@ -1057,6 +1057,18 @@ impl KwaaiNetConfig {
     }
 
     /// Set a top-level key by name (string value coerced to the right type).
+    ///
+    /// **Mutates only — the caller persists.** This used to end in `save()`,
+    /// which made a setter perform disk I/O: every caller wrote the whole
+    /// config to `~/.kwaainet/config.yaml` as a side effect of assigning one
+    /// field. Unit tests calling `set_key` on a `KwaaiNetConfig::default()`
+    /// therefore serialised an empty config over the developer's real one,
+    /// destroying every key a default does not carry — on one machine, 23
+    /// registered knowledge bases, `inference_url`, the storage block and the
+    /// vpk settings.
+    ///
+    /// Separating the two also lets a caller validate several keys before
+    /// committing any of them, which the implicit save made impossible.
     pub fn set_key(&mut self, key: &str, value: &str) -> Result<()> {
         match key {
             "model" => self.model = value.to_string(),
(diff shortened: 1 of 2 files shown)


RezaRassool:
Rebased. #131 was auto-closed when #128's base branch was deleted on merge and GitHub would not reopen it, so the opt-in change now lives in **#133** — this PR is stacked on that instead. Same commits, same content; @dazwin's review on #131 is addressed there.

Stack is now: #133 (sharding opt-in) → #132 (this).

Issue #108, opened 2026-08-18
native p2p: unfiltered peer addresses make a node dial itself on 127.0.0.1


RezaRassool:
## What happens

On the native stack (`native_p2p: true`), outbound calls intermittently fail with:

```
WARN kwaai_p2p_daemon::persistent: Daemon error for call 80069242-…:
  Dial failed: Dial error: Unexpected peer ID 12D3KooWAourfFoxBjBiXWHdZjxPAuenuyYLFjxHa1C9zknjYA7A
  at /ip4/127.0.0.1/tcp/8080/p2p/12D3KooWLMizEbViSoL4WGJUMsLVRyLccyymosX36MDKdbYgGFzE
```

Read that address carefully: we are dialing **metro-win** (`12D3KooWLMizEbVi…`)
at **`127.0.0.1:8080`**, and the peer ID that answers
(`12D3KooWAourfFox…`) is **our own node**. Port 8080 is our own listen port.
So the node dials itself, notices the peer ID does not match the one it wanted,
and fails the call.

Observed 12 times in ~20 minutes during a D6 graph build, alongside 4 ×
`Daemon error: Network not initialized`. It is self-correcting — the retry
picks a working address — so it shows up as latency and noise rather than a
hard failure. It does not appear on the p2pd path.

## Why it happens

`unary.rs` and `raw_stream.rs` both keep a `libp2p::swarm::behaviour::PeerAddresses`
cache fed straight from the swarm:

- `unary.rs:797` — `self.addresses.on_swarm_event(&event);`
- `raw_stream.rs:471` — same

and hand the whole cache back as dial candidates:

```rust
// unary.rs:783-794
fn handle_pending_outbound_connection(…) -> Result<Vec<Multiaddr>, ConnectionDenied> {
    Ok(match maybe_peer {
        Some(peer) => self.addresses.get(&peer).collect(),
        None => Vec::new(),
    })
}
```

`on_swarm_event` absorbs `NewExternalAddrOfPeer`, so **any** address a peer
advertises — via identify or the DHT — becomes a dial candidate verbatim. A
peer that advertises its own `127.0.0.1` (which nodes routinely do; it is the
*listener's* job not to announce it, and not every implementation is careful)
gets that loopback address dialed by *us*, against *our* machine.

The filter for exactly this already exists and is already correct —
`addresses.rs::is_routable_v4` rejects loopback at line 83, and
[...]

PR #137, opened 2026-08-23
fix(p2p): routed dials fall back to a DHT walk and never dial our own address

dazwin:
On the native stack, a unary or stream request addressed by a bare `PeerId` failed on the **first** call to any peer with `WrongPeerId`, and a NATed peer failed on every call. The same peer was discoverable, pingable, and reachable via `p2p peers connect` — which looks the peer up in the DHT before dialing. This is what made the VPK storage protocol (`/kwaai/storage/1.0.0`) look unreachable while everything around it looked healthy.

Two separate defects, one code path.

## The path

`NetworkService::dispatch_routed` (`service.rs`) is where a call by peer id enters the swarm. On `main` it forwarded the request the moment the routing table held *any* address for the peer:

```rust
if self.swarm.is_connected(&peer) || !self.known_addresses(&peer).is_empty() {
    self.forward_routed(peer, request);   // -> unary.send_request / raw_stream.open_stream
    return;
}
```

`forward_routed` hands the request to the behaviour, and the behaviour's dial is **open to the behaviours' address suppliers**. It never uses the address list we just checked.

### 1. Our own address, filed under someone else's id

`kad::Behaviour::discovered` inserts whatever `multiaddrs` a remote peer reports during a query walk — verbatim, inside libp2p, without passing anything of ours. Every node on this network listens on the same port, so a peer's `/ip4/127.0.0.1/tcp/8080` arrives filed under *its* id and resolves to *us*.

The swarm's own self-dial filter does not catch it: that filter drops an address that is exactly a bare listen address of ours, and the `/p2p/<other>` suffix defeats the comparison. So the swarm dialled our own listener alongside the real address, `ConcurrentDial` took whichever handshake finished first, and loopback-to-self usually won — the whole attempt failed `WrongPeerId`.

`unary::Behaviour::handle_pending_outbound_connection` (`unary.rs`) already rejects our own addresses from *its* supplier — that was issue #108. Kad's and identify's suppliers were not covered, and they are the ones holding the raw entry.

### 2. A NATed peer's table entries are its listen addresses

Kad stores what a peer *listens* on. For a NATed peer that is a LAN address on another subnet, or a port mapping that has since closed. The address that reaches it is the **relay circuit in its DHT record**, which kad never holds. So the first dial failing is the normal case for a NATed peer, not an exceptional one — and on `main` the first dial failing was the end of the attempt.

## What changed

**The routed dial is explicit and closed.** `dispatch_routed` no longer forwards on "we have some address". The first request in a burst calls `dial_routed` (`service.rs`), which dials `DialOpts::peer_id(peer).addresses(dial_candidates)` — and *only* that list. An explicit address list is the one filter the swarm honours; leaving the dial open to the behaviours is what supplied our own address in the first place.

**Two address sets, split from one.** `known_addresses` and the new `dial_candidates` share a body, `candidate_addresses` (`service.rs`), and differ in exactly one predicate:

| | routing-table entries | live-connection address | our own addresses | read by |
|---|---|---|---|---|
| `known_addresses` | announceable only | always kept | dropped | `DhtFindPeer` |
| `dial_candidates` | all non-empty | always kept | dropped | `dial_routed` |

The announceability filter answers "worth telling the world about". Dialing is a different question: loopback and RFC1918 are exactly how two nodes on one machine, or one subnet, reach each other, so filtering them out of a *dial* set makes a reachable peer unreachable. The only address dropped from both is one that provably cannot name anyone else — ours. The test is `is_ours`, not "is it loopback".

**Our own addresses are read off the swarm, not accumulated.** `own_addrs()` (`service.rs`) builds an `OwnAddresses` per call from `swarm.listeners()` and `swarm.external_addresses()`. A snapshot cannot drift out of step with the listener set the way a separately-maintained copy can. `OwnAddresses::insert` (`addresses.rs`) is the new entry point for that; `on_swarm_event` now routes through it, so both ways in agree.

**`OwnAddresses` ignores circuit addresses entirely** (`addresses.rs`). Our relay reservation and another peer's circuit through the same relay are the *same string* once `/p2p` hops are stripped — and on this network every NATed node reserves on the same bootstraps. Without this, a peer whose only route is that circuit disappears from `dht_find_peer` and from `unary`'s supplier. A circuit is also never the hazard the type exists for: the swarm appends the destination id and the relay opens a circuit to *that* peer or refuses, so a circuit address filed under X can only ever reach X. Only a direct address can be ours and labelled as someone else's.

**On failure, exactly one DHT walk, then one more dial.** In the `OutgoingConnectionError` arm (`service.rs`), a failed dial that belongs to a routed attempt goes to `start_routed_lookup` instead of failing the request. `lookup_spent` bounds it: the post-lookup dial failing ends the attempt rather than walking the DHT forever.

The attempt state is one `RoutedAttempt` per peer (`dial`, `lookup_spent`, `evicted`), cleared in `fail_all_pending`. `dial` is a `RoutedDial`: `Ours(ConnectionId)` — only that connection's outcome ends the attempt — or `Shared`, when our dial was refused by `PeerCondition::DisconnectedAndNotDialing` because another dial to the peer was already in flight, in which case any outcome for that peer is the one we were waiting on.

**The walk needs the stale entry out of the way first.** `forget_exhausted_entry` (`service.rs`) removes the peer from the routing table before the lookup starts. This is load-bearing, and it is the least obvious part of the change: kad keeps a walk's discoveries in the query's own address book and never in the table, so the only way a walk delivers a peer is by dialing it *itself* — and it will not do that for a peer already in the table. The iterator contacts it at the table's dead address, marks it failed, and never retries it when a bootstrap reports the live one.

Two constraints on that eviction:

- **Evidence-based.** Only addresses the swarm reports having tried count (`DialError::Transport` attempt list, or the single `WrongPeerId` address), and only when they cover the *whole* entry. Kad's own `on_dial_failure` has already removed all but the last failed address by the time this runs.
- **Provisional.** The removed entries go into `RoutedAttempt::evicted`, and `fail_routed` puts them back if the lookup finds nothing better. A peer that is merely restarting is still best reached where it was; without the restore, a transient refusal costs it its table entry until something happens to re-seed it.

Ordering matters here and is deliberate: the pre-existing `WrongPeerId` eviction (`service.rs`) runs **before** `forget_exhausted_entry`, so an address proven to belong to someone else is never put in the stash and restored. Otherwise it comes straight back and every later call repeats the same wrong dial, walk and restore.

**`flush_routed` re-dials rather than assuming the walk's discoveries survive** (`service.rs`). It now branches on `is_connected` alone. If we are connected, forward. If not, this is a completed lookup that left us unconnected, and it tries `dial_routed` once more before failing. Nothing is lost by not reading the walk's in-query address book: kad removes a finished query before reporting `OutboundQueryProgressed`, so those addresses are gone by the time `flush_routed` runs *whichever way the dial is built*. A walk that finds the peer connects to it itself, and `ConnectionEstablished` flushes the parked requests directly — which also promotes the live address into the table, which is what the retry dial reads.

**New error variant: `P2PError::Abandoned`** (`error.rs`). `NetworkHandle::call` mapped *both* failure modes onto `NotInitialized`: a rejected send (the event loop is gone — the node really is down) and a dropped reply (the loop took the command and then lost the responder). On the routed paths the second one happens when the connection carrying an in-flight request dies, taking the handler and the request's reply channel with it. Reporting that as "Network not initialized" describes a dead local node and sends the reader looking in the wrong place. `call` (`handle.rs`) now keeps `NotInitialized` for the send and returns `Abandoned` for the dropped reply.

This is additive: nothing in the workspace matches exhaustively on `P2PError`, and the only consumers of these two variants are inside `kwaai-p2p`. `dht_service::maintenance_loop` stops on any `Err` from `routing_peers()`, so its documented "the service is gone" behaviour is unchanged; the shutdown test at `service_unary.rs:446` still asserts `NotInitialized`, because `fail_all_pending` delivers that as a *value* on the reply channel, not as a drop.

## What deliberately did not change

- **No new config keys, and nothing here is tunable.** The behaviour is unconditional. There is nothing an operator can set that would half-work.
- **p2pd is untouched.** This is `kwaai-p2p`'s native swarm only.
- **The `WrongPeerId` eviction is unchanged** — it is only *ordered* relative to the new one.
- **`known_addresses` keeps its announceability filter.** It is the advertising answer (`dht_find_peer`), and the dial path no longer reads it. Its one behavioural change is that it, too, now drops our own addresses — that is the whole point of the `our_own_address_filed_under_another_peer_is_not_a_route_to_them` test.
- **`collect_routing_peers`** (`service.rs`) still reports raw table entries with no filter, including unusable ones, so the peer table keeps showing "known of but not dialable" as distinct from "not connected yet".
- **`forward_routed`'s `Connect` arm still builds a behaviour-supplied dial** (`service.rs`). It is only reached when the swarm reports us connected, where it returns early — a race-window path, left alone rather than duplicating the candidate logic into it.
- **The retry is bounded at one lookup.** Not a loop, not a backoff, no timer.

## Tests

Five integration tests in `core/crates/kwaai-p2p/tests/service_unary.rs`, all spawning real services on loopback:

| test | what it pins |
|---|---|
| `a_routed_call_falls_back_to_the_dht_when_the_known_address_is_dead` | the reported defect, minus NAT: the caller's only entry for the responder is a closed port; the call must still complete via the walk. Fails on `main`. |
| `our_own_address_filed_under_another_peer_is_not_a_route_to_them` | our own dialable address seeded under a ghost id; `dht_find_peer` must return nothing rather than offering it. |
| `a_routed_call_never_dials_our_own_address_even_when_kad_holds_it` | the same invariant on the *dial* path, which is the one that matters. Our address is seeded **first** so anything dialing the table in order hits it first, and there is no bootstrap in this test — so there is no walk to recover with, and a self-dial makes the call fail outright. |
| `a_second_local_node_on_another_port_is_still_reachable` | the complement — the case a blanket "reject loopback" rule would have broken. |
| `an_evicted_entry_is_restored_when_the_lookup_finds_nothing_better` | the provisional half of `forget_exhausted_entry`: dead address, a bootstrap that has never heard of the target, call fails `DialFailed` — and the peer is still in `routing_peers()` afterwards. |

One unit test in `addresses.rs`, `another_peers_circuit_through_our_relay_is_not_ours`: with a reservation held on relay R, X's circuit through the same R must survive `is_ours` and `reject_self` — and our own reservation must not be recorded at all.

Two helpers are added: `stripped()` shapes an address the way kad stores it, and `free_port()` binds and releases an ephemeral port. `free_port` has a small race window by construction, noted in its doc comment — the alternative is naming a port before it is known.

| command | result |
|---|---|
| `cargo clippy -p kwaai-p2p -p kwaai-hivemind-dht -- -D warnings` | clean |
| `cargo test -p kwaai-p2p -p kwaai-hivemind-dht` | **248 passed, 0 failed**, 5 ignored (baseline 242 + the 6 added here) |

### End-to-end

Verified on the `kwaaiai-env` NAT-test Docker topology (simulated internet on 198.18.0.0/24, RFC1918 LANs behind iptables NAT routers), with a Bob→Eve reachability matrix over `/kwaai/storage/1.0.0` across the NAT axis — public and NATed on both sides. Each pair is run **cold** (a bare RPC by `PeerId`, no connection held) and **warm** (after `p2p peers connect`, which walks the DHT first). Warm passed on `main`; cold did not. With this branch all four pairs pass cold and warm, including NAT→NAT.

A NAT→NAT `vpk bench` through a relay circuit also completes, **but that result depends on a sibling PR**: `relay::Config::default()` caps a circuit at 128 KiB / 2 min, so a multi-hundred-KB batch dies mid-write regardless of this change. That fix is #138 — independent, targets `main` separately, and is not in this diff. This PR fixes the dial path — reaching a peer at all; large transfers over a circuit need the relay-limits PR as well.

## Follow-ups / not done here

Nothing is deferred out of this change — the dial path is complete as it stands, and the bound of one lookup per burst is intended rather than a placeholder. Two observations from the work, neither actionable in this PR:

- The address-supplier problem is structural: any dial left open to the behaviours can be handed our own address under another peer's id, because kad's raw entry is the source and libp2p does not filter it. This branch closes the routed path by supplying an explicit list; `unary` closed its own supplier in #108. Any *future* dial path has to do one or the other, and there is nothing in the type system that says so.
- `forget_exhausted_entry` exists because kad will not re-dial a tabled peer whose address failed. That is upstream behaviour we work around rather than something we can fix here.

### DIFF
--- a/core/crates/kwaai-p2p/src/service.rs
+++ b/core/crates/kwaai-p2p/src/service.rs
@@ -34,7 +34,7 @@ use tracing::{debug, info, trace, warn};
 
 use crate::addresses::{
     dest_peer_id, is_announceable_with, is_circuit, peer_id_from_multiaddr, strip_dest_p2p,
-    strip_p2p,
+    strip_p2p, OwnAddresses,
 };
 use crate::behaviour::{KwaaiBehaviour, KwaaiBehaviourEvent};
 use crate::config::NetworkConfig;
@@ -116,6 +116,8 @@ pub struct NetworkService {
     /// Stream requests parked behind a `RoutedDial` lookup, flushed when the
     /// lookup completes or a connection to the peer establishes first.
     pending_routed: HashMap<PeerId, Vec<RoutedRequest>>,
+    /// What each peer's parked routed requests are waiting on.
+    routed_attempts: HashMap<PeerId, RoutedAttempt>,
     /// Live connections, per peer, keyed by connection so multiple connections
     /// to one peer are tracked independently.
     connections: HashMap<PeerId, HashMap<ConnectionId, Connection>>,
@@ -207,6 +209,27 @@ enum RoutedRequest {
     Connect {
         reply: oneshot::Sender<P2PResult<PeerId>>,
     },
+}
+
+/// How far one burst of parked routed requests has got: dial, then at most one
+/// DHT lookup, then dial again.
+#[derive(Default)]
+struct RoutedAttempt {
+    /// The dial in flight, if any.
+    dial: Option<RoutedDial>,
+    /// Whether the attempt's one lookup has been used.
+    lookup_spent: bool,
+    /// Routing-table entries [`NetworkService::forget_exhausted_entry`] took
+    /// out, put back if the lookup finds nothing better.
+    evicted: Vec<Multiaddr>,
+}
+
+enum RoutedDial {
+    /// A dial we started; only this connection's outcome ends the attempt.
+    Ours(ConnectionId),
+    /// Ours was refused because a dial to the peer was already in flight, so
+    /// any outcome for that peer is the one we are waiting on.
+    Shared,
 }
 
 impl NetworkService {
@@ -356,6 +379,7 @@ impl NetworkService {
             pending_dials: HashMap::new(),
             pending_kad: HashMap::new(),
             pending_routed: HashMap::new(),
+            routed_attempts: HashMap::new(),
             connections: HashMap::new(),
             observed_addrs: HashMap::new(),
             require_global_ips: config.require_global_ips,
@@ -915,10 +939,26 @@ impl NetworkService {
         }
     }
 
-    /// Addresses we already know for `peer`: routing table entries first, then
-    /// any live connection's address.
+    /// Announceable addresses we know for `peer`: routing-table entries first,
+    /// then any live connection's address.
     fn known_addresses(&mut self, peer: &PeerId) -> Vec<Multiaddr> {
         let strict = self.require_global_ips;
+        self.candidate_addresses(peer, |a| is_announceable_with(a, strict))
+    }
+
+    /// Every address we could try for `peer`; unlike [`Self::known_addresses`]
+    /// it keeps loopback and LAN, which is how two nodes on one host reach each other.
+    fn dial_candidates(&mut self, peer: &PeerId) -> Vec<Multiaddr> {
+        self.candidate_addresses(peer, |_| true)
+    }
+
+    /// `table_filter` decides which routing-table entries qualify; a live
+    /// connection's address always does, since we are on it.
+    fn candidate_addresses(
+        &mut self,
+        peer: &PeerId,
+        table_filter: impl Fn(&Multiaddr) -> bool,
+    ) -> Vec<Multiaddr> {
         let mut addrs: Vec<Multiaddr> = Vec::new();
 
         if let Some(bucket) = self.swarm.behaviour_mut().kad.kbucket(*peer) {
@@ -926,23 +966,12 @@ impl NetworkService {
                 if entry.node.key.preimage() == peer {
                     // Skip empty entries: an address with no transport cannot
                     // be dialed, only mistaken for reachability.
-                    //
-                    // Filter here, on the way *out*, because kad has more ways
-                    // in than we control. Identify is filtered at the handler,
-                    // but `kad::Behaviour::discovered` also inserts whatever
-                    // `multiaddrs` a remote peer reports during a query walk,
-                    // verbatim and inside libp2p — so a peer's loopback and LAN
-                    // addresses arrive without passing anything of ours. This
-                    // is the one place every consumer (connect, routed dial,
-                    // the GUI's peer table) reads them back, so one check here
-                    // covers paths that would otherwise need finding one at a
-                    // time.
                     addrs.extend(
                         entry
                             .node
                             .value
                             .iter()
-                            .filter(|a| !a.is_empty() && is_announceable_with(a, strict))
+                            .filter(|a| !a.is_empty() && table_filter(a))
                             .cloned(),
                     );
                 }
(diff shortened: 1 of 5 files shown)


core/crates/kwaai-p2p/src/service.rs
```diff
@@ -972,30 +1008,228 @@ impl NetworkService {
...
+        // id and resolves to *us*. Dialing it connects to our own listener and
+        // fails `WrongPeerId`, which is the one outcome guaranteed to be wrong
+        // no matter whose address it really is.
+        //
+        // The test is `is_ours`, not "is it loopback": a second node on the
+        // same machine, on a different port, is a legitimate peer and its
+        // loopback address is the only way to reach it. Only an address that is
+        // literally one of ours can never name anyone else. Same reasoning as
+        // `unary::Behaviour::handle_pending_outbound_connection` (issue #108),
+        // applied at the other supplier.
```
dazwin: Reduce to 1-2 lines. This is not a complex section of code

dazwin: Done — two lines now (`8a196754`), and the same standard applied across the branch: comments 255 → 72 lines, diff 664 → 500. The three routed-state maps are one `RoutedAttempt` per peer with a `RoutedDial` enum, so the prose that explained how they related is gone with them.

RezaRassool (approved):
Approving. This is the fix for #108's other half, and the analysis is better than the issue I filed — I described the symptom on one supplier; you found that the supplier problem is structural.

## Verified

Merged current `main` (`685bfb7a`) into the branch: clean, **248 tests pass**, clippy clean. No interaction with #139 (`kwaai-cli`) or #140 (`kwaai-rag`).

I traced the parts most likely to be subtly wrong:

**`routed_attempts` lifecycle is complete** — cleared on success in `flush_routed`, on failure in `fail_routed`, wholesale in `fail_all_pending`, with `attempt.dial = None` on `ConnectionEstablished`. No path leaves stale state behind.

**The eviction is genuinely evidence-based.** It fires only when *every* remaining table address appears in the failed set, and the restore in `fail_routed` puts them back. The ordering comment earns its place — *"Must follow the WrongPeerId eviction above so that address is never stashed and restored"* — since restoring an address kad has just proven wrong would be worse than not evicting at all.

**The circuit exemption is correct**, and it is the part I would most have expected to be wrong. Stripped of `/p2p` hops our reservation and X's circuit through the same relay are byte-identical, so without the exemption a NATed peer's only route disappears from both `dht_find_peer` and unary's supplier. The argument that a circuit can only ever reach its appended destination is what makes skipping it safe rather than merely convenient, and `another_peers_circuit_through_our_relay_is_not_ours` pins exactly that collision.

## On the diagnosis

Two things I got wrong in #108 that this corrects:

- I treated it as unary's problem and fixed unary's supplier. Kad and identify hold the same raw entry, and `kad::Behaviour::discovered` inserts remote-reported multiaddrs verbatim inside libp2p — so closing one supplier was never going to be enough.
- I did not connect it to NATed peers at all. That kad stores *listen* addresses, so a NATed peer's table entry is structurally the wrong address and first-dial failure is the normal case rather than an exception, is the observation that turns this from a bug into a design constraint.

The `is_ours` versus "is it loopback" distinction is also the right call, and worth restating because it is easy to get backwards: I nearly used `is_routable_v4` for #108, which would have made two nodes on one machine — and our own two-node local test — unreachable.

## Small suggestions, none blocking

- **`own_addrs()` rebuilds an `OwnAddresses` per call to `candidate_addresses`.** Deliberate, and I agree with the trade — a snapshot that cannot drift beats a cached copy that can. Worth a word in the doc comment saying the allocation is accepted on purpose, so nobody "optimises" it back into a field later.
- **`forget_exhausted_entry` uses `kad.remove_peer` then restores with `add_address`.** That restores the addresses but not necessarily the entry's prior bucket state. Almost certainly irrelevant — the peer is about to be re-dialled either way — but if a restored peer ever behaves oddly in routing, this is where to look.

## The #138 dependency

Thank you for stating it plainly rather than letting the NAT→NAT `vpk bench` result imply more than it shows. Worth carrying into the merge order: this PR makes peers *reachable*, #138 makes a circuit *usable* for real payloads, and either alone leaves a visible gap.

PR #142, opened 2026-08-25
fix(kwaai-cli): restore the Windows build of the gRPC server handle

RezaRassool:
## Summary

Current `main` does not compile on Windows — the whole `kwaainet` binary, not just tests:

```
error[E0063]: missing field `net` in initializer of `GrpcServerHandle`
```

#130 added the `net` slot to `GrpcServerHandle` and populated it in `spawn_on_tcp_port`'s `cfg(unix)` branch, but not the `cfg(not(unix))` one. The field itself is not cfg-gated, so the Windows branch constructs the struct with a field missing.

`net_slot` is already bound at function scope and visible to both branches (it was simply unused on Windows), so filling it in is the entire fix. As a bonus this makes `attach_network` actually meaningful on Windows instead of dropping the service's only route to the swarm.

## Test plan
- [x] `cargo check -p kwaainet` on Windows 11 x86_64 (MSVC): **fails** on `455c7e7`, **clean** with this commit
- [x] Change is inside `cfg(not(unix))`, so Unix/macOS builds are untouched by construction
- [ ] CI: worth confirming why this reached `main` — a Windows `cargo check`/`cargo test` job would have caught it

Found while trying to run tests for a separate updater fix on a Windows node; that work is blocked on this.

### DIFF
--- a/core/crates/kwaai-cli/src/grpc_server.rs
+++ b/core/crates/kwaai-cli/src/grpc_server.rs
@@ -1970,6 +1970,7 @@ pub(crate) fn spawn_on_tcp_port(config: KwaaiNetConfig, tcp_port: u16) -> GrpcSe
         drop(service); // suppress unused warning on non-unix
         GrpcServerHandle {
             shutdown_tcp: Some(shutdown_tcp_tx),
+            net: net_slot,
         }
     }
 }


RezaRassool:
Merged — `main` did not compile on Windows, which blocked metro-win and any Windows contributor.

Recording how this got through, since it is mine.

**The break.** `GrpcServerHandle.net` (`grpc_server.rs:2027`) is declared ungated, so every platform must set it. The `#[cfg(not(unix))]` constructor arm sets only `shutdown_tcp`, hence `E0063`. `net_slot` (`:1931`) is *not* gated — the `#[cfg(unix)]` above it applies to the `shutdown_unix` channel — so this one-line fix is correct and needs nothing further.

Verified on real Windows by the machine that found it: clean build with `--features cuda-windows` in 13m17s, and the binary went on to run a full regression pass.

**Why review missed it.** I reviewed #130, merged current `main` into it, built and tested — on macOS. CI agreed, because **there is no Windows job on pull requests**: every workflow is `runs-on: ubuntu-latest`, and Windows is built only in the release pipeline's `build-cuda-artifacts`.

So a Windows-breaking change can pass review *and* CI, land on `main`, and surface only when someone builds there or a release is cut. That is exactly what happened, and nothing currently prevents a recurrence.

**Worth its own issue:** a `cargo check --target x86_64-pc-windows-msvc` job on PRs touching `core/`. Checking rather than building avoids needing the full MSVC and CUDA toolchain on the runner, and would have caught this in about a minute. I tried that cross-check locally to verify the fix and it dies in `ring`'s build script, which wants a Windows C toolchain — trivial on a Windows runner, impossible from macOS.

PR #156, opened 2026-08-29
feat(grpc): per-instance control surface — home-scoped socket, settable port

dazwin:
## Why

Two KwaaiNetGUI instances on one machine cannot coexist. Every daemon-identity
artifact is machine-global, so the second GUI to start attaches to the first
one's daemon, and a quit from either kills the other's node:

| Artifact | Where | Problem |
|---|---|---|
| gRPC Unix socket | `~/.kwaainet/run/kwaai.sock` via `dirs::home_dir()` | ignores `KWAAINET_HOME`, unlike pid/lock/status |
| gRPC TCP port | hardcoded `127.0.0.1:8093` | no flag, no env, no config key |
| p2pd children | — | `kill_orphaned_p2pd()` SIGKILLs *every* `p2pd` on the machine |

The GUI-side changes are separate; this is the daemon half, and it is useful on
its own to anyone running two nodes under different `KWAAINET_HOME`s.

## What

**The Unix socket follows `KWAAINET_HOME`.** `unix_socket_path()` now resolves
through `config::kwaainet_dir()` rather than `dirs::home_dir()`. This is the
highest-value line in the change: `serve_unix` unlinks any existing socket
before binding (`grpc_server.rs`, "stale socket blocks bind() with EADDRINUSE"),
so two daemons on one path meant **the second one deleted the first one's live
socket**.

**The TCP port is settable.** `--grpc-port` on `start` / `run-node`, else
`KWAAINET_GRPC_PORT`, else the existing `DEFAULT_GRPC_TCP_PORT` (8093).
The env layer is what a supervisor can actually rely on, because
`start --daemon` re-execs `run-node` with no argv (`spawn_daemon_child(&[])`)
but does inherit the environment; the flag is forwarded explicitly so the CLI
form works too. This also fixes a latent bug — `start --daemon --port N`
silently dropped `N` for the same reason. Port 0 binds an ephemeral port.

**Bind failure is fatal only when the port was requested.** The listener is now
bound up front instead of inside the detached serve task, so a clash surfaces
immediately and port 0 resolves to a real number. Defaulting to the well-known
port and losing it stays survivable — the node still serves p2p, and that
leniency predates this change. But a port a supervisor *asked* for and did not
get used to leave a live pid, a written status file, and nothing listening;
that case now exits so the caller can retry.

**The bound port is reported** in `run/kwaainet.grpc`, written once the listener
exists and removed on shutdown — so a supervisor that restarts can find a daemon
it did not spawn. The socket file is removed there too: `serve_unix` cleans up
after itself, but a SIGTERM'd process exits before that task observes the
signal, and a stale socket makes a client's `exists()` check say yes to a daemon
that has gone.

**`kill_orphaned_p2pd()` is scoped** to this instance's `KWAAINET_SOCKET`, which
p2pd carries in `-listen`. With no override there is only one socket on the
machine, so every p2pd is ours and the name-only match is unchanged.

## Compatibility

Nothing changes for a single default-configured daemon: same port, same socket
path, same leniency on a default-port clash. `KWAAINET_HOME` users get a socket
that matches where the rest of their `run/` already lives.

## Tests

Four new tests in `grpc_server.rs`, using the existing `EnvGuard` / `TEST_LOCK`
helpers: the socket and port file follow `KWAAINET_HOME`; flag-then-env-then-
default precedence; an explicit-port clash errors while a default-port clash
does not; and the port file records the resolved port (via port 0) and clears on
shutdown. `cargo test -p kwaainet` — 211 pass; clippy and fmt clean.

Verified end to end on macOS: two daemons under separate `KWAAINET_HOME`s, each
with its own pid, socket and gRPC port, and stopping one leaves the other
running. `~/.kwaainet/run/` untouched throughout.

### DIFF
--- a/core/crates/kwaai-cli/src/cli.rs
+++ b/core/crates/kwaai-cli/src/cli.rs
@@ -167,7 +167,16 @@ A coordinator discovers the chain via DHT and orchestrates inference hop-by-hop.
 
     /// Internal: run the node in the foreground (used by daemon mode)
     #[command(hide = true)]
-    RunNode,
+    RunNode(RunNodeArgs),
+}
+
+/// `run-node` is spawned by `start --daemon`, so anything `start` accepts and
+/// the node needs has to be forwarded here explicitly.
+#[derive(Args, Debug)]
+pub struct RunNodeArgs {
+    /// TCP port for the gRPC control surface (0 = ephemeral)
+    #[arg(long)]
+    pub grpc_port: Option<u16>,
 }
 
 // ---------------------------------------------------------------------------
@@ -187,6 +196,11 @@ pub struct StartArgs {
     /// TCP port for P2P connections
     #[arg(long)]
     pub port: Option<u16>,
+
+    /// TCP port for the gRPC control surface (0 = ephemeral). Forwarded to
+    /// the daemon child; also settable via KWAAINET_GRPC_PORT.
+    #[arg(long)]
+    pub grpc_port: Option<u16>,
 
     /// Disable GPU acceleration (use CPU only)
     #[arg(long)]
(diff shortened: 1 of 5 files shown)


dazwin:
The reason I added this is so that I can run both a production release concurrently with a debug/vscode release

PR #163, opened 2026-08-31
fix(tests): Windows has never actually run the test suite — make it

dazwin:
## What

`Test (windows-latest)` has failed on every push to main since #84. Three layers, because `cargo test` stops at the first failing target — each fix exposed the next:

1. **`kwaai-p2p-daemon/tests/control_server.rs`** — all 19 tests bound `/unix/<tmpdir>/kwaai.sock`, which `ControlServer::bind` refuses on Windows. The suite is about the control protocol, not its transport, so Windows now runs the same 19 tests over loopback TCP (`/ip4/127.0.0.1/tcp/0`, resolved via `local_addr()`) instead of skipping them. Everything past this target had **never been compiled-and-run on Windows** — which is how layer 2 stayed invisible.

2. **`grpc_server::tests` shutdown probes** — these assert that dropping the handle closes the TCP listener. They did it by connecting to the port and deciding, from how long the connect stayed silent, whether anything was still there. On Windows that question has no cheap answer: a refused loopback connect is reported only after an in-stack SYN retry. **The probe no longer asks it.** It tries to `bind` the port instead, which is immediate and unambiguous on every platform. Detail below, because getting here took three wrong answers.

3. **`node_native` fixture** — set `KWAAINET_SOCKET` to a bare filesystem path, which `control_socket_addr()` wraps in `/unix/…` only on POSIX; on Windows the variable must already be a multiaddr. The fixture now uses `/ip4/127.0.0.1/tcp/0` there.

No production code changes — test harnesses and one test-only probe.

## Layer 2, and why it took four commits

Every version before the last tuned the same number: how long to wait before calling a silent port closed. Each was wrong, and the two failure modes are not symmetric — a test that *cannot pass* is loud, a test that *cannot fail* is silent.

| | probe | verdict |
|---|---|---|
| original | 250 ms ceiling, timeout ⇒ still listening | fails on Windows always: Winsock never answers that fast |
| `8514a0bf` | timeout ⇒ closed | **unfalsifiable** — a starved runtime reads a live listener as gone, which is the exact regression the test exists to catch |
| `0cfb47db` | 1500 ms ceiling, timeout ⇒ still listening | unfalsifiable the other way: below Winsock's ~2.04 s floor, so no probe could ever observe a close |
| `d97b8dec` | 10 s ceiling | correct, but still a threshold guessing at a machine |
| `e307da64` | **`bind` the port** | no threshold at all |

The floor is not a constant. Measured on two boxes, both stock (`MaxSynRetransmissions = 4`, no registry overrides):

| box | mean | worst |
|---|---|---|
| Win 11 Pro Workstations, 64 cores | ~2.04 s | 2.12 s |
| Win 10 22H2, i7-3667U 2c/4t | ~2.30 s | 2.43 s |

Same retry count, different per-retry latency — it scales with the hardware, in the direction CI runners sit. Any ceiling is a guess about a machine.

`bind` needs no guess. Modelling the server's own listener (`std::net::TcpListener::bind`, no socket options — std sets `SO_REUSEADDR` on Unix only), measured on Windows 10 22H2:

- against a live listener, fails in **under a millisecond** with `WSAEADDRINUSE`; on Unix `SO_REUSEADDR` permits `TIME_WAIT` but never an active listener;
- after a connection closed **server-side**, stranding `TIME_WAIT` on that very port, succeeds in **~0.3 ms** with the entry still present in `netstat`.

The second was the failure mode worth fearing — it would report a correctly closed listener as still up. It does not occur, and it was tested in the close order that produces it: closing the client first puts `TIME_WAIT` on the client's ephemeral port and exercises nothing.

One deliberate trade: the probe asks "is this port bindable", not "did *our* listener close", so something else taking the port would report a leak that isn't there. That is a spurious **failure**, never a silent pass — a leaked listener still holds the port. For a test guarding a shutdown regression that is the right direction, and the opposite of what the connect-probe had.

`LISTENER_CLOSE_BUDGET` survives at 10 s but is now an ordinary liveness budget: it bounds how long a genuine leak takes to be reported, and nothing is inferred from it.

## Evidence

Verified on a temporary branch-CI trigger (dropped before this PR; the Windows *test* job only runs on pushes to main, so a fix could otherwise never be verified pre-merge):

- Before: 19 failures in `control_server`, run aborted there — [33414217706](https://github.com/Kwaai-AI-Lab/KwaaiNet/actions/runs/33414217706) (main, same failures)
- Layer 2 surfaced: `control_server` 19/19, then 3 failures in the `kwaainet` bin — [33417416472](https://github.com/Kwaai-AI-Lab/KwaaiNet/actions/runs/33417416472)
- Fully green including `Test (windows-latest)` — [33419106106](https://github.com/Kwaai-AI-Lab/KwaaiNet/actions/runs/33419106106)

**Falsifiability is checked, not argued** — a test that cannot fail is the defect this branch twice reintroduced. `mem::forget` on the handle fails with `127.0.0.1:53796 still not bindable 10s after dropping the handle`. Unmodified: 219 tests pass, `clippy --all-targets -D warnings` and `fmt` clean.

Windows hardware runs on `e307da64` are in progress on a second box; the socket-level measurements above are already from real Windows, not inference from macOS.

## Follow-up

Making the Windows *test* job part of the PR gate. This PR's own job list checks only ubuntu plus the Windows *type* check, which is exactly why this went unseen for six weeks — and "Windows is green" on a PR has been cited as evidence elsewhere when it meant only that the code type-checks. Worth doing alongside #144 the moment this lands.

### DIFF
--- a/core/crates/kwaai-cli/src/grpc_server.rs
+++ b/core/crates/kwaai-cli/src/grpc_server.rs
@@ -2407,34 +2407,55 @@ mod tests {
             .is_ok()
     }
 
-    /// True iff a fresh TCP connect to `port` is refused quickly (used to
-    /// assert the listener is gone after shutdown).
-    async fn tcp_refused(port: u16) -> bool {
-        // ConnectionRefused is the happy-path answer; any other Err (e.g.
-        // network unreachable) we also treat as "not accepting".
-        //
-        // Silence must NOT count as "not accepting". Winsock retries a RST'd
-        // connect in-stack and reports WSAECONNREFUSED only after ~1s, so the
-        // budget has to clear that — but treating a timeout as "closed" would
-        // make the assertion unfalsifiable from the other direction. These are
-        // current-thread `#[tokio::test]`s and `timeout` measures wall clock,
-        // including time the inner future is not polled; the probe runs while
-        // the server task, its connection drain and the client channel are all
-        // still scheduled on this one thread. A starved runtime would then
-        // report a live listener as closed, which is precisely the
-        // serve_with_shutdown regression these tests exist to catch.
-        //
-        // So: wait long enough for Winsock to answer, and keep silence
-        // meaning "still up".
+    /// What one connect probe learned about a port.
+    #[derive(Clone, Copy, PartialEq, Eq, Debug)]
+    enum Probe {
+        Accepting,
+        /// Refused, or unreachable — either way nothing is serving.
+        Refused,
+        /// No answer inside the ceiling; tells us nothing either way.
+        Silent,
+    }
+
+    /// Winsock answers a refused loopback connect only after a fixed in-stack
+    /// retry, measured at ~2.04s on Windows and the same for a port that never
+    /// had a listener. Probing below that reads every closed port as `Silent`.
+    const REFUSAL_PROBE_CEILING: Duration = Duration::from_secs(4);
+
+    /// Three probes at the ceiling, so one starved answer is not the verdict.
+    /// Kept next to the ceiling it derives from: as independent literals the
+    /// two drifted apart twice, each time breaking the assertion below.
+    const LISTENER_CLOSE_BUDGET: Duration = Duration::from_secs(12);
+
+    async fn probe_port(port: u16) -> Probe {
         match tokio::time::timeout(
-            Duration::from_millis(1500),
+            REFUSAL_PROBE_CEILING,
             tokio::net::TcpStream::connect(("127.0.0.1", port)),
         )
         .await
         {
-            Ok(Ok(_)) => false, // still accepting
-            Ok(Err(_)) => true, // refused / unreachable
-            Err(_) => false,    // timed out — something is listening but not answering yet
+            Ok(Ok(_)) => Probe::Accepting,
+            Ok(Err(_)) => Probe::Refused,
+            Err(_) => Probe::Silent,
+        }
+    }
+
+    /// Poll until `port` refuses. `Silent` never counts as closed: `timeout`
+    /// measures wall clock, so on these current-thread tests a probe starved by
+    /// the server task it is shutting down would report a live listener as gone
+    /// — the `serve_with_shutdown` regression these tests exist to catch.
+    /// Returns the last outcome seen, so a failure says which one it was.
+    async fn wait_for_close(port: u16) -> Result<(), Probe> {
+        let deadline = Instant::now() + LISTENER_CLOSE_BUDGET;
+        loop {
+            let seen = probe_port(port).await;
+            if seen == Probe::Refused {
+                return Ok(());
+            }
+            if Instant::now() >= deadline {
+                return Err(seen);
+            }
+            tokio::time::sleep(Duration::from_millis(25)).await;
         }
     }
 
@@ -2497,11 +2518,11 @@ mod tests {
         // tonic's serve_with_shutdown returns -> listener is closed.
         drop(handle);
 
-        let down = wait_for(Duration::from_secs(10), || tcp_refused(port)).await;
-        assert!(
-            down,
-            "TCP listener on 127.0.0.1:{port} did not close within 10s of dropping the handle"
-        );
+        wait_for_close(port).await.unwrap_or_else(|last| {
+            panic!(
+                "TCP listener on 127.0.0.1:{port} still {last:?} {LISTENER_CLOSE_BUDGET:?} after dropping the handle"
+            )
+        });
 
         #[cfg(unix)]
         {
@@ -2581,8 +2602,9 @@ mod tests {
         drop(handle);
         // Wait for the listener to actually go away before the next test
         // tries to bind the same port.
-        let down = wait_for(Duration::from_secs(10), || tcp_refused(port)).await;
-        assert!(down, "TCP listener did not close after handle drop");
+        wait_for_close(port).await.unwrap_or_else(|last| {
+            panic!("TCP listener still {last:?} {LISTENER_CLOSE_BUDGET:?} after handle drop")
+        });
     }
 
     #[test]
--- a/core/crates/kwaai-cli/src/grpc_server.rs
+++ b/core/crates/kwaai-cli/src/grpc_server.rs
@@ -2413,6 +2413,13 @@ mod tests {
         // ConnectionRefused is the happy-path answer; any other Err (e.g.
         // network unreachable) we also treat as "not accepting". We bound
         // the dial with a short timeout so a slow stack can't lie to us.
+        //
+        // Silence counts as "not accepting" too: Winsock retries a RST'd
+        // connect in-stack and reports WSAECONNREFUSED only after ~1s, so on
+        // Windows a closed port looks like a timeout from inside this window.
+        // The reverse cannot mislead us — a live loopback listener completes
+        // the handshake in-kernel (backlog) long before 250ms, without the
+        // server ever calling accept.
         match tokio::time::timeout(
             Duration::from_millis(250),
             tokio::net::TcpStream::connect(("127.0.0.1", port)),
@@ -2420,8 +2427,7 @@ mod tests {
         .await
         {
             Ok(Ok(_)) => false, // still accepting
-            Ok(Err(_)) => true, // refused / unreachable
-            Err(_) => false,    // timed out = something is listening but not answering yet
+            _ => true,          // refused, unreachable, or silence
         }
     }
 
--- a/core/crates/kwaai-cli/src/grpc_server.rs
+++ b/core/crates/kwaai-cli/src/grpc_server.rs
@@ -2418,14 +2418,31 @@ mod tests {
     }
 
     /// Winsock answers a refused loopback connect only after a fixed in-stack
-    /// retry, measured at ~2.04s on Windows and the same for a port that never
-    /// had a listener. Probing below that reads every closed port as `Silent`.
-    const REFUSAL_PROBE_CEILING: Duration = Duration::from_secs(4);
+    /// retry — identical for a port that never had a listener, so it is the
+    /// stack, not the server. Probing below it reads every closed port as
+    /// `Silent`, which is a test that cannot pass on Windows.
+    ///
+    /// Measured, `MaxSynRetransmissions = 4` and no registry overrides on both:
+    ///
+    /// | box | | worst |
+    /// |---|---|---|
+    /// | Win 11 Pro Workstations, 64 cores | ~2.04s | 2.12s |
+    /// | Win 10 22H2, i7-3667U 2c/4t | ~2.30s | 2.43s |
+    ///
+    /// Same retry count, different per-retry latency: the floor tracks the
+    /// hardware, so it is not a constant to sit just above. This is 4x the
+    /// slower reading rather than a tight margin on either, because the two
+    /// directions do not cost the same. Too low and every probe expires as
+    /// silence and the assertion can never pass — shipped twice, at 250ms and
+    /// 1500ms. Too high costs only how long a *genuinely* failing test takes
+    /// to give up: the probe returns the moment Winsock answers, so a healthy
+    /// shutdown still resolves in one refusal latency whatever this says.
+    const REFUSAL_PROBE_CEILING: Duration = Duration::from_secs(10);
 
     /// Three probes at the ceiling, so one starved answer is not the verdict.
     /// Kept next to the ceiling it derives from: as independent literals the
     /// two drifted apart twice, each time breaking the assertion below.
-    const LISTENER_CLOSE_BUDGET: Duration = Duration::from_secs(12);
+    const LISTENER_CLOSE_BUDGET: Duration = Duration::from_secs(30);
 
     async fn probe_port(port: u16) -> Probe {
         match tokio::time::timeout(
--- a/core/crates/kwaai-cli/src/grpc_server.rs
+++ b/core/crates/kwaai-cli/src/grpc_server.rs
@@ -2407,70 +2407,59 @@ mod tests {
             .is_ok()
     }
 
-    /// What one connect probe learned about a port.
-    #[derive(Clone, Copy, PartialEq, Eq, Debug)]
-    enum Probe {
-        Accepting,
-        /// Refused, or unreachable — either way nothing is serving.
-        Refused,
-        /// No answer inside the ceiling; tells us nothing either way.
-        Silent,
+    /// Whether a fresh listener can be bound to `port` — which is the
+    /// question these tests actually ask, and unlike "is the port refusing
+    /// connections?" it has an immediate, unambiguous answer on every
+    /// platform.
+    ///
+    /// The connect-probe this replaces had to tell "closed" from "slow to
+    /// answer" by waiting, and on Windows a refused loopback connect is only
+    /// reported after an in-stack SYN retry — ~2.04s on one box, ~2.30s on a
+    /// slower one, scaling with the hardware. Every version of that test was a
+    /// guess at a threshold, and two of them shipped wrong in the direction
+    /// that cannot pass at all.
+    ///
+    /// `bind` needs no threshold. Measured on Windows 10 22H2, and matching
+    /// how the server binds (`std::net::TcpListener::bind`, no socket options
+    /// set — std sets `SO_REUSEADDR` on Unix only):
+    ///
+    /// - against a live listener it fails in under a millisecond with
+    ///   `WSAEADDRINUSE`; on Unix `SO_REUSEADDR` permits `TIME_WAIT`, never an
+    ///   active listener;
+    /// - after a connection closed *server-side*, leaving `TIME_WAIT` on this
+    ///   very port, it succeeds in ~0.3ms with the entry still in `netstat`.
+    ///   That was the failure mode worth fearing, and it does not occur.
+    ///
+    /// The one way this differs from the old probe: it answers "is the port
+    /// bindable", not "did *our* listener close". Something else taking the
+    /// port in between would report a leak that isn't there. That is a
+    /// spurious *failure*, never a silent pass — a leaked listener still holds
+    /// the port and still fails the assertion — which is the direction to err
+    /// in for a test guarding a shutdown regression.
+    fn port_is_free(port: u16) -> bool {
+        // Bound and dropped inside the call: holding it would be indis-
+        // tinguishable from the leak we are testing for.
+        std::net::TcpListener::bind(("127.0.0.1", port)).is_ok()
     }
 
-    /// Winsock answers a refused loopback connect only after a fixed in-stack
-    /// retry — identical for a port that never had a listener, so it is the
-    /// stack, not the server. Probing below it reads every closed port as
-    /// `Silent`, which is a test that cannot pass on Windows.
+    /// How long to wait for an asynchronous shutdown to release the port.
     ///
-    /// Measured, `MaxSynRetransmissions = 4` and no registry overrides on both:
-    ///
-    /// | box | | worst |
-    /// |---|---|---|
-    /// | Win 11 Pro Workstations, 64 cores | ~2.04s | 2.12s |
-    /// | Win 10 22H2, i7-3667U 2c/4t | ~2.30s | 2.43s |
-    ///
-    /// Same retry count, different per-retry latency: the floor tracks the
-    /// hardware, so it is not a constant to sit just above. This is 4x the
-    /// slower reading rather than a tight margin on either, because the two
-    /// directions do not cost the same. Too low and every probe expires as
-    /// silence and the assertion can never pass — shipped twice, at 250ms and
-    /// 1500ms. Too high costs only how long a *genuinely* failing test takes
-    /// to give up: the probe returns the moment Winsock answers, so a healthy
-    /// shutdown still resolves in one refusal latency whatever this says.
-    const REFUSAL_PROBE_CEILING: Duration = Duration::from_secs(10);
+    /// An ordinary liveness budget, not a threshold anything is inferred
+    /// from: each probe is decisive on its own, so this only bounds how long
+    /// a genuine leak takes to be reported. `drop` -> oneshot ->
+    /// `serve_with_incoming_shutdown` returns -> listener closed is normally
+    /// well under a millisecond.
+    const LISTENER_CLOSE_BUDGET: Duration = Duration::from_secs(10);
 
-    /// Three probes at the ceiling, so one starved answer is not the verdict.
-    /// Kept next to the ceiling it derives from: as independent literals the
-    /// two drifted apart twice, each time breaking the assertion below.
-    const LISTENER_CLOSE_BUDGET: Duration = Duration::from_secs(30);
-
-    async fn probe_port(port: u16) -> Probe {
-        match tokio::time::timeout(
-            REFUSAL_PROBE_CEILING,
-            tokio::net::TcpStream::connect(("127.0.0.1", port)),
-        )
-        .await
-        {
-            Ok(Ok(_)) => Probe::Accepting,
-            Ok(Err(_)) => Probe::Refused,
-            Err(_) => Probe::Silent,
-        }
-    }
-
-    /// Poll until `port` refuses. `Silent` never counts as closed: `timeout`
-    /// measures wall clock, so on these current-thread tests a probe starved by
-    /// the server task it is shutting down would report a live listener as gone
-    /// — the `serve_with_shutdown` regression these tests exist to catch.
-    /// Returns the last outcome seen, so a failure says which one it was.
-    async fn wait_for_close(port: u16) -> Result<(), Probe> {
+    /// Poll until `port` can be bound again, i.e. the listener is gone.
+    async fn wait_for_close(port: u16) -> Result<(), ()> {
         let deadline = Instant::now() + LISTENER_CLOSE_BUDGET;
         loop {
-            let seen = probe_port(port).await;
-            if seen == Probe::Refused {
+            if port_is_free(port) {
                 return Ok(());
             }
             if Instant::now() >= deadline {
-                return Err(seen);
+                return Err(());
             }
             tokio::time::sleep(Duration::from_millis(25)).await;
         }
@@ -2535,9 +2524,9 @@ mod tests {
         // tonic's serve_with_shutdown returns -> listener is closed.
         drop(handle);
 
-        wait_for_close(port).await.unwrap_or_else(|last| {
+        wait_for_close(port).await.unwrap_or_else(|()| {
             panic!(
-                "TCP listener on 127.0.0.1:{port} still {last:?} {LISTENER_CLOSE_BUDGET:?} after dropping the handle"
+                "127.0.0.1:{port} still not bindable {LISTENER_CLOSE_BUDGET:?} after dropping the handle"
             )
         });
 
@@ -2619,8 +2608,8 @@ mod tests {
         drop(handle);
         // Wait for the listener to actually go away before the next test
         // tries to bind the same port.
-        wait_for_close(port).await.unwrap_or_else(|last| {
-            panic!("TCP listener still {last:?} {LISTENER_CLOSE_BUDGET:?} after handle drop")
+        wait_for_close(port).await.unwrap_or_else(|()| {
+            panic!("port still not bindable {LISTENER_CLOSE_BUDGET:?} after handle drop")
         });
     }
 
--- a/core/crates/kwaai-cli/src/grpc_server.rs
+++ b/core/crates/kwaai-cli/src/grpc_server.rs
@@ -2411,23 +2411,30 @@ mod tests {
     /// assert the listener is gone after shutdown).
     async fn tcp_refused(port: u16) -> bool {
         // ConnectionRefused is the happy-path answer; any other Err (e.g.
-        // network unreachable) we also treat as "not accepting". We bound
-        // the dial with a short timeout so a slow stack can't lie to us.
+        // network unreachable) we also treat as "not accepting".
         //
-        // Silence counts as "not accepting" too: Winsock retries a RST'd
-        // connect in-stack and reports WSAECONNREFUSED only after ~1s, so on
-        // Windows a closed port looks like a timeout from inside this window.
-        // The reverse cannot mislead us — a live loopback listener completes
-        // the handshake in-kernel (backlog) long before 250ms, without the
-        // server ever calling accept.
+        // Silence must NOT count as "not accepting". Winsock retries a RST'd
+        // connect in-stack and reports WSAECONNREFUSED only after ~1s, so the
+        // budget has to clear that — but treating a timeout as "closed" would
+        // make the assertion unfalsifiable from the other direction. These are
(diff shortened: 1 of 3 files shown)


RezaRassool (commented):
## Review — Windows test suite

Read all three hunks plus the surrounding code (`ControlServer::bind`/`local_addr`, `P2PClient::connect_stream`, `control_socket_addr`, `wait_for`, and every caller of `tcp_refused`).

The diagnosis is good and the layered write-up made it easy to follow. Two of the three changes are clean; one weakens the assertion it is meant to preserve.

### Verified sound

- **`control_server.rs`** — `ControlServer::local_addr()` already exists on the merge base and returns `/unix/<path>` verbatim on the Unix branch, so the Unix `socket` string is byte-identical to before. `local_addr()` is correctly called before `server.run()` consumes `self`, `TestNode` binds once so port 0 never needs re-resolving, and `StreamEndpoint` was already loopback TCP.
- **`node_native.rs`** — matches `control_socket_addr()`'s cfg split exactly. The test never dials the control socket (it calls `node.announce()` directly), so the unresolved port 0 is harmless.

### Finding — `grpc_server.rs:2430`

Folding the timeout arm into `_ => true` turns `tcp_refused` into `!tcp_accepting` with a 250 ms cap, which removes the thing the assertion was checking.

The justification holds for the kernel but not for the *observation*. Both callers are plain `#[tokio::test]`, i.e. a current-thread runtime, and `tokio::time::timeout` measures wall time including time the inner future is not polled. In `chat_returns_invalid_argument_on_empty_inbound_stream` the probe runs immediately after `drop(handle)` while the tonic server task, its connection drain and the client channel are all still scheduled on that same thread. If the runtime is starved for >250 ms — reachable on a loaded Actions Windows runner — the first probe returns `true`, `wait_for` returns immediately, and `assert!(down, …)` passes while the listener is still open.

That is precisely the `serve_with_shutdown` regression the test exists to catch, and it would now be a silent green. The same arm also swallows genuinely unexpected errors (`WSAENOBUFS`, `EMFILE`), which would make both shutdown assertions trivially pass.

Since the diagnosis here is that Winsock reports `WSAECONNREFUSED` after ~1 s, the fix that keeps the assertion meaningful is to raise the per-dial budget above that (~2 s, with the outer `wait_for` budget raised to match) and keep `Err(_) => false` — accommodate the timing rather than making silence mean success.

### Notes, not findings

- On Windows the suite now runs the control protocol over TCP, and neither `P2PClient::connect_stream` nor the server's `Listener::Tcp` accept path sets `TCP_NODELAY` (only the stream-handler relay does). Probably harmless — every control request is a single `write_all` + `flush` in strict ping-pong — but it is the one behavioural difference between the two transports the suite now spans.
- `node_native.rs:986` — `remove_var("KWAAINET_SOCKET")` is on the success path only, so a panic mid-test leaks the var. Pre-existing; the Unix branch had the same shape.

### Agreed on the follow-up

Making the Windows *test* job part of the PR gate is worth doing as soon as this lands. Until then "Windows is green" on a PR means only that it type-checks — which is worth knowing, because it has been cited as evidence on other PRs, including mine.

RezaRassool:
Pushed `8514a0bf` to this branch addressing the `tcp_refused` finding — hope that's alright rather than leaving it as a suggestion; this one is wanted urgently.

Kept your diagnosis, changed the remedy. Winsock genuinely does read a closed port as a timeout inside 250 ms, so the window had to move — but widening it (1500 ms, past the ~1s `WSAECONNREFUSED` report) preserves the assertion, whereas reinterpreting silence as "closed" removes it.

The outer `wait_for` budget also went 2s → 10s, because at a 1500 ms ceiling a 2s budget only fits a single dial, which would have swapped one single-sample flake for another. Nothing slows down in the normal case — a closed listener still answers on the first probe and the suite runs in 0.04s.

219 tests pass, clippy and fmt clean. Revert or rework it freely if you'd rather solve it differently; the finding is what matters, not my patch.

Separately, strong agreement on your closing suggestion: the Windows *test* job belongs in the PR gate. Until it is, "Windows is green" on a PR means only that it type-checks — I leaned on exactly that claim as evidence on #165 earlier today, which was wrong of me and is a good argument for your follow-up.

dazwin:
Reviewed 8514a0bf ("keep silence meaning 'still listening' in tcp_refused") — the fix is right, with one inaccuracy in the commit message worth flagging.

**Why the change is correct.** The parent commit's `_ => true` arm made the shutdown assertion unfalsifiable: both callers are current-thread `#[tokio::test]`s, and `tokio::time::timeout` measures wall clock even while the probe isn't polled. On a starved runner the probe could time out while the listener was still up, `wait_for` would return immediately, and the test would pass on exactly the `serve_with_shutdown` regression it exists to catch. Returning timeouts to "still listening" and raising the dial ceiling to 1500 ms (past Winsock's ~1 s refused-report delay) restores what the assertion checks. Verified the mechanics:

- Both `tcp_refused` call sites got the 2 s → 10 s budget bump, and no other instance of the timeout-means-closed pattern exists elsewhere on the branch (the `control_server.rs` changes from the earlier commits don't use it).
- The budget math holds: 1500 ms ceiling + 25 ms `wait_for` sleep fits ~6 probes in 10 s, so one slow Winsock report can't fail the test — the old 2 s budget would have allowed exactly one probe at this ceiling, a single-sample flake. Happy path is unchanged (connect fails immediately); the 10 s only bites on genuine failure.
- A backlog-full or starved-runtime probe now degrades toward a *failing* assertion rather than a silently passing one — the right direction for these tests.

**Commit-message inaccuracy.** It says the old timeout arm "also swallowed WSAENOBUFS and EMFILE." It didn't, and the fix doesn't change their handling: fd/buffer exhaustion surfaces as `Ok(Err(...))` from `connect`, not as a timeout, and the new code still maps every `Ok(Err(_))` to "closed". An EMFILE'd runner can still pass this assertion falsely. Optional strengthening, in the spirit of this commit: `Ok(Err(e)) => e.kind() == ErrorKind::ConnectionRefused` — on loopback, refused is the only honest "closed" signal — letting anything else loop to a clear 10 s failure. The doc comment openly declares the any-error-counts choice, so this is a conscious trade-off; only the message overclaims.

Two trivial nits: the doc comment still says "refused *quickly*", which is now up to 1500 ms; and `wait_for` checks the deadline after the probe, so a probe started just before it can stretch the real budget to ~11.5 s — harmless, just don't read the assert message's "10s" as exact.

CI corroborates the message's claim: Windows check, clippy, and fmt green on the PR head at review time (Ubuntu test still pending).

RezaRassool:
## Windows hardware results — metro-win (Win 11 Pro Workstations, x86_64-pc-windows-msvc)

Ran `8514a0bf` on real Windows. Short version: **the two shutdown tests fail, but not for the reason we expected — the listener closes fine. The 1500 ms budget in `tcp_refused` is below Winsock's actual refusal latency, so the probe can never observe the close.**

### Results

`cargo test --workspace --no-fail-fast` → **222 passed; 2 failed**

| Target | Result |
|---|---|
| `kwaai-p2p-daemon --test control_server` | ✅ 19/19, 2.24s |
| `kwaainet --bin kwaainet node_native` | ✅ 1/1, 0.13s |
| `kwaainet --bin kwaainet grpc_server` | ❌ 11 passed, 2 failed, 21.65s |

```
thread 'grpc_server::tests::server_binds_and_shuts_down_cleanly' panicked at
crates\kwaai-cli\src\grpc_server.rs:2501:9:
TCP listener on 127.0.0.1:58771 did not close within 10s of dropping the handle

thread 'grpc_server::tests::chat_returns_invalid_argument_on_empty_inbound_stream' panicked at
crates\kwaai-cli\src\grpc_server.rs:2585:9:
TCP listener did not close after handle drop
```

### Measured Winsock refusal latency

Direct measurement on this box, loopback, via .NET `TcpClient.ConnectAsync`:

```
A. never-used port (baseline):   REFUSED  2124 ms / 2039 ms / 2040 ms
B. bind -> close -> connect:     REFUSED  2028 ms / 2044 ms / 2029 ms
```

**~2.04 s**, and identical for a port that never had a listener — so this is pure Winsock, independent of the server.

The budget is 1500 ms. `tcp_refused` maps timeout to `Err(_) => false` ("still up"), so every probe expires at 1500 ms before Winsock answers at ~2040 ms. It **cannot return `true` on Windows regardless of listener state**; the 10 s loop is just repeated guaranteed-false probes.

### Confirmation

Changed only `1500` -> `3000`, nothing else:

```
test grpc_server::tests::server_binds_and_shuts_down_cleanly ... ok
test grpc_server::tests::chat_returns_invalid_argument_on_empty_inbound_stream ... ok
test result: ok. 13 passed; 0 failed ... finished in 4.20s
```

13/13, and the suite drops 21.65s -> 4.20s. Reverted locally; nothing pushed.

### What this means

The mechanism in the comment is right, the magnitude isn't. It says *"reports WSAECONNREFUSED only after ~1s, so the budget has to clear that"* — the real figure here is **~2.04 s**, so 1500 ms doesn't clear it.

The comment also guards against making the assertion *"unfalsifiable from the other direction"* if timeout counted as closed. Guarding that way produced the mirror-image failure: with the budget under Winsock's floor the assertion is unfalsifiable in **this** direction — it always fails, so it can't distinguish a leaking listener from a healthy one either.

So `did not close within 10s` firing on Windows is **not** evidence the listener stays open. Concretely, the patch doesn't only convert false greens into failures — on Windows it converts *every* green into a failure. 250 ms was too short; 1500 ms is too short as well, just less obviously.

Suggested fix: budget comfortably above the measured floor (3000 ms verified green here). Since ~2.04 s looks like a fixed Winsock retry, a 2x margin seems prudent — a loaded CI runner will be slower — and it's probably worth naming the constant with the measurement recorded next to it, given the same trap has now bitten twice.

### Two unrelated notes

- The suite writes 47 lines into the tracked file `tests/kwaai-network/results/metrics.jsonl` (tagged `"git_rev":"8514a0bf"`), so running the tests dirties the working tree.
- Confirmed the `GrpcServerHandle { net: net_slot }` fix is on this branch, so it compiles on Windows. Worth noting `main` was uncompilable on Windows for ~6 days before that landed (#142) — a per-PR Windows `cargo check` would have caught it (#144).

Happy to re-run anything on this hardware.

dazwin:
You're right, and the measurement settles it — thanks for taking the time to instrument it. Fixed in 0cfb47db.

My comment asserted "~1 s" as the Winsock floor. That was an estimate carried over from the commit message, not a measurement, and 1500 ms didn't clear the real ~2.04 s. The consequence is exactly as you describe: every probe expired as silence, `tcp_refused` could never return `true` on Windows, and the 10 s loop was repeated guaranteed-false probes. I traded one unfalsifiable direction for the other, which is worse than the bug I was fixing — the 250 ms version at least passed when the listener really closed.

What's on the branch now:

- **Ceiling 4 s**, 2× your measured floor as you suggested, since your figure is a fixed timer and a loaded runner adds scheduling delay on top of it. It costs nothing on the happy path — the probe returns the moment Winsock answers, so this only bounds how long we wait before giving up on one probe.
- **Both numbers are named constants**, with your measurement recorded on the ceiling and the budget stated as three probes at that ceiling. They were edited independently twice and drifted apart both times, which is the actual recurring defect.
- **The probe returns a tri-state** (`Accepting` / `Refused` / `Silent`) instead of a bool. `did not close within 10s` is the message that sent you after a listener leak that was never happening; a failure now reports what was observed — `still Accepting` or `still Silent`.

On verification, I took your point that green isn't enough — the failure mode here was a test that couldn't fail. So I checked falsifiability directly: leaking the handle with `mem::forget` fails with `TCP listener on 127.0.0.1:60174 still Accepting 12s after dropping the handle`, at the full budget. Unmodified, 219 tests pass, clippy `-D warnings` and fmt clean, suite in 0.06 s on macOS.

If you have a moment on that hardware, the run worth having is `cargo test -p kwaainet --bin kwaainet grpc_server` — I'd expect 13/13 with the two tests each taking ~2 s, so a suite around 4 s, matching your 3000 ms experiment. The number I can't check from here is whether 4 s holds under Actions load rather than on an idle box; if you see `still Silent` in a failure, that's the ceiling being too low again and the message will now say so plainly rather than blaming the listener.

On your two unrelated notes:

- **metrics.jsonl** — confirmed tracked, last written by #96, and this branch doesn't touch `tests/` at all, so it's pre-existing on `main` rather than something these commits introduced. Left it alone rather than fold an unrelated change into a Windows test fix; worth its own issue, since "running the tests dirties the working tree" will bite anyone who runs the suite before committing.
- **Windows `cargo check` per PR (#144)** — agreed, and this PR is the argument for it. Three commits in a row here were shaped by guessing at Windows behaviour from macOS; the first one that had actual Windows numbers attached is the one that's correct.

RezaRassool:
## Integration run on real Windows — `test/integration-163-167-168` @ `bf46c072`

Ran the merged 163+167+168 tree on metro-win (Win 11 Pro Workstations, x86_64-pc-windows-msvc, RTX A5000). Daemon stopped for the run, so it's comparable to the macOS baseline.

**1248 passed · 2 failed · 11 ignored** across 67 targets — vs macOS 1245 / 0.

The only two failures are the shutdown probes, unchanged from my earlier run on `8514a0bf`:

```
test grpc_server::tests::server_binds_and_shuts_down_cleanly ... FAILED    finished in 10.82s
test grpc_server::tests::chat_returns_invalid_argument_on_empty_inbound_stream ... FAILED  10.81s
```

### `control_server` — 19/19, 2.26s

The loopback-TCP conversion works on Windows. This is the part of the PR that does what it says.

### The shutdown probes — timing is the signal

Not milliseconds, not ~1–1.5s: **each burns the full 10s `wait_for` budget and then fails.** Identical with the daemon stopped, so contention from a live node is ruled out.

This branch still has `Duration::from_millis(1500)` at `grpc_server.rs:2430`. Measured Winsock loopback refusal latency on this machine is **~2.04s** — same figure for a never-used port and for a properly-closed listener, so it is OS behaviour, not the server:

```
never-used port:        REFUSED  2124 / 2039 / 2040 ms
bind -> close -> conn:  REFUSED  2028 / 2044 / 2029 ms
```

Because `tcp_refused` maps timeout to `Err(_) => false` ("still up"), every probe expires at 1500 ms before Winsock answers at ~2040 ms. It cannot return `true` on Windows regardless of listener state — hence the full 10s burn rather than a fast fail. Changing only `1500` -> `3000` makes both green (13/13, suite 21.65s -> 4.20s).

So the listener does close correctly; the probe just can't observe it. Detail in my earlier comment above.

### Findings for the other two PRs

- **#167** — `fetch-patches.sh` ran clean and fetched `libp2p-kad 0.48.0`; both patch dirs present. Confirmed the README doc bug (commented on #167).
- **#168** — `bootstrap_reseed` did not reproduce the tick race here: 180 runs, 0 failures, including single-core pinned (commented on #168).

Environment: `cargo test --workspace --no-fail-fast`, daemon stopped, no leftover cargo/rustc processes.

dazwin:
Thank you for the hardware run — but the tree you ran is stale, and the two failures are the bug you already found, not a new one.

`test/integration-163-167-168` @ `bf46c072` was cut at `8514a0bf`; `0cfb47db` — the commit that answers your measurement — is not an ancestor of it. That is why you still see `Duration::from_millis(1500)` at `grpc_server.rs:2430` and the full 10 s burn: you re-measured the same defect on the same code. Confirmed here:

```
$ git merge-base --is-ancestor 0cfb47db origin/test/integration-163-167-168; echo $?
1
```

I have rebuilt the integration branch from the three current PR heads (`0cfb47db`, `5850e23d`, `0b042bc8`) on the same `61304fc0` base, so a re-run picks up the fix. On that tree `tcp_refused` has a **4 s** ceiling — 2× your ~2.04 s figure, with the measurement recorded next to the constant — and the probe returns a tri-state, so a failure now says `still Accepting` or `still Silent` rather than blaming the listener.

The run worth having on metro-win is:

```
cargo test -p kwaainet --bin kwaainet grpc_server
```

Expected 13/13 with the two shutdown tests each around 2 s, so a suite near 4 s — matching your 3000 ms experiment. If you see `still Silent` in a failure, that is the ceiling under the floor again and the message will say so; `still Accepting` would be a genuine listener leak.

Two things from your comment I have not folded in, deliberately:

- **`metrics.jsonl` dirtying the working tree** — pre-existing on `main` (last written by #96), and this branch does not touch `tests/` at all. Left for its own issue rather than folded into a Windows test fix.
- **Windows test job in the PR gate** — agreed, and this PR is the argument for it. Worth doing as a follow-up the moment this lands, alongside #144.

dazwin:
Integration branch rebuilt and pushed: `test/integration-163-167-168` is now `efa8a2c4`, merging `0cfb47db` (#163) + `5850e23d` (#167) + `0b042bc8` (#168) onto the same `61304fc0` base. macOS: **1197 passed, 0 failed** across the workspace.

dazwin:
## The ceiling is gone — `e307da64`

@RezaRassool your Winsock measurement turned out to be the beginning of the answer rather than the end of it. Two commits since, and the second removes the thing we were arguing about.

**First, a second measurement.** A different Windows box — Win 10 22H2, i7-3667U 2c/4t, a generation slower than metro-win — gives **~2.30 s mean, 2.43 s worst**, against your ~2.04 s. Both stock: `MaxSynRetransmissions = 4`, no `TcpInitialRTT`/`TcpMaxConnectRetransmissions` overrides on either. Same retry count, different per-retry latency, so the floor **tracks the hardware** — in the direction a shared CI runner sits. Your figure was a floor observation, not a floor, and 4 s was 1.65× the slower reading rather than the 2× a single measurement implied.

`d97b8dec` raised it to 10 s on asymmetric-cost grounds, and recorded both numbers with hardware attached — a bare constant invited this mistake twice.

**Then the actual fix.** Raising the number was still tuning a threshold that guesses at a machine, which is the real defect. `e307da64` stops asking the question: the probe now tries to **`bind`** the port rather than inferring closure from silence.

Measured on that same Windows box, modelling the server's own listener (`grpc_server.rs:2033` — plain `std::net::TcpListener::bind`, no socket options; std sets `SO_REUSEADDR` on Unix only):

- against a live listener: fails in **<1 ms** with `WSAEADDRINUSE`. On Unix, `SO_REUSEADDR` permits `TIME_WAIT` but never an active listener.
- after a connection closed **server-side**, leaving `TIME_WAIT` on that exact port: succeeds in **~0.3 ms**, with the entry visible in `netstat` immediately before the bind — not a won race.

That second case was the one that could have sunk the approach, since it would report a correctly closed listener as still up. Worth noting how nearly it was missed: I originally specified closing the *client* first, which strands `TIME_WAIT` on the client's ephemeral port and exercises nothing — it would have returned five clean greens against a hazard never tested. Both close orders were run.

The trade, stated plainly: this asks "is the port bindable", not "did *our* listener close", so something else taking the port reports a leak that isn't there. Spurious **failure**, never a silent pass — a leaked listener still holds the port. Given this branch has twice shipped a test that could not fail, erring toward false alarms is the right direction.

`REFUSAL_PROBE_CEILING` is deleted. `LISTENER_CLOSE_BUDGET` survives at 10 s as an ordinary liveness budget — it bounds how long a real leak takes to report, and nothing is inferred from it.

**Falsifiability checked rather than asserted**, since that is the specific failure this branch keeps reproducing: `mem::forget` on the handle fails with `127.0.0.1:53796 still not bindable 10s after dropping the handle`. Unmodified: 219 pass, clippy `-D warnings`, fmt clean.

PR body rewritten — the old one described `8514a0bf`'s "silence now counts as closed", which has been wrong for two commits.

**What would be useful from metro-win**, if you have a moment: `cargo test -p kwaainet --bin kwaainet grpc_server` on `e307da64`. The number to watch is different now — both shutdown tests should finish in **milliseconds**, not ~2 s, since nothing waits on Winsock any more. Seconds would mean my reasoning is wrong somewhere. A run of the `mem::forget` check on your hardware would be worth more than a green: Windows is the platform where both previous versions were unfalsifiable, and neither of us can reproduce that from macOS.

I have not rebuilt `test/integration-163-167-168` onto `e307da64` yet — waiting on the in-flight Windows run against this branch first, so a failure there is attributable to one PR rather than three.

dazwin:
## Verified on Windows hardware — `e307da64`, both halves

Run on a second Windows box (DARREN-WINPC, Win 10 22H2, i7-3667U 2c/4t — deliberately slower than metro-win), toolchain 1.93.1 per the repo pin.

### Fast when correct

```
test result: ok. 13 passed; 0 failed; 0 ignored; 0 measured; 211 filtered out; finished in 0.13s
```

Per-test, three runs each:

| test | libtest |
|---|---|
| `server_binds_and_shuts_down_cleanly` | 0.04s, 0.05s, 0.04s |
| `chat_returns_invalid_argument_on_empty_inbound_stream` | 0.05s, 0.06s, 0.06s |

**40–60 ms each, against a Winsock refusal floor of ~2300 ms measured on that same box** — a same-machine comparison, not one across hardware. Nothing in the path waits on the refusal timer any more.

### Fails when broken

`drop(handle)` → `std::mem::forget(handle)` in `server_binds_and_shuts_down_cleanly` only, leaving the other two drop sites intact so a broken harness can't pass for a working probe:

```
thread '...' panicked at crates\kwaai-cli\src\grpc_server.rs:2528:13:
127.0.0.1:20046 still not bindable 10s after dropping the handle
test result: FAILED. 0 passed; 1 failed ... finished in 10.04s
```

`wait_for_close`'s own panic, naming the port the leaked listener actually held — not a compile error, a hang, or some other assertion firing first. It consumed `LISTENER_CLOSE_BUDGET` and stopped, so a genuine leak is reported in bounded time rather than hanging the suite.

**Same test, same box, same binary: 0.04 s with `drop`, 10.04 s with `mem::forget`.** A ~250× separation between the healthy and leaked paths, with no interpretation of silence anywhere in it.

### Why both halves were needed

Neither is sufficient alone, and this is the distinction the branch kept getting wrong. Green proves nothing waits on Winsock; it does not prove a leak would be caught. The `mem::forget` run proves the second. Both previous versions of this test would have passed the first check and failed the second — on Windows specifically, which is why neither Reza nor I could have caught it from macOS.

That is the property #163 has been trying to establish for three attempts, and it is now measured rather than argued, on the platform where it kept failing.

---

Two build-system defects surfaced on the way, both unrelated to this PR and both invisible on a fully-provisioned machine. Raising them separately:

- **Go is an ungated hard requirement for `kwaainet`.** `kwaai-p2p-daemon` is a non-optional dep of `kwaai-cli` (`Cargo.toml:86`, nothing in `[features]` touches it) and its `build.rs:159-179` panics without a Go toolchain, with no env override — so a clean Windows machine panics ~13 minutes into its first `cargo test`, and this holds even for a `native_p2p: true` deployment whose whole purpose is not running p2pd.
- **`kwaai-rpc/build.rs` treats exit status as download success.** A truncated protoc download (1.1 MB of ~4 MB, intact `PK` header) surfaced as `Expand-Archive`'s "End of Central Directory", then as a misleading "Could not find protoc". Three variants of one root cause: no checksum on the download (`:54-75`), `curl -L -o` with no `--fail` so a 404 body is written and exits 0, and `.output().is_ok()` at `:13` treating mere process spawn as a working protoc — while never consulting `$PROTOC`, the escape hatch prost-build names in the very error it produces.

The integration branch has been rebuilt on the three current heads for combined validation.

dazwin:
Integration branch rebuilt on the current heads: `test/integration-163-167-168` is now `c0e93746` — `e307da64` (#163) + `5850e23d` (#167) + `0b042bc8` (#168) on the same `61304fc0` base. macOS: **1197 passed, 0 failed** across the workspace. Combined validation running on Windows hardware now, including a repeat of the `mem::forget` falsifiability check against the merged tree — #167 and #168 both touch `kwaai-p2p`, so that property is worth re-establishing after the merge rather than assuming it survived.

dazwin:
## Combined validation on Windows — `c0e93746`, all three PRs

Second Windows box (Win 10 22H2, i7-3667U 2c/4t), `cargo test -p kwaai-p2p -p kwaainet`, toolchain 1.93.1 per the repo pin.

**433 passed, 0 failed, 5 ignored** across 19 suites, exit 0. The 5 ignored are the `live_*` tests that self-gate off a live network.

| suite | | suite | |
|---|---|---|---|
| `kwaainet` unittests | 225 | `swarm` | 16 |
| `service_unary` | 19 | `raw_stream` | 14 |
| `reachability` | 12 | `relay` | 10 |
| `unary` | 10 | `dht_service` | 6 |
| `quic` | 3 | `port_reuse` | 2 |

#168's areas are directly exercised — `relay` 10/10, `reachability` 12/12, `dht_service` 6/6, `swarm` 16/16. #167's kad config rides in the `kwaai-p2p` lib tests. #163 is the 225 `kwaainet` unittests.

### The probe is unchanged by the merge

Both #167 and #168 touch `kwaai-p2p`, so this was a real question rather than a formality:

| | `e307da64` alone | `c0e93746` merged |
|---|---|---|
| `grpc_server` suite | 13 passed, 0.13s | 13 passed, **0.13s** |
| `server_binds_and_shuts_down_cleanly` | 0.04s | **0.04s** |
| `chat_returns_invalid_argument_on_empty_inbound_stream` | 0.05s | 0.06s |
| same test, `mem::forget(handle)` | FAILED, 10.04s | **FAILED, 10.03s** |

```
127.0.0.1:20692 still not bindable 10s after dropping the handle
```

Same panic site, same message, different ephemeral port from the isolated run (20692 vs 20046) — so it's a live bind each time, not a fixture. **The falsifiability property survived the merge**: the probe still detects a leaked listener, not merely an absent one.

---

### Follow-up for the Windows-in-CI work

Getting this run to complete took three aborts on a disk-constrained box, and the binding constraint was not parallelism or incremental compilation — it was **debug symbols: ~300 MB of pdb per test binary, ~6 GB total, for a pass/fail run that reads none of them.** `target/` went 6.9 GB → 13 GB on three crates.

`CARGO_PROFILE_DEV_DEBUG=0 CARGO_PROFILE_TEST_DEBUG=0` kept peak inside 10.8 GB and finished in 17 minutes. Recommend that as the default for any Windows test job — runner disk is the constraint that has kept the Windows *test* job out of the PR gate, and the only cost is symbolised backtraces on a panic, which a single-test rerun restores cheaply.

Also worth knowing for branch-hopping verification: cargo never garbage-collects artifacts across commits. Three stale `libkwaai_p2p` rlibs (637M / 637M / 418M) from successive checkouts were sitting in `target/debug/deps`; `cargo clean -p` freed 9.7 GB across two passes.

RezaRassool:
## Windows verification on metro-win — green

Read-only run on real hardware: Windows 11 Pro for Workstations 10.0.22631, `x86_64-pc-windows-msvc`.

Tree: `test/integration-163-168` @ `7a956833`, merging **#163 @ `e307da64`** and **#168 @ `0b042bc8`** — both confirmed current PR heads and ancestors of the tested tree. Nothing merged, pushed, tagged, or modified; no binary installed.

```
cargo test --workspace --no-fail-fast
1253 passed, 0 failed, 11 ignored, across 67 targets — exit 0
```

`--no-fail-fast` throughout, so nothing is hidden behind a first failure. `core/patches/multistream-select/` was already present, so the build never hit `failed to load source for dependency`.

### The count reconciles against the macOS baseline

1253 here against 1248 on macOS. The difference is five `#[cfg(windows)]` tests in `kwaai-cli/src/updater.rs` that can only run on this platform — the four `backup_path_*` cases and `windows_gpu_detected`. All five passed. Same suite, plus the Windows-only ones, not five tests behaving differently.

### The targets that carried the risk

| Target | Result |
|---|---|
| `kwaai-p2p-daemon` → `control_server` | **19 passed, 0 failed** — the full set, over loopback TCP |
| `kwaainet` → `grpc_server::tests::server_binds_and_shuts_down_cleanly` | ok |
| `kwaainet` → `grpc_server::tests::chat_returns_invalid_argument_on_empty_inbound_stream` | ok |
| `kwaainet` → `node_native::tests::a_bootstrap_preset_publishes_nothing_and_a_node_still_does` | ok — the `KWAAINET_SOCKET` multiaddr fixture holds |
| `kwaai-p2p` → `bootstrap_reseed` | 2 passed |
| `kwaai-p2p` → `relay_manager::tests::a_discovered_relay_that_only_gained_an_address_is_not_moved` | ok |

### The shutdown probes: the Winsock cost is gone

Timed individually:

```
server_binds_and_shuts_down_cleanly                    finished in 0.04s
chat_returns_invalid_argument_on_empty_inbound_stream  finished in 0.06s
```

40 ms and 60 ms, against the ~2.04 s / ~2.30 s the connect-probe version cost on Windows boxes. Asking whether the port can be *bound* rather than how long a refused connect takes to report does exactly what it was meant to.

### `bootstrap_reseed`: no flakiness in four runs

```
run 1: 2 passed  2.29s
run 2: 2 passed  2.18s
run 3: 2 passed  2.21s
run 4: 2 passed  2.17s
```

A 0.12 s spread, well clear of the 500 ms tick an earlier version raced. Four runs cannot prove a rare race absent, but nothing here suggests one.

### Full per-target breakdown

<details>
<summary>57 targets with tests; 10 empty targets omitted</summary>

```
   13p  kwaai_compression      lib             24p  01_unit_dht                 103p  kwaai_p2p            lib
   30p  kwaai_distributed      lib              7p  02_unit_p2p                   2p  bootstrap_reseed
   68p  kwaai_hivemind_dht     lib             16p  03_unit_rpc                   2p  connection_limits
   24p  kwaai_inference        lib              9p  04_integration_daemon         3p  dcutr
                                                5p  05_integration_relay          6p  dht_service
    9p  kwaai_p2p_daemon       lib              7p  06_network_bootstrap          2p  port_reuse
   19p  control_server                          7p  07_wire_interop               3p  quic
                                                3p  08_unary_swarm_interop       14p  raw_stream
  194p  kwaai_rag              lib              4p  09_service_unary_interop     12p  reachability
   33p  axiom                                   3p  10_native_dht_interop        10p  relay
    9p  pipeline                                5p  11_control_server_interop    19p  service_unary
  210p  suite                                   4p  12_pipe_mode_interop         15p  swarm
                                                5p  13_native_node_assembly      10p  unary
   42p  kwaai_storage          lib              5p  14_bootstrap_seed
   10p  api                                     4p  15_decentralized_dht        252p  kwaainet     main.rs
    6p  storage                                                                  12p  proto_roundtrip
    6p  kwaai_trust            lib          Ignored (11): live_bootstrap 2,       2p  kwaai_wasm   lib
                                            live_circuit_reachability 1,
   Doc-tests: 5 passed, 5 ignored           live_dht_announce 1, live_map_probe 1,
                                            bench_trig 1, doc 5
```
</details>

No failures, so there is no failure output to report.

### Minor, not blocking

Compilation emits unused-import and unused-variable warnings in `kwaai-p2p-daemon` — the `sock` bindings behind the TCP path, `DEFAULT_SOCKET_NAME`, `std::os::windows::process::CommandExt`, and a few others. Cosmetic leftovers of the Unix-socket removal, and `clippy -D warnings` is not run over that configuration here. Left untouched.

### Scope

This covers `cargo test --workspace` only. `clippy`, `fmt`, and the release build were not run on this box, and neither was any live-network target — the 11 ignored tests are the `live_*` ones, which stay ignored by default.

🤖 Verification run driven by [Claude Code](https://claude.com/claude-code)

PR #164, opened 2026-08-31
docs(knowledge): extend the integrity checker to plans, and record the evidence-vector null

RezaRassool:
Two related things: plans were drifting the same way `CLAUDE.md` files do, and an experiment had run without its result being written down.

## 1. The integrity checker now covers `projects/*/plans/`

`scripts/check_memory_integrity.py` was built because `CLAUDE.md` files describe code that moves without anyone opening them. Plans have the identical failure and were not covered. Six days after it was written, `PerKBOntology-plan.md` still said **"not yet started"** though #150 had merged phases 0–2, and `D6-OntologyAB-testplan.md` still said **"not yet run"** with a prerequisite section describing work already done.

Plans can't take the `CLAUDE.md` rules unmodified — a plan legitimately names files it *proposes* to create, and applying those rules produced 42 errors that were almost all proposals. So severity is split:

| check | memory files | plans |
|---|---|---|
| unresolvable file/fn ref | error | warning (`proposed?`) |
| `file.rs:N` whose symbol is defined elsewhere in that file | error | error |
| `Status: not yet …` while the described code moved | — | warning |
| age drift | warning | not reported (a results doc is meant to be old) |

The line-anchor check is the one with teeth. It found **19 stale pointers** — `IN_SCOPE_RELATION_TYPES` documented at `relation_extract.rs:76` when it's at `:100`, `expected_fields` at `graph.rs:343` when it's at `:600`, and so on. All fixed here.

Line numbers are branch-relative, so a PR inserting lines above a definition stales every pointer to it in docs it never touched. That's real staleness, but if the remedy were manual the check would get silenced — so **`--fix`** rewrites them in one command.

**`--self-test`** covers the behaviours above on synthetic fixtures and runs in CI ahead of the check itself. Every case in it is a bug hit while building this; proximity-pairing (a line naming two symbols and two line numbers reported both as wrong) is the sharpest.

Also disambiguates the two version numbers in `ontologies/D6.yaml`: the comment header's "Version 3" counts authored derivations, `ontology.version` counts file revisions and is **9**, and the plans cite the latter.

## 2. The evidence-vector null, recorded

`PredicateVectors-exploration.md` §5 listed "embed the evidence, not the label" as open. It wasn't — it ran on 2026-08-26 and its outputs sat untracked, so the plan asked for work already done.

New §6: embedding a predicate's evidence separates same-predicate from different-predicate edges by **+0.000** at chunk granularity and **−0.004** at span granularity. Both null.

**The sample can't carry that conclusion**, and §6 says so rather than banking a clean negative. Only three predicates cleared n ≥ 3 in either run, and they're the vaguest in the vocabulary — two locatives and the escape hatch itself. Every kinship predicate fell below the floor, which is exactly where §2 found the disqualifying polarity failure and where D6 is densest. The hypothesis is untested where it matters, not refuted.

The span run failed for a second reason worth keeping: of the production D6 graph's 340 relations only **21** had a locatable span, because the second endpoint is usually a pronoun. That makes `Phase6-PersistentMentionIndex` a prerequisite for retrying this rather than unrelated work.

Also records that `D6-FullAB-results.md` §5's "cheap fix first" recommendation **is already done** — the kinship aliases are on the built-in `genealogy` module with a test pinning the 7/6/5/5 recovered edges — so the ~9-hour production-config re-run is unblocked.

## Review notes

- `results/predicate_vectors.json` (1.6 MB) and `evidence_vectors.json` (628 KB) are embedding caches. Committed because `results/` already tracks 331 files including a 2 MB one, and they pin the exact measurement. Say the word and they can be gitignored instead.
- `predicate_collapse.html` **is** gitignored — it regenerates from tracked inputs, like `glossary.html` and `ontology_diagrams.html` beside it.
- The `memory` CI job runs the extended check for the first time on this PR. It is green locally on `main`'s line numbers.

https://claude.ai/code/session_01NveCG5VHw4f7osHg2EPugG


RezaRassool (commented):
## Review — plan integrity checker

Checked out the head (`cc733e5a`) in a worktree and ran the checker three ways (`--self-test`, plain, `--fix`) plus instrumented probes over all 47 tracked docs and 60 real `file.rs:N` pointers.

**Verified sound:** `--self-test` passes with 0 failures, `--fix` is idempotent on the branch (0 rewrites), `ontology.version: 9` in `D6.yaml` matches the docstring, the genealogy kinship-alias module and its test exist, and the PredicateVectors §6 `21/340` → 6% reconciliation is arithmetically correct.

### The headline: the checker cannot see the pointers that were actually wrong

`BT_IDENT` (line 108) cannot match `Type.field` or expression forms, so those pointers get no anchor and are dropped **in silence** — 20 of 60 pointers are unanchored, and 27 of 60 are never fully checked.

That is not theoretical. Three genuinely stale pointers survived this PR's sweep because of it:

- `coref-pronoun-resolution.md:46` — `EntityNode.gender` at `graph.rs:220`; actually **251**
- `coref-pronoun-resolution.md:47` — `EntityNode.aliases` at `graph.rs:212`; actually **243**
- `PerKBOntology-plan.md:333` — `entity_id = sha256(...)` at `graph.rs:329`; actually **229** (329 is `default_relation_method`)

I confirmed the first two: `EntityNode` starts at line 228, so both documented pointers land *before the struct exists*. The PR body claims the pointer sweep is complete; these are the ones the checker was built to catch and structurally cannot. (#166's review found the same two independently.)

### `--fix` can destroy correct pointers

Three separate paths, all reproducible:

- **line 184** — the previous-line anchor fallback does not require the previous line to be a wrapped continuation, so it anchors across unrelated lines and table rows. Live case: `confidence-hybrid-extraction.md:22` anchors to `score_entity` from the *previous table row*. Inert today only because `score_entity` is not defined in `graph.rs` — the day such a symbol resolves, `--fix` rewrites a correct pointer to that symbol's line. That is the `--fix` corruption class commit 3 claims to have closed.
- **line 182** — `ANCHOR_SPAN = 60` does not stop cross-cell anchoring. `PerKBOntology-plan.md:261` anchors to `range` 56 characters earlier in a different table cell, under the limit. Self-test case 9 only exercises a 72-char gap, so the real corpus case it was written for slips through.
- **line 183** — `used` is keyed by identifier *name*, not position, so when the nearest identifier is already consumed the code walks *further back* to a wrong anchor instead of skipping. Reproduced on the head: `check_line_refs("see \`BETA\`, \`alpha\` (\`fake.rs:5\`) and \`alpha\` (\`fake.rs:5\`)")` returns a false warning, and `fix_line_refs` would rewrite the correct `:5` to `:40`. A doc line naming the same function twice with two pointers is ordinary prose.

### The one condition that gates CI has no `--fix` remedy

**line 207** — past-EOF stays a CI-gating error, which contradicts the branch-relative argument the same commit uses to demote everything else to a warning: a PR that *deletes* 200 lines from `graph.rs` turns CI red on docs it never opened. And `fix_line_refs` deliberately skips past-EOF refs (`if n > len(src): continue`, line 319), so the only failing condition is the only one `--fix` will not repair.

It can also fire falsely: with a mispaired anchor, `rust_path`'s ambiguous-basename tie-break can resolve `config.rs` to `summit-server/src/config.rs` (31 lines) rather than `kwaai-cli/src/config.rs` (1985), turning a correct `config.rs:1017` into a hard error.

### Smaller

- **line 299** — `check()` runs `check_line_refs` on `strip_foreign_sections(txt)` but `fix_line_refs` reads the raw file, so `--fix` rewrites pointers inside sections the checker deliberately excludes as belonging to another repo, validating them against a same-named file in *this* tree. `projects/kwaai-storage/CLAUDE.md` already has such a section; latent rather than firing.
- **line 401** — self-test case 8 ("ident ending in rs") is vacuous. The fixture is `let parse_headers = 1;` and `DEF_ANY` matches only `const|static|fn|struct|enum|trait|type|mod`, so `def_span` returns `None` and the assertion holds either way. I verified it also passes with the removed `endswith("rs")` filter reinstated — it does not regression-test the bug it names.
- **line 237 / 42** — stale comments from commit 3: "The line-anchor check below stays an error for both" is no longer true, and the module docstring's "Exit non-zero when a file reference does not resolve" no longer holds for plans.

### Two notes

- `plans/_base.docx` (19 KB binary) arrives in commit `cc733e5a`, whose message is entirely about the checker and never mentions it. It is a real input to `build_report_docx.py:98`, so arguably legitimate — but unexplained in a fix commit.
- Net effect on CI today: green, 0 errors and 43 warnings, and warnings never fail the build. The headline miss this PR was written to catch (`PerKBOntology-plan.md` saying "not yet started") is now one warning among 43, so nothing gates on it.

dazwin (changes requested):
## Review

The existing review is accurate; I re-ran the head and confirmed `--self-test` passes (0 failures), the plain run is 0 errors / 43 warnings, and `EntityNode.gender` / `.aliases` are unanchored because `BT_IDENT` cannot match `Type.field`.

### Request changes on the `--fix` paths
The three corruption paths in the existing review are real by inspection of `anchored_refs`:
- previous-line fallback anchors across unrelated lines and table rows;
- `ANCHOR_SPAN` does not respect table-cell boundaries;
- `used` is keyed by identifier name rather than position, so a repeated symbol walks back to a wrong anchor.

`--fix` is not run in CI, so none of these breaks the build, but it should not be run unattended until they are fixed. Also fix the stale comments ("stays an error for both", and the docstring's "Exit non-zero when a file reference does not resolve" no longer holds for plans).

### Verified separately
- Ran the checker over #170's new plan: it pairs `db.rs:131` with `BRUTE_FORCE_THRESHOLD` (nearest backticked symbol) when the prose means `search()`. Same nearest-symbol pairing class as above — worth a self-test case.
- `played_with` / `played_at` are absent from `D6.yaml` here too, so the "cheap fix has landed" note in `D6-FullAB-results.md` is incomplete (see #166).
- `_base.docx` is a real input to `build_report_docx.py`, but it arrives in a commit whose message never mentions it.

The 2.2 MB of embedding-cache JSON is consistent with what `results/` already tracks; no objection.

PR #165, opened 2026-08-31
feat(rag chat): show retrieval as it happens, paced to cover the answer

RezaRassool:
## Why

A `kwaainet rag chat` turn was silent from Enter until the whole answer landed: retrieval →
rerank → one blocking `"stream": false` completion → a single `println!`. The retrieval
narration and the retrieved passages were already computed, then thrown away.

This surfaces that work as it happens, paces it to cover the generation time, and adds the
seams in `kwaai-rag` that the chat UX needs.

## What

Five commits, each standing on its own — the branch bisects cleanly and every commit was
verified to build standalone:

- **`c826633c`** `kwaai-rag` groundwork: reranker dedup fix, `context_plan`, `ChunkOrigin`,
  the marks API, `feedback.rs`, `judge.rs`
- **`0a7278c6`** the UX: retrieval narration, a findings table, completion running in
  parallel with narration, line commands
- **`97118c3d`** quote the real evidence window from the ingest mention index rather than a
  guessed sentence; restores `chunk_id`
- **`10269395`** preload the models at startup instead of on the first question
- **`5e2c8ba4`** stretch the narration to land with the answer

### Elastic narration pacing

The narration used to run at a fixed `--pace`, finish early, and hand over to a bare
"generating" spinner — the same dead air, moved later in the turn. The pace is now
re-derived before each unit, spreading what is left of the script across the time still
expected on the answer.

- `--pace` is the **floor, never the target**: this only slows narration down.
- The stretch caps at **8×**, so an over-estimated deadline cannot leave text crawling.
- The estimate is the session's own generation times, halved into a running average, seeded
  high (25s) on purpose — narrating long is invisible because the answer cuts it off, while
  narrating short puts the spinner back.
- Pacing is skipped entirely when stderr is not a terminal.

## The invariant

stdout carries only `You:` / `Assistant:`; every other surface goes to stderr.
`test_chat_stdout_clean.sh` guards a real bug where chat read its own stdout back as input
and looped forever. Two further tests pin the split from the other side, so the first cannot
be "fixed" by deleting the narration.

## Review

This branch was reviewed after the feature work landed. **15 findings, all fixed in this
PR**, each folded into the commit that introduced the bug rather than bolted on at the end.
The full review is in the PR comments.

Three were blockers, and each one's regression test was verified to **fail against the old
code** before the fix went in:

1. **A panic in the evidence quote.** The trigger offset was found in a sanitized copy but
   its length carried over from the unsanitized span, so a double space (routine in
   PDF-extracted text) put the slice end inside the following multi-byte character:
   `byte index 11 is not a char boundary; it is inside '—'`. Fixed at both levels — derive
   the length from the needle, and snap to a char boundary inside `truncate_around`.
2. **EOF spun the REPL forever.** `read_line` reports EOF as `Ok(0)`, not `Err`, so a piped
   session without a trailing `exit` — or an interactive Ctrl-D — looped on an empty line at
   full CPU. The existing piped test missed it because its input ends with `exit`.
3. **The citation checker contradicted the prompt it was checking.** The prompt numbered
   excerpts by `chunks.len()` while `check_citations` bounded them by `plan.len()`. With a
   1000-char budget over ten chunks the prompt advertised ten sources and rendered two —
   not merely false warnings, but an invitation to cite `[3]`–`[10]` with nothing behind
   them. Both prompt builders now number from the plan they render.

The other twelve: silent fallback when a graph mode's store fails to open; evidence quotes
missing their sentence terminators; a single mark switching on the LLM reranker for the rest
of the session; marks persisting across questions though keyed per-query; `/exit` ignored at
the `You:` prompt; `--no-cache` still writing to the cache; the retry notice colliding with
the redrawing status line; model warm targeting the wrong endpoint on a split Ollama setup;
the budget-dropped count including rows not on screen; `term_width().max(80)` ignoring a
narrower `KWAAINET_WIDTH`; `?N` showing less than the model was given; and the query
embedding being computed even where it is discarded.

## Design notes

- Completion runs in parallel with narration, so a mark applies on `/retry` rather than to
  the in-flight answer.
- Line commands over `read_line`, not a TUI crate — Windows stays first-class and this adds
  no dependencies. The Windows CI check passing is the evidence.
- Completion stays non-streaming because the `p2p://` relay proxy is strictly
  request/response.

## Testing

- **685 unit + integration tests pass** (669 before the review fixes)
- `cargo clippy --all-targets` clean, `cargo fmt --check` clean
- All three chat shell tests — eleven assertions — pass against the live D6 KB
- Release binary built and exercised by hand against D6

Rebased on `main` at v0.6.4; no conflicts with the per-KB ontology work in #150.

## Known follow-ups (not in this PR)

- The graph bullet is still thin (`• Hanaffi Quwatul Islam Mosque`, no detail). The richer
  fix uses the relation `evidence_window`/`evidence_span()` from #150, now merged.
- The LLM judge costs a round trip before the mark prompt (~1–2s). It runs after the answer
  prints so it never delays the answer, and `--no-judge` skips it; properly backgrounding it
  is the better fix.
- Pre-existing and untouched: `unused import: super::*` at `kwaai-rag/src/ontology.rs:1665`,
  introduced by #150.


RezaRassool (commented):
## Automated code review — 15 findings

All nine CI checks pass on this branch, so none of the below is caught by the current gate.
Six findings were verified directly against the branch source; the other nine are reported
with reasoning but their call sites were not all opened. Ordered by severity.

Per the project rule that every bug fixed ships with a regression test that would have
caught it, each proposed fix below names the test worth adding.

---

### Blockers

**1. `core/crates/kwaai-cli/src/chat_ui.rs:87` — evidence quote can panic on a non-char-boundary slice** ✅ verified

`b` is built from a *pre*-sanitize byte length but indexes into the *post*-sanitize string:

```rust
let (a, b) = match clean.find(&sanitize_cell(&w.text[w.trigger_start..w.trigger_end])) {
    Some(i) => (i, i + (w.trigger_end - w.trigger_start)),
```

`sanitize_cell` collapses whitespace runs, so a trigger span like `"Hanover  Street"` yields a
14-byte needle while `trigger_end - trigger_start` is 15. `truncate_around` (`display.rs:124`)
then evaluates `&s[start..end]` and `&s[end..]` guarded only by `.min(s.len())` / `.clamp()` —
those bound the index but never snap to a char boundary. One multi-byte character after the
needle (curly quote, en dash, accented name — all routine in the D6 PDF corpus) panics the turn.

*Proposed fix:* derive `b` from the needle's own length, and make `truncate_around` defensive
regardless:
```rust
let needle = sanitize_cell(&w.text[w.trigger_start..w.trigger_end]);
let (a, b) = match clean.find(&needle) { Some(i) => (i, i + needle.len()), None => (0, 0) };
```
plus `floor_char_boundary`-style clamping on both slice sites in `truncate_around`.

*Test:* an evidence window whose trigger contains a double space and is followed by a curly
quote, asserted to render rather than panic.

---

**2. `core/crates/kwaai-cli/src/rag_cmd.rs:1795` — EOF on stdin spins the REPL forever** ✅ verified

`read_line` returns `Ok(0)` at EOF, not `Err`, so this guard never fires:
```rust
if stdin.lock().read_line(&mut line).is_err() { break; }
let q = line.trim().to_string();
if q.is_empty() { continue; }
```
`q` is empty → `continue` → reprint `You: ` → read `Ok(0)` again, at full CPU, forever.
`echo "who is the author?" | kwaainet rag chat` hangs; interactive Ctrl-D does the same.

Note `test_chat_piped_no_prompt.sh` does **not** cover this — its input ends with an explicit
`exit\n`, so the loop breaks before EOF is ever reached.

*Proposed fix:* `if stdin.lock().read_line(&mut line).map_or(true, |n| n == 0) { break; }`

*Test:* extend the piped test with a case whose input has **no** trailing `exit`, asserting
the process exits 0 within the cap rather than being killed at 124.

---

**3. `core/crates/kwaai-cli/src/rag_cmd.rs:2081` — citation check contradicts the prompt it checks** ✅ verified

`check_citations(&answer, plan.len())` bounds citations by the planned chunks, but
`prompt.rs:49` still numbers the excerpts by the full set:
```rust
let n = chunks.len();
// "...are numbered [1]–[{n}] ... Only cite numbers that exist: [1]–[{n}]"
```
Whenever `context_plan(&chunks, 24000)` drops anything — which this PR explicitly expects and
narrates as *"N marked · did not fit the context budget"* — any citation in
`(plan.len(), chunks.len()]` fires the *"answer cites N sources that were not supplied"*
warning even though the model obeyed its instructions exactly.

*Proposed fix:* have `build_chat_messages` take (or return) the plan length and use it for `n`,
so the prompt and the checker agree on one number.

*Test:* a chunk set that overflows the 24 000-char budget, asserting `check_citations` stays
silent for a citation the prompt authorised.

---

### Correctness

**4. `rag_cmd.rs:1898` — `--mode graph` / `--mode iterative` degrade silently.**
The match on `(effective_mode_chat, session_graph)` only handles `Some(graph)`; when the graph
store failed to open it falls through to `_ => retrieve_hybrid(..)` with no message, replacing
what used to be `GraphStore::open(..).context(..)?`. The one warning that exists covers the
remote-storage case, not a failed local open. Same silent-degradation mode as the 2026-07-21 D6
rebuild. *Fix:* warn explicitly (or hard-error) when the requested mode cannot be honoured.

**5. `core/crates/kwaai-rag/src/mentions.rs:447` — the "real evidence window" is not verbatim.**
`resolve_chunk_mentions` splits on `['.', '?', '!', ';']`, which **consumes** the delimiters, and
drops fragments under 12 bytes; `pick_evidence_window` rejoins with a bare space. The result is
rendered inside curly quotes as if quoted from the corpus: `"Before Mike Allie He owned the shop"`.
The unit tests miss it because they hand-build `SentenceMentions` whose `sentence` fields still
carry their periods. *Fix:* persist the terminator at ingest (preferred), or rejoin with `". "`.

**6. `rag_cmd.rs:1980` — a single mark silently switches on the LLM reranker.**
`if rerank || !marks.is_empty()` sends marks through `rerank_listwise`, so the first `-2` in a
session started without `--rerank` adds an unrequested LLM round trip to every later turn and
reorders the pool, hiding what the mark did — contrary to `apply_marks`' own docstring. It also
truncates to `top_k`, which drops real context off the tail of `retrieve_graph_anchored`'s
prepended entity/fact cards. *Fix:* apply `apply_marks` deterministically; only call the
reranker when `rerank` is set.

**7. `rag_cmd.rs:1763` — marks are session-global but persisted per-query.**
`FeedbackStore::record_marks` keys on `(kb, query, doc_name, chunk_index)`, yet the in-memory
`marks` map is never reset between questions. Item numbers are recomputed per turn, so `-2` on
Q1 and `-2` on Q5 are different chunks — while the Q1 chunk stays invisibly suppressed all
session. *Fix:* key the map by query, or clear it on a new question and say so.

**8. `rag_cmd.rs:1796` — `/exit` and `/quit` do not work at the `You:` prompt.** ✅ verified
`parse_cmd` accepts the slash forms and the banner now advertises `/help`, but the top-level read
tests only `q == "exit" || q == "quit"`, so `/exit` is sent to the retriever as a question.
*Fix:* route the top-level line through `parse_cmd`.

**9. `rag_cmd.rs:2095` — `--no-cache` skips reads but still writes.** ✅ verified
Help text says "Skip the semantic query cache for this session"; `no_cache` gates the lookup at
1811 but the store is gated only on `!is_error && !from_cache`, so the session keeps mutating
`query_cache.db`. *Fix:* add `&& !no_cache` to the store condition.

**10. `rag_cmd.rs:1435` — the retry notice fights the status line.**
`eprintln!("  ⟳ inference attempt 2/2 …")` runs in the spawned completion task while the main
task is paced-typing narration or repainting the spinner. `progress.rs` documents that a
redrawing bar cannot share a stream with a second writer. *Fix:* route it through the same
`StatusLine`/channel as everything else.

**11. `rag_cmd.rs:1689` — model warm targets the wrong endpoint on a split setup.**
`ollama_backed` is true if **either** endpoint answers `/api/ps`, but the per-turn warm only hits
the inference URL. With an Ollama embed host and a non-Ollama inference host, every turn spawns a
doomed 300 s-timeout task, while the embed model's `keep_alive` is never refreshed and is evicted
after Ollama's 5-minute default — reintroducing exactly the dead time this section removes.
*Fix:* track residency per target and warm each endpoint that answered.

**12. `chat_ui.rs:322` — the budget-dropped count points at invisible rows.**
`rows.iter().filter(|r| !r.sent()).count()` includes graph rows (rendered as bullets, never with
a `·`) and corpus rows past `max_rows` (folded into "… and N more"). *Fix:* count only rows
actually rendered in the table.

**13. `rag_cmd.rs:1756` — `term_width().max(80)` defeats a narrow override.**
The comment promises "a larger KWAAINET_WIDTH still wins", but `.max(80)` silently discards a
*smaller* one, so an 80-column terminal (or `KWAAINET_WIDTH=70`) gets a table that soft-wraps and
destroys the box alignment. *Fix:* honour an explicit override in both directions.

**14. `rag_cmd.rs:2110` (`ChatCmd::Show`) — `?N` shows less than the model got.**
The handler prints `chunk_meta.text`, but `context_plan`/`context_text` send `surrounding`
whenever it is longer, so `?N` shows a strictly smaller passage than `[N]` actually cites —
undercutting the point of the command. *Fix:* use `context_text` here.

### Efficiency

**15. `rag_cmd.rs:1809` — the query embedding is computed even when discarded.** ✅ verified
`embed_one` now runs unconditionally, above the `if !no_cache` block. For remote-storage KBs
`query_cache` is `None` (line 1651) and under `--no-cache` it is unused — a wasted round trip per
turn in both. It also runs *before* `LiveNarrator::start`, so it shows as unnarrated dead time
between `You:` and the first status line: the exact symptom this PR set out to remove.
*Fix:* compute it lazily, or move it after narration starts.

---

### Confirmed sound

- The `reranker.rs` pointer-identity dedup fix and its regression test are correct.
- `context_plan`'s byte-for-byte `block_matches_legacy_implementation` test is a good guard on
  the prompt refactor.
- `unsafe impl Send for FeedbackStore` follows the existing `QueryCache` / `GraphStore` /
  `SafeConn` precedent — not new risk.
- `run_completion`'s `&t[..t.len().min(200)]` can still slice mid-UTF-8, but it now runs inside
  `tokio::spawn` and is caught by `h.await.unwrap_or_else(..)`, so this PR strictly improves it.

PR #167, opened 2026-08-31
feat(p2p): kad protocol migration — /kwaai/kad/1.0.0 alongside legacy /ipfs/kad/1.0.0

dazwin:
## What

**The kad protocol set is a build decision.** Serving the default `/ipfs/kad/1.0.0` on a public address gets a node absorbed into the global IPFS (Amino) DHT: the old Python bootstraps held 300–600 foreign peers against a ~12-node real routing table, and p2pd was OOM-killed every 30–90 minutes serving them.

| build | serves |
|---|---|
| stock (`cargo build --release -p kwaainet`) | `/kwaai/kad/1.0.0` only — cannot be recruited as an IPFS DHT *server* (the routing-table/OOM vector; foreign peers still connect below kad — that half is #174) |
| `--features kad-multi-protocol` | both names — **the bootstrap-grade build for the migration window** |

There is deliberately **no runtime knob**. Earlier iterations of this PR shipped a `kad_protocols` config key; review found three ways it silently produced the wrong protocol set (an all-invalid list falling back to the legacy-serving default, a single typo'd entry being dropped, validation split across two crates), and every one of them was a config-path defect. The key is gone from `config.yaml` and `kwaainet config set`; `NetworkConfig::kad_protocols` survives only as `#[serde(skip)]` so the swarm tests can build mixed-protocol topologies.

libp2p-kad 0.48 negotiates a `Vec` of names but removed the public setter, so `patches/` gains a second checksum-pinned patched crate restoring `set_protocol_names` (19 added lines, two methods). **Only the feature build touches the patched API** — which is what keeps `cargo publish -p kwaai-p2p` working (`[patch.crates-io]` cannot travel into a package; verified by dry-run). `fetch-patches.sh` is the new umbrella entry point, and `nix/crane.nix` materializes the patch for every flake target.

## ⚠️ Rollout order is normative, not advisory

**A stock upgraded node shares no kad protocol with any deployed v0.6.x node.** The release that contains this PR is a flag day *unless* dual-protocol bootstraps are serving first — every legacy↔upgraded path then routes through them for the window.

The order (already executed for steps 1–2 at time of writing):

1. Build bootstraps with `--features kad-multi-protocol`, **self-update disabled** (`KWAAINET_NO_AUTO_UPDATE=1` — the release pipeline ships no feature artifact, so a stock update would silently replace the bridge). Deployed 2026-09-01; serving production since the `_dnsaddr` cutover 2026-09-02.
2. Verify each logs `kad protocol set … multi_protocol_build=true` at startup — nothing else distinguishes the binaries at runtime.
3. Only then release this to nodes. Nodes cross over as they auto-update, bridged by the bootstraps.
4. Once the fleet has moved: rebuild bootstraps stock, retire the feature, the patch, `fetch-libp2p-kad.sh` and the crane.nix stanza together. That ends the Amino exposure (#174 tracks connection-management during the window).

Anyone who inverts steps 1 and 3 partitions the fleet. That is the cost of having no runtime knob, and it is paid deliberately: the knob's failure modes were silent, this one is loud and procedural.

## Testing

- Default build: 414 tests; feature build: 197 tests (both crates); clippy `-D warnings` + fmt clean in both configurations; `cargo publish -p kwaai-p2p --dry-run` passes.
- CI now compiles and tests the feature build — previously the bootstrap artifact was the only one nothing compiled.
- The migration topology test resolves peers its caller has **never dialled** (forcing real FIND_NODE round trips on each name), with a negative control asserting a legacy-only bridge cannot resolve across the protocol boundary — the earlier version short-circuited via dial-seeded routing entries and passed with negotiation broken (found in review).
- Live: the dual build has been serving the production DHT under a full Amino inbound flood since 2026-09-02 — ~100 KB/connection, zero restarts, kernel-capped (`MemoryMax`), measurements in #174.


RezaRassool (commented):
## Review — kad protocol migration

Reviewed against merge base `61304fc0`. The core design is sound: the dual-protocol default is backward compatible, the patch is minimal, and the negotiation/identify interplay checks out across all three migration roles (legacy-only ↔ dual ↔ kwaai-only). Confirmed `libp2p-kad` is not auto-adopted as a workspace member, so the missing `exclude` entry is a non-issue, and `Cargo.lock` does not drift.

Five findings, the first two worth resolving before merge.

### 1. `cargo publish -p kwaai-p2p` now fails — and takes the rest of the release with it

`[patch.crates-io]` does not travel with a packaged crate, so the published source hits `no method named set_protocol_names`. Reproduced with `cargo publish --dry-run`.

The knock-on is the part that matters: `publish-crates.yml`'s error grep does not match `failed to verify package tarball`, so the run aborts and **every crate after `kwaai-p2p` in the list never publishes**. `publish-crates` succeeded on v0.6.5; this would break the next release. External crates.io consumers break too.

### 2. The two-entry default drops kad off the V1Lazy 0-RTT path

`raw_stream.rs:101` documents the rule this collides with:

> multistream-select only takes the lazy shortcut for the **last** protocol it has to offer

With `[kwaai, legacy]` the preferred protocol negotiates eagerly. kad opens a substream per request, so that is one extra round trip per DHT hop — the regression V1Lazy was configured to remove (#107), on the same ground as the accepted-call RTT that was a 0.6 cutover blocker.

Not mentioned in the PR, and worth a measurement either way: if it is real, ordering or a migration window that avoids carrying two entries long-term would both be options.

### 3. `nix/crane.nix:62` — the riscv release build will not resolve the new patch

The flake's src derivation materializes only `patches/multistream-select`; `core/patches/libp2p-kad/` is gitignored, so `nix build .#kwaainet-riscv64-linux-gnu` in `release-riscv.yml` fails to resolve the path patch. Tag-triggered only, so PR CI stays green and this surfaces at release.

### 4. `config.rs:312` — the operator this PR exists for cannot set the key

`kad_protocols` was not added to `set_key`, so `kwaainet config set kad_protocols …` errors with "Unknown config key", and `docs/BOOTSTRAP.md`'s config table (where `max_connections` and friends live) does not mention it. The documented path to the fix is missing.

### 5. `core/patches/README.md:63` — stale command

"Fresh checkout" still names `fetch-multistream-select.sh`, contradicting line 7 and leaving the workspace unparseable if followed. Same stale command at `tests/kwaai-network/stress/README-metro.md:38, :93, :100`.

RezaRassool (commented):
## Second pass (deeper)

Max-effort pass over the same head, with the patch actually exercised this time: `core/patches/fetch-patches.sh` runs clean and both patches apply; `cargo metadata --locked` fully resolves with `libp2p-kad 0.48.0` pointing at the patched path and no lockfile drift; the patch context matches upstream 0.48.0, `StreamProtocol` is already imported at `behaviour.rs:41`, `ProtocolConfig`'s `UpgradeInfo` yields every name, and `handler.rs:806` intersects the remote's protocols against the whole list. **Multi-name negotiation genuinely works**, and the missing `exclude = ["patches/libp2p-kad"]` is confirmed a non-issue.

Two new findings, and two of the earlier ones are worse than I reported.

### New: the invalid-entry fallback re-enables the protocol this PR exists to remove

`config.rs:348` — `kad_stream_protocols()` drops unparseable entries, and if none survive it falls back to the **dual default**.

The reachable path is a typo. An operator writes:

```yaml
kad_protocols: ["kwaai/kad/1.0.0"]     # missing the leading slash
```

`StreamProtocol::try_from_owned` rejects exactly that. The entry is dropped, the list is empty, the fallback fires, and the node comes back serving `/ipfs/kad/1.0.0` — so it keeps being absorbed by the public IPFS DHT and OOM-killed, which is the incident being remediated. The only signal is a `tracing::warn!` on a bootstrap nobody is tailing.

"Never a kad that can talk to nobody" is the right instinct for an ordinary node. For a **non-empty but entirely invalid** list, on the one node type this feature targets, failing startup — or at minimum `error!` — seems safer than silently restoring the thing the operator was trying to remove.

Note the new test pins the lenient behaviour, and only covers the partial case: `invalid_kad_protocol_entries_are_dropped_not_fatal` pairs `"no-leading-slash"` with a **valid** entry, so the all-invalid path is untested.

This also chains with the `set_key` gap below — with no supported way to set the key, hand-edited YAML is the only route, which is what makes the typo reachable at all.

### Worse than I reported: the nix break is not riscv-only, and the obvious fix does not work

`nix/crane.nix:62` materializes only `patches/multistream-select` (lines 66–67), so the gitignored `patches/libp2p-kad/` is absent from the flake source while `core/Cargo.toml` now references it. Cargo cannot parse the manifest.

That breaks `packages.kwaainet`, **all four `nix/cross.nix` targets**, the nix container tests, and `release-riscv.yml:43` — not just riscv as I first said.

The important part: **adding `fetch-patches.sh` to the workflow does not fix it.** `release-riscv.yml:31` already runs it, and it does not help, because a flake build reads the git tree, not the gitignored working copy. Fixing it needs a second `runCommand` (same pinned sha256 `13d3fd6…`, `libp2p-kad.patch`) plus a matching `cp -r` at line 67. Worth calling out because the natural "fix" leaves it just as broken, and no PR-level job runs nix, so CI stays green either way.

### Worse than I reported: the RTT cost doubles against legacy peers

multistream-select takes the lazy path only on the *last* proposal (`patches/multistream-select/src/dialer_select.rs:136`, the `protocols.peek().is_some()` branch → eager `FlushProtocol`/`AwaitProtocol`). Before this PR kad offered exactly one name, so every kad substream was 0-RTT under the `V1Lazy` override at `service.rs:328`.

With the two-name default, `/kwaai/kad/1.0.0` negotiates eagerly — and **against a legacy-only peer (the p2pd bootstraps) it pays a full refusal round trip before falling back**, so it is +2 RTT there, not +1.

Also: the `- kad:` bullet in the SCOPE comment at `service.rs:310–318` reasons explicitly on the assumption that kad offers a single protocol name. That invariant no longer holds and the comment now describes only the fallback case.

### Unchanged from the first pass

`core/patches/README.md:63` (plus `nix/crane.nix:26` and `core/Cargo.toml:19` comments) still name `fetch-multistream-select.sh`; following the documented fresh-checkout steps leaves the workspace unparseable. Same stale command in `tests/kwaai-network/stress/README-metro.md:38, :93, :100` — which matters more than usual, since that runbook is executed by hand on metro-win/metro-linux and there is no SSH to correct it remotely.

🤖 Second-pass review generated with [Claude Code](https://claude.com/claude-code)

RezaRassool:
## Confirming the patch-fetch doc bug on Windows

Tested via the merged `test/integration-163-167-168` tree (`bf46c072`) on metro-win (Win 11 Pro Workstations, x86_64-pc-windows-msvc).

`fetch-patches.sh` works as intended:

```
$ bash core/patches/fetch-patches.sh
fetching libp2p-kad 0.48.0 and applying the multi-protocol-names patch...
patched libp2p-kad ready at core/patches/libp2p-kad
$ ls -d core/patches/*/
core/patches/libp2p-kad/  core/patches/multistream-select/
```

The full workspace then builds and tests cleanly on Windows (1248 passed, 2 failed — both pre-existing shutdown-probe failures belonging to #163, unrelated to this PR).

**The documentation bug reproduces.** There are now three scripts in `core/patches/`:

```
fetch-libp2p-kad.sh
fetch-multistream-select.sh
fetch-patches.sh
```

and `core/patches/README.md` still directs you to the old `fetch-multistream-select.sh`. The failure mode is worse than a plain stale doc: that script **succeeds** and exits 0, leaving `patches/libp2p-kad/` absent. The error only appears later as a workspace parse failure, before anything compiles, with nothing pointing back at the patch step. Someone following the README hits a confusing failure that looks like a broken checkout.

Worth having the README name `fetch-patches.sh` as the single entry point, and ideally having the individual scripts either not be independently runnable or print a note that they're partial — otherwise the same trap stays available.

Also confirms the kad patch fetch itself works on Windows/Git Bash, which was the other thing that needed real-hardware verification.

dazwin:
## Addressed — `5850e23d`

Both passes, except finding 1, which needs a decision rather than a patch — that is at the bottom.

### The invalid-entry fallback (second pass)

You are right, and it is the worst of the set: the one node type this feature exists for, one missing slash, and it comes back serving the protocol the operator was trying to drop, behind a `warn!` nobody tails.

An empty list still means the default — that is the absent-key contract and I did not want to change it. A **non-empty list with no usable entry is now an error**, surfaced from `NetworkService::spawn`: `kad_stream_protocols()` returns `Result`, `KwaaiBehaviour::new` propagates it, and `SwarmBuilder::with_behaviour` takes a `Result`-returning constructor, so it fails at startup rather than inside the spawned task.

One correction to that finding: the all-invalid path **was** covered — `invalid_kad_protocol_entries_are_dropped_not_fatal` ended with an `all_bad` case asserting the fallback. It pinned the wrong behaviour rather than leaving it untested, which is arguably worse, and it is now two tests: the partial-drop case, and `an_entirely_invalid_kad_protocol_list_is_fatal` using your exact `"kwaai/kad/1.0.0"` as the input.

### 4 — `set_key`, and how the typo became reachable at all

`kwaainet config set kad_protocols /kwaai/kad/1.0.0` works now (comma-separated for more than one), and rejects a missing slash at the point of entry, so the common route no longer produces a config the node will refuse. A rejected value is not half-applied. Documented in `docs/BOOTSTRAP.md`'s table with the thing that is not obvious from the code: **the cutover is one-way for un-upgraded peers** — a legacy-only node cannot reach a kwaai-only bootstrap, so the legacy entry comes out after the fleet moves, not before.

### 3 — nix, including the part where the obvious fix does not work

Confirmed on both counts, and thank you for chasing the second one — I would have added `fetch-patches.sh` to `release-riscv.yml` and called it done. `nix/crane.nix` now materializes `libp2p-kad` the same way it does `multistream-select`: same pinned `13d3fd6…`, `libp2p-kad.patch`, plus the `cp -r` at the source derivation. The comment above it now says why the scripts cannot be used there — a flake build reads the git tree and their output is gitignored — since that is the trap for whoever adds the third patched crate.

### 2 — the RTT cost

Not disputed, and correct that it doubles against legacy peers. Written down rather than worked around, in two places: `NetworkConfig::kad_protocols` (+1 RTT per kad substream against an upgraded peer, +2 against a legacy-only one, per DHT hop) and the `- kad:` bullet in the `SCOPE` comment at `service.rs`, whose single-name assumption you were right to flag as no longer holding.

I have not tried to remove the cost. Reordering defeats the purpose — the remote takes the first name it matches, so legacy-first means legacy always wins — and the cost is bounded by the migration window: a node configured with one name is back on the 0-RTT path, which is what a bootstrap wants anyway. If it turns out to matter on real hops rather than on paper, gating laziness on identify's known-protocol set (the follow-up the same comment already tracks) fixes it for kad and ping together.

### 5 — stale commands, and the trap under them

Fixed in `core/patches/README.md`, `core/Cargo.toml`, and `tests/kwaai-network/stress/README-metro.md` — the runbook one first, given it is executed by hand and there is no SSH to correct it remotely.

Your Windows comment identified the sharper problem: `fetch-multistream-select.sh` **succeeds and exits 0** while leaving the workspace unparseable, so following the README produces something that reads as a broken checkout. Taking the suggestion: each per-crate script now prints a note when run outside `fetch-patches.sh`, which sets an env var to suppress it. Kept them independently runnable — during a version bump you want exactly one of them — but they now say they are only part of the job.

### 1 — `cargo publish` — needs a decision, not a patch

Reproduced exactly as you described:

```
error[E0599]: no method named `set_protocol_names` found for struct `libp2p::libp2p_kad::Config`
error: failed to verify package tarball
```

There is no fix that keeps all three of: the multi-name call, a publishable `kwaai-p2p`, and a published crate that behaves like the one we run. `[patch.crates-io]` is workspace-root-only, so nothing in a packaged crate can carry it; a Cargo feature does not work either, because any published dependent that enables it fails to verify for the same reason. The realistic options are (a) stop publishing `kwaai-p2p` and everything downstream of it while the patch is in force, or (b) publish the patched kad as our own crate and depend on it by version instead of by path — which also fixes external consumers, at the cost of the "nothing here is vendored" policy in `patches/README.md`. That is @dazwin's call and I have not pre-empted it; flagging that it blocks the next release either way, which is the part that matters.

`cargo fmt --all --check`, `cargo clippy --all-targets -- -D warnings`, `cargo test -p kwaai-p2p` (100 + integration targets) and `cargo test -p kwaainet` (220) all green.

Separately: the tree you ran on metro-win, `test/integration-163-167-168` @ `bf46c072`, predates `0cfb47db` on #163 — the two `grpc_server` failures are the defect you measured, not a live one. Rebuilt from the three current heads; detail in the #163 thread.

RezaRassool (commented):
Reviewed against a local materialization of the patched crate, with the build, tests, and a publish dry-run run on this branch.

First, the patch pipeline itself is sound. I downloaded upstream `libp2p-kad` 0.48.0 from crates.io, applied `libp2p-kad.patch`, and diffed the result against a working-tree copy produced by `fetch-libp2p-kad.sh`: **identical**. The tarball sha256 matches the pinned value in the script, and the patch applies clean at `-p1`.

One thing worth recording so it doesn't look like tampering later: on Windows the `.kwaai-patch-stamp` hash does not match the patch file's git blob hash, because `text: auto` gives the working-tree copy CRLF endings and the stamp hashes the file on disk. `sed 's/$/\r/' libp2p-kad.patch | sha256sum` reproduces the stamp exactly. Existing `multistream-select` behaves the same way, and idempotence still holds per machine, so this is a note rather than a request.

## Blocker: `cargo publish -p kwaai-p2p` fails on this branch

`cargo publish -p kwaai-p2p --dry-run` on this branch:

```
error[E0599]: no method named `set_protocol_names` found for struct
              `libp2p::libp2p_kad::Config` in the current scope
   --> src/behaviour.rs:149:20
error: failed to verify package tarball
```

`[patch.crates-io]` is workspace-local, and publish verification builds the packaged crate outside the workspace, so it resolves stock `libp2p-kad` 0.48 — which has no `set_protocol_names`.

This is qualitatively different from the `multistream-select` precedent, and I think that is why it slipped through. That patch only relaxes a runtime validation rule, so a stock build still *compiles* and publish keeps working. This one adds an API that `kwaai-p2p`'s own source calls, so the published crate cannot build at all. `kwaai-p2p` is live on crates.io (0.6.5), and both `publish-crates.yml` and `release.yml` publish it, so the next release fails at that step — and any external consumer of `kwaai-p2p` cannot compile it either.

Options, roughly in order of cleanliness:

1. Publish the fork under its own name (`kwaai-libp2p-kad`) and depend on it as a normal dependency.
2. Set `publish = false` on `kwaai-p2p`.
3. Fall back to `Config::new(single_name)` and give up the dual-name default.

Needs settling before merge either way: the patch's "natural expiry" is a later release, but the publish break lands with this one.

Related trap while you are in there: the publish loop greps failures for `already exists on crates` and treats a match as "already uploaded, skipping". The dry run emits exactly that string as a warning whenever the version is already on the index. On a fresh version number this error fails loudly as intended, but on a re-run after a partial publish it would be silently swallowed.

## Moderate: the p2pd path accepts `kad_protocols` and ignores it

`node_native.rs` is the only place the key is wired through, so on the p2pd path it is validated, saved, and does nothing. It is documented — the doc comment and the BOOTSTRAP.md row both say "native path only" — but the incident driving this PR happened on the p2pd bootstraps, which makes the operator most likely to set this key the one it will not protect, with no signal at startup. A `warn!` when `kad_protocols` is non-empty and `native_p2p` is false would close it.

## Nits

- `core/Cargo.toml`: `exclude` still lists only `patches/multistream-select`, while the comment above it now reads "Generated by patches/fetch-patches.sh". `cargo metadata` confirms `libp2p-kad` is *not* auto-adopted as a workspace member, so nothing breaks — which also means the existing `multistream-select` exclude is not doing anything. Worth either adding both entries or fixing the comment.
- `tests/kwaai-network/stress/FINDINGS.md`: "`core/patches/fetch-patches.sh` (then `fetch-multistream-select.sh`) is bash-only" does not parse.

## What I verified holds up

- Tests pass locally: the four `kad_protocols` unit tests, and `kad_negotiates_across_the_protocol_migration` — the three-node legacy↔dual↔kwaai walk, which is the load-bearing claim. `cargo check --all-targets -p kwaai-p2p` is clean.
- The V1Lazy RTT accounting is exactly right. `dialer_select.rs` gates the 0-RTT shortcut on `this.protocols.peek().is_some()` — last protocol offered only. So "+1 RTT per kad substream against an upgraded peer, +2 against a legacy-only one" is correct, and it is a per-DHT-hop cost the dual default imposes fleet-wide. Documenting that in the config docs and at the `service.rs` laziness comment, rather than leaving it to be discovered, is the right call.
- The README's claim that kad offers every name inbound as well as outbound checks out: `UpgradeInfo::protocol_info` returns all of `protocol_names`.
- All eleven workflows, `setup.sh`, and `nix/crane.nix` are updated consistently, and crane.nix reuses the same pinned sha256. Calling out in `patches/README.md` that Nix has to be updated alongside is a good addition.
- The all-invalid-list-is-fatal change is the right direction: a missing leading slash on a bootstrap would otherwise restore `/ipfs/kad/1.0.0` behind a `warn!` nobody is tailing.
- The BOOTSTRAP.md row is the right way round — columns are `key | bootstrap value | default`, so kwaai-only as the bootstrap value and dual as the default is correct.

🤖 Review assisted by [Claude Code](https://claude.com/claude-code)

dazwin:
## Combined validation on Windows — `c0e93746`

The rebuilt integration branch (`e307da64` + `5850e23d` + `0b042bc8`) is green on Windows: **433 passed, 0 failed, 5 ignored** across 19 suites, exit 0. This PR's `kad_protocols` validation rides in the `kwaai-p2p` lib tests; @RezaRassool separately confirmed `fetch-patches.sh` and the new partial-fetch note on Git Bash. Detail in the [#163 thread](https://github.com/Kwaai-AI-Lab/KwaaiNet/pull/163#issuecomment-5500218559).

Finding 1 (`cargo publish`) is still the open item and is being resolved separately — an update on that shortly.

RezaRassool (commented):
## Third pass — the redesign, reviewed on its own terms

My earlier two reviews were of the config-based design, which no longer exists. Re-reviewed
`1bf6d34a` from scratch.

**First, what the redesign got right.** The config key is genuinely gone —
`NetworkConfig::kad_protocols` survives only as `#[serde(skip)]` with a comment stating
plainly that operators cannot reach it. That closes the silent-fallback finding properly
rather than patching around it. `cargo metadata --locked` is consistent, `patches/libp2p-kad`
is not auto-adopted, and the branch still merges cleanly onto `main` despite being 7 commits
and three releases behind.

Two HIGH findings, and I have to withdraw something I said in this thread.

### 1. `tests/swarm.rs:226` — the test justifying the libp2p-kad patch is vacuous

I ran this test earlier, saw it pass, and reported it as sound. **Passing proved nothing.**

`NetworkService::dial()` (`service.rs:903`) calls `kad.add_address()` for every dialled peer
regardless of kad protocol support, and `Command::DhtFindPeer` (`service.rs:657`)
short-circuits whenever `known_addresses()` is non-empty — its own comment says so: *"if we
already have addresses (routing table or a live connection) there is no need for a network
walk."* So `b.dht_find_peer(c_id)` is answered from the live B→C connection without any kad
substream ever opening.

Proved by making B **legacy-only**, so B and C share zero kad protocols — the exact condition
the doc comment says must fail:

```
test kad_negotiates_across_the_protocol_migration ... ok
test result: ok. 1 passed  finished in 0.10s
```

It passes. And in 0.10s against ~10s for the honest path, which is the short-circuit visible
in the timing.

So the only test of the three-role migration topology would pass with `set_protocol_names`
deleted — with the patch that is the entire reason for forking libp2p-kad. To exercise real
negotiation the probe has to force a kad walk: seed B's table via `kad_add_address` without
dialling C from B, or assert on `b.kad_peers()`, or use a lookup that cannot short-circuit.

### 2. `config.rs:251` — every shipped binary is `/kwaai/kad/1.0.0`-only

With the feature off — the default, and what every workflow and cargo-dist build produces —
`kad_protocols()` returns `[KWAAI_KAD_PROTOCOL]` alone. The production bootstraps advertise
`/ipfs/kad/1.0.0` (recorded as measured in `live_bootstrap.rs:126`), as does every deployed
node. An upgraded node's kad substreams to them fail negotiation outright.

This is the failure the comment deleted from `behaviour.rs:8-11` warned about: *"any custom
protocol name here silently partitions us from the live network."*

The rollout order this implies — **bootstraps dual before any node upgrades** — is what the
Slack plan does (build bootstraps with the flag, *then* merge and release). But the PR
description still describes the replaced design, calling the node default dual and this a
"no-op compatibility release". Anyone following the PR rather than the plan inverts the order
and partitions the fleet. The description needs rewriting before this merges.

### 3. `behaviour.rs:158` — the bootstrap-grade build is compiled by nothing

`kad-multi-protocol` appears in no workflow and nothing in `nix/crane.nix` or cargo-dist
enables it, so `clippy --all-targets -D warnings` and every test job typecheck the crate
*without* the patched API. A libp2p bump that changes the patched signature breaks only the
hand-built bootstrap, discovered at deploy time. It also means `release.yml` never produces
the artifact `docs/BOOTSTRAP.md:53` tells the operator to deploy.

One job fixes it — `cargo test -p kwaai-p2p --features kad-multi-protocol` compiles and
passes today (16 tests, verified). Though note finding 1: it would pass either way.

### 4. `behaviour.rs:157` — `protocols[0]` panics on an empty list

`NetworkConfig` is public in a published crate, so
`NetworkConfig { kad_protocols: vec![], ..Default::default() }` aborts `KwaaiBehaviour::new`
with index-out-of-bounds. The `debug_assert!` above checks `len() > 1`, never zero.
`protocols.first().cloned().unwrap_or(KWAAI_KAD_PROTOCOL)` would be an error or a fallback
instead of a panic.

### 5. `behaviour.rs:148` — the startup log reports protocols the binary does not serve

In a release build without the feature, a two-entry list takes
`kad::Config::new(protocols[0])` while `set_protocol_names` is `#[cfg]`'d away — so one name
is served. But `tracing::info!(protocols = ?…)` logs **both**, and the `debug_assert!`
guarding the mismatch is compiled out in release.

That is the line `docs/BOOTSTRAP.md` tells the operator to check to confirm which protocols
the node serves. It can report a set the binary does not serve, in exactly the "silently
wrong" mode the design note at `config.rs:226` says a build flag eliminates. Log the names
actually installed, or make the mismatch a hard error.

### 6. Three in-tree docs assert a dual default no shipped build has

- `behaviour.rs:8` — "by default on both `/kwaai/kad/1.0.0` and the legacy `/ipfs/kad/1.0.0`"
- `core/patches/README.md:41` — "nodes serve and offer … *and* the legacy"
- `service.rs:322` — "the default list is `[/kwaai/kad/1.0.0, /ipfs/kad/1.0.0]`"

All three also point at `NetworkConfig::kad_protocols` as the knob, which `config.rs:139`
documents as unreachable. These are the comments a maintainer reads to answer "am I
partitioned from the bootstraps?" and they answer it wrongly. The `service.rs` V1Lazy RTT
paragraph is likewise premised on two names — with the real default, the 0-RTT path is
retained and that cost does not exist.

### 7. LOW — the feature is published to crates.io but uncompilable there

`publish-crates.yml` publishes `kwaai-p2p`; `[patch.crates-io]` does not travel with a
package, so an external consumer enabling `kwaai-p2p/kad-multi-protocol` gets
`no method named set_protocol_names`. Publishing succeeds because verification builds default
features only. A `compile_error!` guard pointing at `core/patches/README.md` would fail at the
right moment with the right message.

### 8. LOW — `exclude` not extended to the new patched crate

`core/Cargo.toml:20` still reads `exclude = ["patches/multistream-select"]` while the comment
above explains why patched path crates are excluded. Confirmed **not** a live bug —
`cargo metadata` lists 14 members, none of them libp2p-kad — but the invariant now has one
entry and two patched crates.

🤖 Third-pass review generated with [Claude Code](https://claude.com/claude-code)

core/crates/kwaai-p2p/src/config.rs
```diff
@@ -194,6 +215,54 @@ fn default_identify_min_confirmations() -> usize {
...
+/// Retire the feature, and the patched libp2p-kad with it, once the fleet has
+/// moved.
+///
+/// **Two names cost a round trip.** `V1Lazy` takes the 0-RTT shortcut only on
+/// the last protocol offered, so the preferred name negotiates eagerly: +1
+/// RTT per kad substream against an upgraded peer, +2 against a legacy-only
+/// one, which refuses the kwaai name before falling back. kad opens a
+/// substream per request, so that is per DHT hop — another reason this
+/// belongs to the migration window rather than to steady state.
+pub fn kad_protocols() -> Vec<StreamProtocol> {
```
RezaRassool, line 251: **Blocker — the shipped default makes this a flag day, not a no-op release.**

`kad-multi-protocol` is off by default (`Cargo.toml:44`), so this returns `["/kwaai/kad/1.0.0"]` alone for every stock build. `DEFAULT_NATIVE_P2P = true` since v0.6.0 means every deployed v0.6.x node runs this kad on `/ipfs/kad/1.0.0` — an upgraded node shares no kad protocol with any of them, nor with `native_p2p: false` p2pd nodes, nor with the current Python `run_dht` bootstraps.

Concretely: after upgrading, `dht_find_peer` / `known_peers` (which feeds `DHTStorage::update_peer_ids`, and therefore announce and shard placement) works only through a bootstrap someone hand-built with `--features kad-multi-protocol`. And because `kad_protocols` is `#[serde(skip)]`, there is no runtime override — recovering from a partition means rebuilding and redeploying binaries, to machines we have no SSH to.

The reasoning in this doc comment for *why* it's a build flag rather than config is genuinely good, and I'd keep it. The issue is only which way the default points. Shipping the dual list as the node default for one release — what commit 1 did — removes the flag day at the cost of the documented +1 RTT, and the build flag then becomes what it should be: the bootstrap-grade hardening that guarantees a public node can never be absorbed.

core/crates/kwaai-p2p/src/behaviour.rs
```diff
@@ -5,10 +5,13 @@
 //! - [`ping`] — liveness / RTT, and it keeps otherwise-idle connections honest.
 //! - [`identify`] — protocol/agent advertisement plus the **observed address**
 //!   feed that later phases use for reachability detection.
-//! - [`kad`] — Kademlia peer routing on the *default* `/ipfs/kad/1.0.0`
-//!   protocol. This is deliberate: the Python bootstraps run hivemind's
-//!   go-libp2p daemon with no `ProtocolPrefix`, so any custom protocol name
-//!   here silently partitions us from the live network.
+//! - [`kad`] — Kademlia peer routing, by default on both `/kwaai/kad/1.0.0`
```
RezaRassool: This module doc is inverted relative to the code below it. It says kad runs "by default on both `/kwaai/kad/1.0.0` and the legacy `/ipfs/kad/1.0.0`" and that "bootstrap-grade deployments configure the kwaai name alone" — but after commit 3 the shipped default is kwaai-only, and the bootstrap-grade *build* is the one that serves both. The two halves are swapped.

It also points readers at `NetworkConfig::kad_protocols` as the knob, which is now `serde(skip)` and effectively test-only.

This is the first place someone looks to answer "what does this binary serve?", and right now it gives them the opposite answer.

core/patches/README.md
```diff
@@ -1,9 +1,60 @@
...
+
+Kad's negotiation machinery holds a `Vec<StreamProtocol>` and offers every
+entry on both inbound and outbound streams, but upstream removed the public
+`set_protocol_names` setter, leaving `Config::new` with a single name. The
+patch restores the setter on `Config` (and adds one on `ProtocolConfig`,
+where the field lives) — no behavioral change to negotiation itself.
+
+We need it for the kad protocol migration: nodes serve and offer
+`/kwaai/kad/1.0.0` *and* the legacy `/ipfs/kad/1.0.0`
+(`NetworkConfig::kad_protocols`), so upgraded peers negotiate the kwaai name
```
RezaRassool: Same inversion as the `behaviour.rs` module doc, and it matters more here because this is the file the next maintainer reads before touching the patch: "nodes serve and offer `/kwaai/kad/1.0.0` *and* the legacy `/ipfs/kad/1.0.0` (`NetworkConfig::kad_protocols`) … so bootstrap-grade nodes configure the kwaai name alone."

An operator who follows this builds a bootstrap with default features and thereby cuts off exactly the legacy peers the bootstrap exists to bridge — the failure is silent, and it's the one machine where it does the most damage.

`docs/BOOTSTRAP.md` has it right; this file and the `behaviour.rs` header both contradict it.

core/crates/kwaai-p2p/Cargo.toml
```diff
@@ -43,3 +43,9 @@ sha1 = "0.10"                     # DHTID derivation in the live DHT test
 [features]
 default = ["tcp"]
 tcp = []
+# Serve /ipfs/kad/1.0.0 alongside /kwaai/kad/1.0.0 — the bootstrap-grade build
+# for the migration window. Needs the patched libp2p-kad in [patch.crates-io]
+# (patches/libp2p-kad.patch restores `set_protocol_names`, removed upstream in
+# 0.47), so it is OFF by default: the published crate carries no patch and must
+# still compile. See `kad_protocols()` in src/config.rs.
+kad-multi-protocol = []
```
RezaRassool, line 51: No CI job enables this feature. `ci.yml:112` runs `cargo clippy --all-targets` with default features only (the comment at `ci.yml:11` records that `--all-features` was deliberately dropped for CUDA/Metal reasons), and `ci.yml:185` runs `cargo test --all` without the feature name. Nothing in any workflow passes it.

So the `#[cfg]`-gated call to the patched `set_protocol_names` (`behaviour.rs:158`) and the whole `kad_negotiates_across_the_protocol_migration` test never compile, let alone run, in CI. The one build that must not break — the bootstrap artifact — is the only one we don't test.

Both work today; I verified locally. The risk is the next libp2p bump silently invalidating `libp2p-kad.patch` or this call site, and us discovering it while hand-building a bootstrap under outage pressure. One step fixes it:

```yaml
- run: cargo test -p kwaai-p2p --features kad-multi-protocol
```

core/crates/kwaai-p2p/src/service.rs
```diff
@@ -317,6 +317,21 @@ impl NetworkService {
                             //   peers only from identify plus a live FIND_NODE
                             //   probe.
                             //
+                            //   That reasoning assumed kad offers a single
```
RezaRassool: This comment documents a protection that isn't in effect under the shipped default.

It argues the earlier "kad admits unconfirmed peers" escape hatch is now closed because offering multiple kad names forces eager negotiation, making refusals visible. But with the kwaai-only default there's a single name again: V1Lazy takes the 0-RTT shortcut, `on_fully_negotiated_outbound` fires before the remote has confirmed anything, and with `BucketInserts::OnConnected` every legacy peer gets inserted as a kad server.

So the original caveat ("latent today — every fleet peer serves kad") isn't just still true, it's precisely what this PR invalidates: during the migration the routing table fills with peers the node cannot query, and each query burns a hop and a timeout on them. That's the opposite of what the comment claims.

Restoring the dual default would make the comment accurate as written.

docs/BOOTSTRAP.md
```diff
@@ -38,6 +38,32 @@ Without these, the process is an ordinary node, not a bootstrap.
 | `no_relay` | `false` | `false` | Keep the circuit relay hop service **on**: NATed nodes reserve circuits through the bootstraps. Set `true` only to deliberately withhold relay. |
 | `max_connections` | sized to the host | `100` | The ceiling on established connections, and the only bound on how far the swarm's memory grows — idle connections are held for ten minutes. 100 suits a node that mostly dials out; a bootstrap mostly receives, and wants several hundred to low thousands. Budget roughly a megabyte per connection. Inbound is capped at 4/5 of this so the node can always still dial. |
 
+### The kad protocol set is a build flag, not a key
```
RezaRassool, line 41: This section is correct about the build flag (unlike `patches/README.md`), but the rollout it implies has a gap: it depends on a `--features kad-multi-protocol` bootstrap, and the release pipeline builds no such artifact — while `contribute.auto_update` defaults to `true` (`kwaai-cli/src/config.rs:465`).

So a hand-built bootstrap gets replaced by the stock kwaai-only release binary at its next auto-update, silently ending legacy bridging partway through the migration. Same class as the metro-win update bug: the machine keeps running, nothing errors, and the degradation only shows up as peers quietly failing to resolve.

Either tell the operator to set `auto_update: false` here (and say why), or have the release build the multi-protocol artifact so an update keeps the capability.

core/crates/kwaai-p2p/src/behaviour.rs
```diff
@@ -138,9 +141,22 @@ impl KwaaiBehaviour {
...
+        debug_assert!(
+            protocols.len() <= 1 || KAD_MULTI_PROTOCOL_BUILD,
+            "more than one kad protocol on a build that cannot serve them"
+        );
+        tracing::info!(
+            protocols = ?protocols.iter().map(|p| p.to_string()).collect::<Vec<_>>(),
+            multi_protocol_build = KAD_MULTI_PROTOCOL_BUILD,
+            "kad protocol set"
+        );
+        let mut kad_config = kad::Config::new(protocols[0].clone());
```
RezaRassool: Minor: `protocols[0]` indexes without a length check.

`kad_protocols` is a `pub Vec`, and the surrounding code is careful about degenerate lists everywhere else — the patch itself deliberately treats an empty list as "keep what you have" (`set_protocol_names` ignores empty), and commit 2 made an all-invalid list fall back to the default. This line has neither guard, so a test or helper passing `&[]` (say `spawn_swarm_with_kad_protocols(&[])`) panics with an opaque index-out-of-bounds from inside the swarm builder.

```rust
let mut kad_config = kad::Config::new(
    protocols.first().cloned().unwrap_or_else(|| kad_protocols()[0].clone()),
);
```

keeps the invariant local instead of relying on callers.

RezaRassool (commented):
## Review

The mechanics here are solid. I applied both patches via `fetch-patches.sh`, confirmed with `cargo metadata` that neither patched crate is auto-adopted as a workspace member (so the missing `exclude` entry for `patches/libp2p-kad` is harmless), read the patched `libp2p-kad` source — `Config::new` is behaviourally identical to the removed `Config::default()`, and `ProtocolConfig::protocol_names` really is the single source for both inbound and outbound `UpgradeInfo`, so the patch is sound — and ran `cargo test -p kwaai-p2p --features kad-multi-protocol --test swarm`: 16/16 pass, including the new migration test. The IPFS-absorption diagnosis is well-evidenced and the build-flag rationale in `kad_protocols()` is convincing.

The problems are all in what the **default** build does, plus docs that still describe the design commit 1 shipped and commit 3 replaced.

**One blocker:** because `kad-multi-protocol` is off by default, a stock build serves `/kwaai/kad/1.0.0` alone — which makes this a flag day, not the "no-op compatibility release" the description promises. `DEFAULT_NATIVE_P2P` has been `true` since v0.6.0, so an upgraded node shares no kad protocol with any deployed v0.6.x node, with `native_p2p: false` p2pd nodes, or with the current Python `run_dht` bootstraps. `dht_find_peer` / `known_peers` — which feeds `DHTStorage::update_peer_ids` and hence announce and shard placement — would then work only through a bootstrap someone hand-built with the feature on. And since `kad_protocols` is `#[serde(skip)]`, there is no runtime override: recovering from a partition means rebuilding and redeploying binaries.

Commit 1's dual-list default removed that risk entirely, at the cost of the documented +1 RTT. I'd take that trade for one release and keep the build flag as the bootstrap-hardening knob on top.

Details inline. Ranked: 1 blocker, 5 worth fixing before merge, 1 minor.

dazwin:
## Addressed — `fb956457`, and the blocker answered with deployment state

Both passes accepted in full except the blocker's *remedy*, where there's a constraint you didn't have visibility into. Everything else is on the branch.

### The vacuous test — you were right, and it's now unfakeable

Your `DhtFindPeer` short-circuit diagnosis was exact: the probe resolved a peer its caller had already dialled, and `dial()` seeds the routing table. The probes now target peers the caller has **never dialled** — B resolves D, which only C knows, forcing a real FIND_NODE round trip on `/kwaai/kad/1.0.0`; A resolves D end-to-end, forcing the legacy leg too. And your negative control is now a permanent test: `a_legacy_only_bridge_cannot_resolve_across_the_protocol_boundary` runs your exact falsification (legacy-only B) and asserts the walk **fails** — if it ever yields addresses, the short-circuit is back and the suite says so.

### Accepted outright

- **CI compiles the bootstrap build**: `cargo test -p kwaai-p2p --features kad-multi-protocol` is a CI step now — your point that the one artifact that must not break was the only one nothing compiled is the whole justification, verbatim in the step comment.
- **`behaviour.rs` module doc and `patches/README.md`**: both rewritten. They described the config design this branch replaced, inverted — and your observation that the README would lead an operator to build the exact bootstrap that cuts off the peers it exists to bridge is now the text's own warning.
- **`service.rs` V1Lazy comment**: now states plainly that the single-name default reopens the kad-admits-unconfirmed caveat (follow-ups: #174 and identify-gated laziness), and that the dual build keeps refusals eager at +1 RTT — accepted for the window on the hosts that run it.
- **`protocols[0]`**: your `first().cloned().unwrap_or_else(...)` suggestion, as written, keeping the invariant local and matching the patched setter's own empty-list behaviour.
- **`auto_update` on a dual bootstrap**: BOOTSTRAP.md now instructs disabling it and says why. The production units already run `KWAAINET_NO_AUTO_UPDATE=1`; the doc catches everyone else. (Your "same class as the metro-win update bug" framing was apt — we hit exactly this designing the fleet units.)
- **PR description**: rewritten around the build-flag design, with the rollout order marked **normative** and the inversion hazard stated.

### The blocker — holding the kwaai-only default, for a structural reason

Restoring commit 1's dual default for one release isn't a smaller-risk variant of this design — it's the design this branch had to abandon: **a dual default puts the patched `set_protocol_names` call in every stock build, and `[patch.crates-io]` cannot travel into a published crate.** That's the `cargo publish` failure you found in your first review (`failed to verify package tarball`, taking the eight downstream crates with it). Feature-off-by-default is what fixed it; dual-by-default un-fixes it. The triangle — dual default, publishable `kwaai-p2p`, no vendored fork — only ever lets us pick two.

What makes the flag day acceptable is the deployment state, which your third pass anticipated: the dual bootstraps are **live in production now** — deployed 2026-09-01, serving since the `_dnsaddr` cutover on 2026-09-02, `multi_protocol_build=true` in their startup logs, self-update disabled, and currently bridging the actual fleet under a full Amino inbound flood (~100 KB/connection, zero restarts; measurements in #174). Step 1 of the normative order is already done, which means the release containing this PR lands on a network where the bridge precedes it. The remaining exposure is node↔node legacy kad during each node's own update window — bounded by one auto-update wave at an 11-node fleet, with records replicated on the bootstraps.

So: docs and description now make the order impossible to miss, CI guards the bridge build, and the default stays kwaai-only because the alternative re-breaks the release pipeline. If the fleet were 500 nodes rather than 11 I'd weigh the same triangle differently — and #174 is where that future gets its connection-level answer.

core/crates/kwaai-p2p/src/config.rs
RezaRassool: **Retracting this — I tested it against the live network and I was wrong.**

I built both variants off `1bf6d34` and ran them as isolated nodes (own identity, `announce_self: false`) against the live bootstraps. The default kwaai-only build is **not** partitioned from the fleet:

```
kwaai-only build — shard chain
  Querying 32 blocks from DHT…
    7 server(s) — 32/32 blocks covered
  ✅ Full model coverage — distributed inference ready
```

Identical to the `kad-multi-protocol` build, and it lists the real fleet (christophermayfield-mac, john-linux-draca, annikavriens, projxai, john-linux-spectre, rezarassool-macos) including nodes on 0.6.7 that serve only the legacy name.

My error: I assumed block discovery and announcement ride libp2p-kad. They don't. `decentralized_dht` defaults to `false` (`kwaai-cli/src/config.rs:903`), so a default node queries the bootstraps' **Hivemind DHT over `/kwaai/*` unary RPC** (`rpc_ping`/`rpc_store`/`rpc_find`) — protocols this PR doesn't touch. kad only feeds `routing_peers()` → `DHTStorage::update_peer_ids` (`dht_service.rs:184` → `service.rs:618`, `kad.kbuckets()`), and that maintenance loop exists for *decentralized DHT placement*, which is off by default.

So there is no flag day, and the "no-op compatibility release" framing in the description is fair for default-configured nodes. Sorry for the noise.

What I'd still note, much more weakly: for `decentralized_dht: true` nodes the kad routing table *is* what the local DHT server places records from, so those nodes see a genuinely smaller candidate set during the migration window. That's a documentation point, not a blocker — worth a sentence in `docs/BOOTSTRAP.md` so anyone running decentralized placement knows to use the multi-protocol build until the fleet is upgraded.

I also withdraw my recommendation to restore the dual list as the node default — see my top-level comment; the measurements point the other way.

RezaRassool (commented):
## Correction to my review above — I tested this on the live network

I built both variants off `1bf6d34` and ran them as isolated nodes against the live bootstraps. Two of my findings don't survive contact with reality, and one new thing showed up. Details on the individual threads; summary here.

### 1. The blocker was wrong — retracted

The kwaai-only default does not partition from the fleet. Both builds return identical `shard chain` results: 32/32 blocks, full coverage, real fleet nodes listed, including 0.6.7 peers that serve only the legacy name.

I'd assumed block discovery rode libp2p-kad. It doesn't — `decentralized_dht` defaults to `false`, so default nodes query the bootstraps' Hivemind DHT over `/kwaai/*` unary RPC, untouched by this PR. Full reasoning on the `config.rs:251` thread. **I also withdraw the recommendation to restore the dual list as the node default.**

### 2. The premise is strongly confirmed on my own machine

Before touching the test builds, my live 0.6.7 node showed:

| metric | value |
|---|---|
| established connections | **100** (= `max_connections`) |
| on `/tcp/4001` (IPFS default) | 37 |
| bootstraps connected | 1 of 2 |
| real KwaaiNet nodes (`map.kwaai.ai` `num_peers`) | **13** |

~87 of 100 connection slots spent on foreign peers, at a ceiling low enough to crowd out a bootstrap. Same shape as your 300–600 measurement, capped by `max_connections`. This problem is real and live.

### 3. New finding: the rename does not stop connection-table absorption

This is the one thing I'd ask you to look at. I sampled both test nodes every 45s:

```
time      kwaai-only            dual
13:33:01  tot=25  4001=17       tot=100 4001=51
13:33:46  tot=24  4001=17       tot=100 4001=47
13:34:31  tot=97  4001=88       tot=100 4001=51
13:35:16  tot=92  4001=82       tot=100 4001=49
13:36:01  tot=92  4001=59       tot=100 4001=62
13:36:46  tot=100 4001=62       tot=100 4001=51
```

The dual build saturates in ~2 minutes. The kwaai-only build takes ~7 — but **it gets there too**, ending pinned at 100 with 62 foreign. (I reported an early "kwaai-only stays at 27 and prunes" reading to my own team before the series completed; the later samples killed it. Slower, not immune.)

So this claim in the description is too strong:

> a node that never serves the legacy name can never be absorbed by the public IPFS DHT

It can't be absorbed *as a DHT server* — it answers no `/ipfs/kad/1.0.0` queries, which is very plausibly what was OOM-killing the bootstraps, and that's a real win. But the connection table still fills with foreign peers, because absorption there happens below kad: TCP + noise + yamux + identify all still negotiate, and the addresses come back from the bootstraps' own (already-absorbed) routing tables. Nothing in this PR gates that.

Worth either narrowing the claim, or pairing the protocol change with connection-level gating — a lower `max_connections` alone won't help, since the slots fill with foreign peers either way and that's what starved the second bootstrap on my node. Not necessarily this PR's job, but the rollout shouldn't assume the bootstraps are fixed by the rename alone.

### 4. Unaffected findings

The patch mechanics are sound and I verified them: both patches apply via `fetch-patches.sh`, `set_protocol_names` works (`protocols=["/kwaai/kad/1.0.0", "/ipfs/kad/1.0.0"] multi_protocol_build=true` at startup), `cargo test -p kwaai-p2p --features kad-multi-protocol --test swarm` passes 16/16, and neither patched crate is auto-adopted as a workspace member.

Still stand as written: the three inverted docs (`behaviour.rs:8`, `patches/README.md:43` — these now contradict the shipped default in the *opposite* direction from what I first assumed, but they're still wrong), the missing CI feature job (`Cargo.toml:51`), the `auto_update` gap (`docs/BOOTSTRAP.md:41`), and the `protocols[0]` index (`behaviour.rs:157`).

The `service.rs:320` comment finding stands but softer — the routing-table pollution is real, and the test does show foreign peers accumulating, but with `decentralized_dht: false` by default the practical cost is bounded.

Net: no blocker from me. Fix the docs and add the CI job and I'm happy.

dazwin:
Crossed in flight — your fifth pass was written against `1bf6d34a`, and your two remaining asks were already on `fb956457` (pushed five minutes before it): the CI feature job and both doc rewrites, plus the unfakeable test rebuild from your third pass (probes now target never-dialled peers, and your legacy-only-bridge falsification is a permanent negative-control test).

Retraction acknowledged with respect — building both variants and running them against the live fleet is the review this PR needed, and "block discovery rides the Hivemind unary DHT, untouched by this PR" settles the partition question better than my deployment-state argument did.

**Your finding 3 is accepted as stated, and it's now in the tree as `3896e8a2`:** the description, `behaviour.rs` module doc, and `patches/README.md` all narrow the claim to what the rename actually ends — recruitment as an IPFS DHT *server* (the routing-table/OOM vector) — and say plainly that foreign peers still connect below kad, with the connection-table half pointed at #174. Your time-series (kwaai-only saturating in ~7 min vs dual's ~2, both pinned with majority-foreign tables) and the live node datum — 100/100 slots, 87 foreign, **one of two bootstraps crowded out** — are going onto #174 now: they move connection management from "bootstrap hardening" to "every default node at `max_connections: 100` is starving today," which materially reranks that issue.

One nuance your mechanism note sharpens: the foreign addresses "come back from the bootstraps' own (already-absorbed) routing tables" — meaning nodes aren't just *receiving* foreign dials, they're *making* them, from polluted referrals. So #174 needs the dial-side filter (don't walk into peers advertising no kwaai protocols) as well as the admission/eviction side. Credited to this pass there.

RezaRassool (commented):
## Consolidated verdict — read this one, not the sequence above

I left a review, then a retraction, then a correction on this PR. That's confusing to read in order, so here is the single conclusion.

### Not blocked. Fix two things and this should merge.

**1. Three docs describe the opposite of what ships.** Commit 3 made the default kwaai-only and moved the dual list behind `kad-multi-protocol`, but these still describe commit 1's design:

- `core/crates/kwaai-p2p/src/behaviour.rs:8` — says kad runs "by default on both" names and that bootstrap-grade deployments "configure the kwaai name alone". Both halves are backwards. It also points at `NetworkConfig::kad_protocols`, now `serde(skip)` and test-only.
- `core/patches/README.md:43` — same inversion, in the file the next maintainer reads before touching the patch. An operator following it builds a bootstrap with default features and cuts off exactly the legacy peers the bootstrap exists to bridge.
- `docs/BOOTSTRAP.md:41` is correct and contradicts both.

**2. No CI job enables `kad-multi-protocol`.** `ci.yml:112` is clippy with default features; `ci.yml:185` is `cargo test --all` without the feature. So the `#[cfg]` call to the patched `set_protocol_names` (`behaviour.rs:158`) and the whole `kad_negotiates_across_the_protocol_migration` test never compile in CI — the bootstrap build is the one build that must not break and the only one untested. One step fixes it:

```yaml
- run: cargo test -p kwaai-p2p --features kad-multi-protocol
```

### Optional, your call

- `behaviour.rs:157` — `protocols[0]` indexes without a length check, while the patch treats an empty list as "keep what you have" and commit 2 made an all-invalid list fall back to the default. `protocols.first().cloned().unwrap_or_else(|| kad_protocols()[0].clone())` keeps the invariant local.
- `service.rs:320` — the comment argues the dual default forces eager negotiation so kad refusals become visible. Under the shipped single-name default that isn't in effect. Lower stakes than I first said (`decentralized_dht: false` bounds the cost), but it documents a protection that isn't there.

### Withdrawn: my blocker, and the dual-default recommendation

I claimed the kwaai-only default was a flag day that would partition the fleet, and pushed to restore the dual list as the node default. **Both were wrong.** I had assumed libp2p-kad was load-bearing for block discovery. It isn't — `decentralized_dht` defaults to `false`, so discovery rides the Hivemind DHT over `/kwaai/*` unary RPC, which this PR never touches. kad only feeds `routing_peers()` → `DHTStorage::update_peer_ids` (`dht_service.rs:184`), used for decentralized placement, which is off by default.

I also reported that the kwaai-only build prunes foreign IPFS connections and settles low. It doesn't — it saturates `max_connections` too, just over ~7 minutes instead of ~2. I posted that before the sample series finished.

### Positively validated on real hardware

Built both variants off `1bf6d34` and ran the exact migration case: a **kwaai-only client** driving **metro-win on 0.6.7, which serves only legacy `/ipfs/kad/1.0.0`**, under a Dream RAG load over `/kwaai/inference-mux/1.0.0`.

| run | completions | workers | elapsed | transport errors |
|-----|------------|---------|---------|------------------|
| A | 20 | 4 | 16.8s | 0 |
| B | 50 | 8 | 43.3s | 0 |
| C1-C3 | 50 each, consecutive | 8 | 44s / 7s / 6s | 0 |

~170 completions, zero 502/503, zero dropped streams, zero dial failures, no panics. Capacity leasing backpressured correctly throughout (4 granted, rest `DeniedAtCapacity`). The known concurrent-GPU p2p crash did not reproduce.

Also verified: both patches apply via `fetch-patches.sh`; the patched `Config::new` is behaviourally identical to the removed `Config::default()`; `ProtocolConfig::protocol_names` really is the single source for inbound and outbound `UpgradeInfo`; `cargo test -p kwaai-p2p --features kad-multi-protocol --test swarm` passes 16/16; neither patched crate is auto-adopted as a workspace member, so the missing `exclude` for `patches/libp2p-kad` is harmless.

The diagnosis behind this PR is sound and the build-flag rationale in `kad_protocols()` is the right call. Sorry for the noisy path to that conclusion.

### Two things that are *not* this PR's problem

Both reproduce identically on stock 0.6.7 and should not gate the merge:

- **#179** — nodes announce block coverage while publishing no dialable address. This is why I had to dial metro-win explicitly to run the test at all.
- **QUIC transport mismatch** — the only relay circuit addresses on the fleet are `quic-v1`/`webrtc-direct`, while `enable_quic` defaults to `false` since #162. A node holds a circuit reservation it cannot itself dial out over. Details in #179.

### One thing this PR's rollout should still address

`docs/BOOTSTRAP.md` depends on a hand-built `--features kad-multi-protocol` bootstrap, but the release pipeline builds no such artifact and `contribute.auto_update` defaults to `true` (`kwaai-cli/src/config.rs:465`). I hit this for real: a test node of mine with the default silently downloaded 0.6.7, overwrote the binary, respawned, and stopped — replacing the build under test. A hand-built bootstrap will do the same mid-migration. Either tell the operator to set `auto_update: false`, or ship the artifact.

dazwin:
Every item in the consolidated verdict is already on the branch — the review and the pushes keep crossing; ledger against current head `3896e8a2`:

| verdict item | status |
|---|---|
| `behaviour.rs:8` inversion | fixed in `fb956457`, claim further narrowed in `3896e8a2` (server-recruitment vs connection-fill, per your finding 3) |
| `patches/README.md:43` inversion | same two commits |
| CI job for `kad-multi-protocol` | added in `fb956457` — your exact step, with your "the one build that must not break is the only one untested" as the step comment |
| `protocols[0]` | your `first().cloned().unwrap_or_else(...)` verbatim, `fb956457` |
| `service.rs:320` comment | rewritten to state per-build truth: single-name default reopens the caveat (bounded by `decentralized_dht: false`, follow-ups #174 + identify-gated laziness); dual build keeps refusals eager |
| `auto_update` on a hand-built bootstrap | BOOTSTRAP.md now instructs `auto_update: false`/`KWAAINET_NO_AUTO_UPDATE=1` and why — the production units already run the env var, set for exactly the failure you then hit live |
| vacuous migration test (third pass) | rebuilt in `fb956457`: probes target never-dialled peers; your legacy-only falsification is the permanent negative-control test `a_legacy_only_bridge_cannot_resolve_across_the_protocol_boundary` |

The hardware validation is the piece nobody else could supply — a kwaai-only client against a stock-0.6.7 legacy node under real inference load is the exact pair the migration window creates, and ~170 completions with zero transport errors is the empirical form of the "Hivemind RPC is untouched" argument. Thank you for running it, and for the retraction discipline throughout; the noisy path was the review working.

On your two non-PR findings: **#179 converges with an independent investigation from the map side today** — 11 of 15 peers on map.kwaai.ai have *zero* announced addresses (that's the "Location Unknown" growth: not a GeoIP bug, the peers are undialable), and not one fleet peer holds a relay reservation on the kwaai bootstraps; the four locatable NATed nodes all sit on third-party IPFS relays. Your QUIC observation likely explains part of *why*: circuit reservations landing on `quic-v1` relays that an `enable_quic: false` node can't dial back out of. Cross-linking that evidence onto #179.

Merging on the operator's word — the rollout precondition (dual bootstraps live and bridging) has been in force since yesterday.

PR #168, opened 2026-08-31
fix(p2p): re-seed configured bootstraps, so a bootstrap restart cannot wedge a node

dazwin:
## The wedge

Configured bootstrap multiaddrs are dialled exactly once, at startup (`node_native.rs`'s single `handle.bootstrap(addrs)` call). Everything after that goes by peer id: the announce fan-out discards the transport address, the periodic kad refresh only walks peers still in the routing table, and identify needs a live connection. So once a bootstrap's routing entry is evicted — dial failures during its restart window, bucket replacement on a busy DHT — nothing can ever turn its peer id back into an address again.

The result is the half-alive state seen on the live fleet on 2026-08-30 when both bootstraps restarted together: nodes kept listening (TCP up, `is_reachable` fine) but stopped announcing and lost their relay reservations until each node was manually restarted. Reproduced in the sealed nat-test topology with an 8-minute all-bootstrap outage:

```
STORE RPC failed (/ip4/…/tcp/8000/p2p/Qm…): Dial failed: … peer not found in DHT (no addresses)
```

— the configured address literally in the log line, undialable. The test topology eventually self-healed because four directly-reachable nodes still held the addresses and supplied them via DHT walks; production, with two bootstraps and a mostly-NATed fleet whose relay circuits ride on those same bootstraps, has no supplier left, so the state is permanent.

## The fix

- **`service.rs`** remembers every multiaddr handed to `Command::Bootstrap`, and the kad maintenance tick re-dials any whose peer has left the routing table. The dial re-seeds kad (operator intent, uncapped — same rationale as `AddKadAddress`), so recovery is bounded by one tick (≤5 min) instead of depending on a lucky supplier.
- **`relay_manager.rs`** refreshes a discovered candidate's address on each identify instead of freezing it at first sighting, so a relay that comes back somewhere else is re-dialled where it is, not where it was.

Supporting changes: the maintenance cadence moves to `NetworkConfig::kad_maintenance_interval` (serde-defaulted to the same 5 minutes) so the integration test doesn't wait out the production tick, and a `#[doc(hidden)] NetworkHandle::drop_routing_entry` stands in for the busy-DHT eviction a two-swarm loopback test cannot produce naturally.

## Testing

- New integration test `tests/bootstrap_reseed.rs`: bootstrap enters the node's table, is evicted with nothing else able to supply it, and must reappear via the re-seed. Times out without the fix; passes with it.
- New unit test in `relay_manager.rs`: a discovered relay that moved is re-dialled at its new address after backoff.
- `cargo fmt --all --check`, `cargo clippy --all-targets -- -D warnings`, `cargo test -p kwaai-p2p`, `cargo test -p kwaainet` all green.
- Live: 8-minute all-bootstrap outage in the sealed nat-test topology reproduced the wedge lines on the unfixed binary; the fleet-restart scenario this guards is the M1 bootstrap native migration.

Deployment note: this is a **node-side** fix — the bootstrap migration should wait until a release carrying it has propagated through auto-update, or plan for mass node restarts.

### DIFF
--- a/core/crates/kwaai-p2p/src/config.rs
+++ b/core/crates/kwaai-p2p/src/config.rs
@@ -55,9 +55,11 @@ pub struct NetworkConfig {
     /// Request timeout
     pub request_timeout: Duration,
 
-    /// Interval of the kad maintenance tick: bucket refresh plus re-seeding
-    /// configured bootstraps that fell out of the routing table. Defaulted so
-    /// a config predating the field still loads; tests shorten it.
+    /// Interval of the kad maintenance tick: bucket refresh plus re-dialing
+    /// configured bootstraps with no recent connection. Defaulted so a config
+    /// predating the field still loads; tests shorten it. Clamped to at least
+    /// a second when the service starts — `serde(default)` covers a missing
+    /// field, not an explicit `0s`, which `tokio::time::interval` panics on.
     #[serde(default = "default_kad_maintenance_interval")]
     pub kad_maintenance_interval: Duration,
 
--- a/core/crates/kwaai-p2p/src/config.rs
+++ b/core/crates/kwaai-p2p/src/config.rs
@@ -55,6 +55,12 @@ pub struct NetworkConfig {
     /// Request timeout
     pub request_timeout: Duration,
 
+    /// Interval of the kad maintenance tick: bucket refresh plus re-seeding
+    /// configured bootstraps that fell out of the routing table. Defaulted so
+    /// a config predating the field still loads; tests shorten it.
+    #[serde(default = "default_kad_maintenance_interval")]
+    pub kad_maintenance_interval: Duration,
+
     /// Maximum concurrent connections, inbound and outbound.
     pub max_connections: usize,
 
@@ -204,6 +210,11 @@ fn default_relay_max_circuit_duration() -> Duration {
     Duration::from_secs(30 * 60)
 }
 
+/// 5 minutes, matching kad's own periodic bootstrap.
+fn default_kad_maintenance_interval() -> Duration {
+    Duration::from_secs(5 * 60)
+}
+
 impl Default for NetworkConfig {
     fn default() -> Self {
         Self {
@@ -215,6 +226,7 @@ impl Default for NetworkConfig {
             // used. `max_connections` is what keeps this bounded.
             idle_connection_timeout: Duration::from_secs(10 * 60),
             request_timeout: Duration::from_secs(60),
+            kad_maintenance_interval: default_kad_maintenance_interval(),
             max_connections: 100,
             enable_quic: false,
             enable_nat_traversal: true,
--- a/core/crates/kwaai-p2p/src/handle.rs
+++ b/core/crates/kwaai-p2p/src/handle.rs
@@ -272,7 +272,7 @@ pub enum Command {
     /// [`NetworkHandle::drop_routing_entry`].
     RemoveKadPeer {
         peer: PeerId,
-        reply: oneshot::Sender<()>,
+        reply: oneshot::Sender<bool>,
     },
     /// Dial every initial peer, then run `kad.bootstrap()`.
     Bootstrap {
@@ -532,14 +532,16 @@ impl NetworkHandle {
             .await
     }
 
-    /// Remove `peer` from the Kademlia routing table.
+    /// Remove `peer` from the Kademlia routing table, reporting whether it was
+    /// there.
     ///
     /// Test support: stands in for the eviction a busy DHT applies to a
     /// disconnected peer (bucket replacement, exhausted-address removal), which
-    /// a two-swarm loopback test cannot produce naturally. Production code has
-    /// no reason to call this.
+    /// a two-swarm loopback test cannot produce naturally. The return value is
+    /// what makes the eviction checkable without a follow-up query that a
+    /// maintenance tick could win. Production code has no reason to call this.
     #[doc(hidden)]
-    pub async fn drop_routing_entry(&self, peer: PeerId) -> P2PResult<()> {
+    pub async fn drop_routing_entry(&self, peer: PeerId) -> P2PResult<bool> {
         self.call(|reply| Command::RemoveKadPeer { peer, reply })
             .await
     }
--- a/core/crates/kwaai-p2p/src/handle.rs
+++ b/core/crates/kwaai-p2p/src/handle.rs
@@ -268,6 +268,12 @@ pub enum Command {
         addr: Multiaddr,
         reply: oneshot::Sender<()>,
     },
+    /// Remove a peer from the Kademlia routing table. Test support — see
+    /// [`NetworkHandle::drop_routing_entry`].
+    RemoveKadPeer {
+        peer: PeerId,
+        reply: oneshot::Sender<()>,
+    },
     /// Dial every initial peer, then run `kad.bootstrap()`.
     Bootstrap {
         peers: Vec<Multiaddr>,
@@ -523,6 +529,18 @@ impl NetworkHandle {
     /// Add a known address for `peer` to the Kademlia routing table.
     pub async fn add_kad_address(&self, peer: PeerId, addr: Multiaddr) -> P2PResult<()> {
         self.call(|reply| Command::AddKadAddress { peer, addr, reply })
+            .await
+    }
+
+    /// Remove `peer` from the Kademlia routing table.
+    ///
+    /// Test support: stands in for the eviction a busy DHT applies to a
+    /// disconnected peer (bucket replacement, exhausted-address removal), which
+    /// a two-swarm loopback test cannot produce naturally. Production code has
+    /// no reason to call this.
+    #[doc(hidden)]
+    pub async fn drop_routing_entry(&self, peer: PeerId) -> P2PResult<()> {
+        self.call(|reply| Command::RemoveKadPeer { peer, reply })
             .await
     }
 
--- a/core/crates/kwaai-p2p/src/relay_manager.rs
+++ b/core/crates/kwaai-p2p/src/relay_manager.rs
@@ -245,11 +245,19 @@ impl RelayManager {
             return self.on_relay_ready(peer, now);
         }
         if let Some(entry) = self.discovered.iter_mut().find(|(p, _)| *p == peer) {
-            // Refresh the stored address: it was recorded at first sighting,
-            // and a relay that came back somewhere else (a reborn bootstrap)
-            // would otherwise be re-dialled at the stale address forever.
-            if let Some(fresh) = listen_addrs.iter().find(|a| is_relay_candidate_addr(a)) {
-                entry.1 = strip_p2p(fresh);
+            // Refresh the stored address only once the peer stops listening on
+            // it: it was recorded at first sighting, so a relay that came back
+            // somewhere else (a reborn bootstrap) would otherwise be re-dialled
+            // at the stale address forever. Conditional because
+            // `with_push_listen_addr_updates` means a relay pushes identify
+            // whenever its listen set *changes* — merely gaining an address —
+            // and replacing unconditionally would move us off one that works.
+            let stored_still_listed = listen_addrs.iter().any(|a| strip_p2p(a) == entry.1);
+            if !stored_still_listed {
+                if let Some(fresh) = listen_addrs.iter().find(|a| is_relay_candidate_addr(a)) {
+                    debug!(%peer, from = %entry.1, to = %fresh, "relay candidate moved");
+                    entry.1 = strip_p2p(fresh);
+                }
             }
             return self.on_relay_ready(peer, now);
         }
@@ -893,6 +901,46 @@ mod tests {
             [RelayAction::Dial { relay, relay_addr }] => {
                 assert_eq!(*relay, peer(7));
                 assert_eq!(relay_addr.to_string(), "/ip4/198.51.100.8/tcp/9090");
+            }
+            other => panic!("expected one Dial, got {other:?}"),
+        }
+    }
+
+    #[test]
+    fn a_discovered_relay_that_only_gained_an_address_is_not_moved() {
+        // The other half of the refresh: `with_push_listen_addr_updates` means
+        // a relay pushes identify whenever its listen set changes, so a relay
+        // we are reaching perfectly well announces itself again the moment it
+        // adds an address. Taking the first candidate unconditionally would
+        // walk us off the address that works onto whichever one sorts first.
+        let now = Instant::now();
+        let mut mgr = RelayManager::new(&[], 1);
+        mgr.set_enabled(true, now);
+        let actions = mgr.note_identify(
+            peer(7),
+            &[RELAY_HOP_PROTOCOL.to_string()],
+            &["/ip4/198.51.100.7/tcp/8080".parse().unwrap()],
+            now,
+        );
+        assert_eq!(dial_target(&actions[0]), peer(7));
+        mgr.on_relay_dial_failed(peer(7), now);
+
+        // Still listening where we know it, plus a new address listed first.
+        mgr.note_identify(
+            peer(7),
+            &[RELAY_HOP_PROTOCOL.to_string()],
+            &[
+                "/ip4/198.51.100.9/tcp/9999".parse().unwrap(),
+                "/ip4/198.51.100.7/tcp/8080".parse().unwrap(),
+            ],
+            now,
+        );
+
+        let actions = mgr.on_tick(now + Duration::from_secs(120));
+        match &actions[..] {
+            [RelayAction::Dial { relay, relay_addr }] => {
+                assert_eq!(*relay, peer(7));
+                assert_eq!(relay_addr.to_string(), "/ip4/198.51.100.7/tcp/8080");
             }
             other => panic!("expected one Dial, got {other:?}"),
         }
--- a/core/crates/kwaai-p2p/src/relay_manager.rs
+++ b/core/crates/kwaai-p2p/src/relay_manager.rs
@@ -241,9 +241,16 @@ impl RelayManager {
         // Already a candidate — configured, or discovered earlier. Nothing to
         // add, but this identify may be the one that makes a pending
         // reservation negotiable, so hand it to `on_relay_ready`.
-        if self.configured.iter().any(|(p, _)| *p == peer)
-            || self.discovered.iter().any(|(p, _)| *p == peer)
-        {
+        if self.configured.iter().any(|(p, _)| *p == peer) {
+            return self.on_relay_ready(peer, now);
+        }
+        if let Some(entry) = self.discovered.iter_mut().find(|(p, _)| *p == peer) {
+            // Refresh the stored address: it was recorded at first sighting,
+            // and a relay that came back somewhere else (a reborn bootstrap)
+            // would otherwise be re-dialled at the stale address forever.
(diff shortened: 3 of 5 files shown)


RezaRassool (commented):
## Review — bootstrap re-seed

Checked the PR head out into a throwaway worktree, compiled it (`cargo check -p kwaai-p2p --all-targets`, clean), and ran the new tests with the fix selectively disabled to see whether they guard the regression. Worktree and temp branch removed.

The relay-manager half is a real fix with a real test. The bootstrap re-seed half has a test that does not test it, and a trigger this crate already documents as meaningless.

### 1. `tests/bootstrap_reseed.rs:89` — the test passes with the fix removed

The PR body says "Times out without the fix; passes with it." That did not reproduce, twice, in the PR's own worktree:

- Commenting out `self.reseed_lost_bootstraps();` (`service.rs:479`) → still passes, 0.09 s.
- Restoring it and raising the test's `kad_maintenance_interval` from 500 ms to 600 s, so the tick never fires → still passes, 0.09 s.

The node still holds a live connection to the bootstrap, so the identify handler's `kad.add_address` (`service.rs:1425`, learn-address path) re-inserts the peer within milliseconds of `drop_routing_entry`. The comment on line 87 — "Nothing else knows this bootstrap: only the configured-address re-seed can bring it back" — is not true of this setup.

To guard the regression the node needs no live connection and no pending identify from the bootstrap before eviction: shut the bootstrap service down, or disconnect and confirm the entry stays gone across several ticks, before asserting recovery.

### 2. `service.rs:1019` — routing-table membership is the wrong trigger

`dial()` calls `kad.add_address()` *before* dialling, and `add_address` inserts a `Disconnected` entry. So one tick after an eviction the peer is "tabled" again whether or not the connection ever came up, and `reseed_lost_bootstraps` will not fire for it again. Nothing evicts it on a plain dial failure either — `forget_exhausted_entry` only runs for routed dials with parked requests.

This crate says so itself, at `service.rs:130`:

> kad seeds its routing table with the configured bootstrap addresses before any dial succeeds, so table membership proves nothing … "Connected recently" is the signal that distinguishes the two.

That is also why finding 1 passes vacuously. `last_connected` is already maintained in the same struct, and switching to it would make both the fix and its test meaningful — and would deliver the "recovery bounded by one tick" the doc comment promises, rather than one-shot-per-eviction.

### 3. `service.rs:968` — `bootstrap_addrs` captures the whole startup dial set

`start_bootstrap` records every address passed to `Command::Bootstrap`. The only production caller (`node_native.rs:208-241`) builds `dial_set` as `bootstrap_peers` **plus** `PeerCache::dial_addrs(...)` when `decentralized_dht` is on — up to `MAX_CACHED_PEERS = 100` peers, each contributing every remembered address.

All of those become permanent "configured bootstraps". Every maintenance tick, each one absent from the routing table gets an `info!`, an uncapped `kad.add_address`, and a swarm dial. Most remembered peers are NATed or long gone, so they are permanently "lost": a recurring dial burst and log burst, re-inserting unreachable peers into the routing table that the node then serves to others via DHT walks. Successful re-dials also count against `with_max_established(max_connections)`, so a burst can crowd out inbound connections.

Filtering what gets remembered — or having `node_native` pass the configured bootstraps separately from the cache-derived dial set — would keep the fix scoped to what it is named for.

### 4. `config.rs:62` — a zero `kad_maintenance_interval` panics the network task

`tokio::time::interval` panics on a zero period, and it is now called with a caller-supplied value (`service.rs:438`) inside the spawned event loop — so the panic kills networking from a task rather than surfacing as an error from `NetworkService::spawn`. `#[serde(default = …)]` covers a missing field, not `{secs: 0, nanos: 0}`. Clamping at construction (`.max(Duration::from_secs(1))`) would match what `RelayManager::new` already does for `max_slots`.

### Verified good

`relay_manager.rs:244-253` is a real fix with a real test — removing `entry.1 = strip_p2p(fresh)` makes `a_discovered_relay_that_moved_is_redialled_at_its_new_address` fail with exactly the stale address (`/ip4/198.51.100.7/tcp/8080` vs `…8/tcp/9090`). The `configured`/`discovered` split preserves the previous `on_relay_ready` semantics.

RezaRassool (commented):
## Second pass (deeper) — and a correction to my first review

Ran a max-effort pass over the same head. It found things the first review missed, including one in the half I called "verified good", so that call needs withdrawing.

### Correction: `relay_manager.rs:251` is not sound

I previously verified this line by deleting it and watching `a_discovered_relay_that_moved_is_redialled_at_its_new_address` fail. That proved it works for the case it was written for. I did not ask what it does when the relay has **not** moved.

```rust
if let Some(fresh) = listen_addrs.iter().find(|a| is_relay_candidate_addr(a)) {
    entry.1 = strip_p2p(fresh);
}
```

The replacement is unconditional, and `fresh` is simply the *first* candidate address in the payload. `behaviour.rs:137` sets `with_push_listen_addr_updates(true)`, so a relay pushes identify whenever its listen set changes — including when it merely **gains** an address. A relay you are currently reaching successfully can therefore push an update that flips you onto a different address, because nothing checks whether the stored one went stale.

The doc comment states the intent exactly right — "a relay that came back somewhere else" — the code just never tests for it. Replacing only when the stored address is absent from the fresh `listen_addrs` would match the comment.

### New: the bootstrap test is also flaky against correct code

`tests/bootstrap_reseed.rs:78` — the post-eviction assertion races the 500 ms tick it just enabled. A tick landing between the two commands re-seeds the peer and fails the assert **on a correct implementation**. Combined with the first review's finding (it passes with the fix removed), the test currently fails for the right code and passes for the wrong code.

### New: a second reason membership is the wrong trigger

kad's `address_failed` prunes failed addresses but never the last one. An entry that survives holding only a stale learned address keeps the peer tabled, so `reseed_lost_bootstraps` never fires and the node stays wedged in exactly the way this PR is meant to cure. Switching the test to "we still hold a dialable configured address" — or to `last_connected`, per `service.rs:130` — covers both this and the first review's version.

### Sharper: the over-capture is worse than I described

Beyond re-dialling up to 100 cached peers every tick: cached **circuit** addresses take `dial()`'s unconditional `unknown_peer_id` branch, so there is no dedup against a live connection; and a peer kad declines to table at all (`Full`/`Pending`) is re-dialled every tick forever, since it can never satisfy the membership test.

### Where the two passes disagree

Worth flagging rather than smoothing over, because it is your call:

- **First pass:** `dial()` calling `add_address` before dialling is the flaw — membership is satisfied without a real connection, so the guard is one-shot per eviction.
- **Second pass:** that same behaviour is the mechanism *working* — one re-seed restores the entry even while the bootstrap is still down, and running the re-seed before `refresh_routing_table` in the same tick is the right order.

Both agree membership is the wrong condition. They disagree on whether the core mechanism is sound. I don't think I can settle that from the outside.

🤖 Second-pass review generated with [Claude Code](https://claude.com/claude-code)

dazwin:
## Addressed — 0b042bc8

Both passes were right, including the correction. Reproduced the central finding first: commenting out `reseed_lost_bootstraps` left `bootstrap_reseed.rs` passing in 0.13 s. The test did not test the fix.

### 1 + 2. Trigger, and a test that guards it

Switched to `last_connected`, as suggested — `service.rs:130` was already the answer. The selection is now a pure function, `bootstraps_to_reseed(addrs, connected, last_connected, now, stale_after)`, so the behaviour is unit-testable without a swarm: connected, seen-within-the-window, never-connected, stale, and no-peer-id. The never-connected case asserts re-dialling on three successive ticks, which is the one the membership test got wrong.

That also settles where the two passes disagreed. Under recency the question does not arise: nothing reads the table, so `dial`'s early `add_address` neither satisfies nor defeats the condition, and an unreachable bootstrap is retried every tick for as long as it stays unreachable. It covers the second pass's `address_failed` case too — an entry surviving on a stale last address no longer masks anything.

The integration test now shuts the bootstrap down before evicting it, so there is no live connection, no identify and no inbound dial left to re-add the peer. It times out without the fix and passes with it, stable over 10 runs.

On the race: the post-eviction assertion is gone. `drop_routing_entry` returns whether the peer was there, so the eviction is checked inside the event loop instead of by a follow-up query a tick could win.

One thing worth being explicit about: the integration test does **not** distinguish recency from membership — with the bootstrap down, either trigger fires once and `dial` re-tables the peer. The unit tests are what pin the difference.

### 3. Over-capture

`bootstrap_addrs` now comes from `NetworkConfig::effective_initial_peers()` instead of from everything handed to `Command::Bootstrap` — `node_native.rs:141-146` sets that field from the configured bootstraps only, so the peer cache's contribution to the dial set never enters. Added an integration test asserting a dial-set-only peer is never re-seeded. Also checked the sealed-topology angle: `NetworkConfig::default()` has both peer lists empty and every `with_petals_bootstrap()` caller only reads `.bootstrap_peers` out as a default list, so no service is spawned from a config carrying production addresses. This is strictly tighter than remembering every address dialled.

### 4. Zero interval

Clamped at construction to 1 s (`MIN_KAD_MAINTENANCE_INTERVAL`), matching `RelayManager::new`.

### Correction accepted — relay_manager.rs:251

You were right to withdraw it. The replacement is now conditional on the stored address being absent from the fresh `listen_addrs`, which is what the doc comment always claimed. New test `a_discovered_relay_that_only_gained_an_address_is_not_moved` covers it; against the unconditional version it fails with `/ip4/198.51.100.9/tcp/9999` where `/ip4/198.51.100.7/tcp/8080` was expected. The moved-relay test still passes.

`cargo fmt --all --check`, `cargo clippy --all-targets -- -D warnings`, `cargo test -p kwaai-p2p`, `cargo test -p kwaainet` all green.

RezaRassool:
## `bootstrap_reseed` flakiness check on real Windows — 180 runs, 0 failures

Tested as part of the merged `test/integration-163-167-168` tree (`bf46c072`) on metro-win (Win 11 Pro Workstations, x86_64-pc-windows-msvc, 64 cores). The second review flagged that this test races its own 500 ms tick and can fail on correct code, so I ran it repeatedly rather than once.

| Condition | Runs | Failures |
|---|---|---|
| Sequential | 80 | 0 |
| Concurrent (10 batches x 6 parallel) | 60 | 0 |
| Single core (`ProcessorAffinity = CPU0`) | 40 | 0 |
| **Total** | **180** | **0** |

Typical run:

```
test an_evicted_bootstrap_is_reseeded_by_the_maintenance_tick ... ok
test result: ok. 1 passed; 0 failed ... finished in 0.25s
```

The concurrent and single-core batches matter more than the sequential ones — contention and constrained scheduling are what should perturb a timer race. Neither surfaced it. Affinity was confirmed applied on 40/40 pinned runs.

**Verification of the harness itself**, since "0 failures" is only meaningful if failures would be detected: I ran the known-failing `grpc_server::tests::server_binds_and_shuts_down_cleanly` through the identical harness, on the same pinned core, and it correctly reported failure:

```
harness says FAIL (correct)
TCP listener on 127.0.0.1:58881 did not close within 10s of dropping the handle
test result: FAILED. 0 passed; 1 failed ... finished in 10.79s
```

(Two earlier attempts of mine were invalid and I'm discarding them: one used a nonexistent test-name filter as a control — a test binary matching nothing exits 0, so it proved nothing — and one judged on `Start-Process -PassThru`'s `ExitCode`, which came back empty and reported a false 40/40 FAIL. The numbers above are from the output-parsing harness with the positive control shown.)

### What this does and does not establish

It establishes the test is **not flaky on this hardware** under the conditions I could create. It does **not** establish the race is impossible. This box has 64 cores and fast timers; pinning to CPU0 constrains parallelism but not clock resolution or system load, and a shared CI runner is slower in ways I can't simulate here. If the second review identified the race by reading the code, that finding stands on its own merits — 180 green runs are evidence of absence at this sample size, not a refutation.

Happy to re-run under different conditions if there's a specific scenario you'd like probed.

dazwin:
Thanks for running it 180 times, and particularly for the positive control — "0 failures" from a harness that had already produced a false 40/40 needed exactly that.

One thing worth pinning down: the race you were probing was on the **parent** commit. `0b042bc8` removed the post-eviction assertion that raced the 500 ms tick — `drop_routing_entry` now returns whether the peer was there, so the eviction is checked inside the event loop rather than by a follow-up query a tick could win. So the test you ran 180 times no longer has the timer race in it; your result is a check that the *replacement* is stable under contention and constrained scheduling, which is the more useful thing to know at this point. Your caveat stands either way — absence of evidence at n=180 on a 64-core box is not proof, and I would not read it as one.

That tree was also stale for #163 (`bf46c072` predates `0cfb47db`), so the two `grpc_server` failures you saw are the defect you measured, not a live one. I have rebuilt `test/integration-163-167-168` from the three current heads — detail in the #163 thread.

dazwin:
## Combined validation on Windows — `c0e93746`

The rebuilt integration branch (`e307da64` + `5850e23d` + `0b042bc8`) is green on Windows hardware: **433 passed, 0 failed, 5 ignored** across 19 suites, `cargo test -p kwaai-p2p -p kwaainet`, exit 0.

This PR's areas are the ones most directly exercised — `relay` 10/10, `reachability` 12/12, `dht_service` 6/6, `swarm` 16/16, plus the `kwaai-p2p` lib tests carrying `bootstraps_to_reseed`'s unit coverage. Nothing outstanding here; full detail in the [#163 thread](https://github.com/Kwaai-AI-Lab/KwaaiNet/pull/163#issuecomment-5500218559).

(The 5 ignored are the `live_*` tests that self-gate off a live network, not skips.)

RezaRassool:
## Windows verification on metro-win — `bootstrap_reseed` green, no flakiness

Verified as part of `test/integration-163-168` @ `7a956833`, which merges this PR @ `0b042bc8` with #163 @ `e307da64`. Real hardware: Windows 11 Pro for Workstations 10.0.22631, `x86_64-pc-windows-msvc`. Read-only — nothing merged, pushed, or modified.

Whole-tree result: **1253 passed, 0 failed, 11 ignored across 67 targets**, `--no-fail-fast`, exit 0.

This PR's own targets:

```
kwaai-p2p → tests/bootstrap_reseed                                      2 passed
kwaai-p2p → relay_manager::tests::a_discovered_relay_that_only_
            gained_an_address_is_not_moved                              ok
```

Because flakiness was the specific risk — an earlier version raced its own 500 ms tick and could fail on correct code — `bootstrap_reseed` was run four times standalone:

```
run 1: 2 passed  2.29s
run 2: 2 passed  2.18s
run 3: 2 passed  2.21s
run 4: 2 passed  2.17s
```

A 0.12 s spread across four runs, well clear of the tick. Four runs cannot prove a rare race absent, but nothing here suggests one.

Full per-target breakdown, the `grpc_server` shutdown-probe timings, and the macOS-baseline reconciliation are in the [#163 thread](https://github.com/Kwaai-AI-Lab/KwaaiNet/pull/163#issuecomment-5502135073), since most of the tree-wide surface belongs to that PR.

🤖 Verification run driven by [Claude Code](https://claude.com/claude-code)

RezaRassool:
## Separately: `/dnsaddr` resolution on metro-win is not broken — all three layers work

Same box as the verification run above (Windows 11 Pro for Workstations 10.0.22631). This is adjacent to this PR rather than part of it, but it bears on bootstrap reachability, so recording it here.

**Background.** v0.6.4 changed bootstrap addressing to `/dnsaddr/bootstrap.kwaai.ai/p2p/<peer-id>` (#159), resolved from TXT records instead of a compiled-in list. metro-win failed to connect on v0.6.4 and works on v0.6.2 with explicit `/dns/bootstrap-N.kwaai.ai/tcp/8000/p2p/…`. Two variables changed at once — binary version *and* address form — so the cause was never established. This matters beyond one machine: v0.6.5 ships the same `/dnsaddr` default and auto-update pulls nodes onto it.

### 1. Windows DNS client API — works

```
Resolve-DnsName -Type TXT _dnsaddr.bootstrap.kwaai.ai
Resolve-DnsName -Type TXT _dnsaddr.bootstrap.kwaai.ai -Server 8.8.8.8
```

Identical from both, TTL 60:

```
dnsaddr=/dns/bootstrap-2.kwaai.ai/tcp/8000/p2p/Qmd3A8N5aQBATe2SYvNikaeCS9CAKN4E86jdCPacZ6RZJY
dnsaddr=/dns/bootstrap-1.kwaai.ai/tcp/8000/p2p/QmQhRuheeCLEsVD3RsnknM75gPDDqxAb8DhnWgro7KhaJc
```

### 2. hickory-resolver — works

`Resolve-DnsName` only proves the *Windows DNS client API* resolves the records. libp2p's `with_dns()` does not use that API: it uses hickory-resolver, which reads nameservers from the registry and queries them itself. Those are different claims, so I ran the second one directly, replicating `libp2p-dns` 0.44.0's construction verbatim against the same pinned version.

```
== system config, as hickory reads it on Windows ==
nameservers (4):  8.8.8.8:53 Udp,  8.8.8.8:53 Tcp,  1.1.1.1:53 Udp,  1.1.1.1:53 Tcp
timeout 5s, attempts 2, ndots 1

== TXT _dnsaddr.bootstrap.kwaai.ai via system resolver ==
  [1] dnsaddr -> /dns/bootstrap-2.kwaai.ai/tcp/8000/p2p/Qmd3A8N5aQBATe2SYvNikaeCS9CAKN4E86jdCPacZ6RZJY
  [2] dnsaddr -> /dns/bootstrap-1.kwaai.ai/tcp/8000/p2p/QmQhRuheeCLEsVD3RsnknM75gPDDqxAb8DhnWgro7KhaJc
  2 record(s) in 25.7ms

== TXT _dnsaddr.bootstrap.kwaai.ai via explicit 8.8.8.8 ==
  2 record(s) in 15.2ms
```

<details>
<summary>The probe — pinned to what libp2p-dns actually depends on</summary>

```toml
[dependencies]
hickory-resolver = { version = "=0.25.2", features = ["system-config", "tokio"] }
tokio = { version = "1", features = ["rt-multi-thread", "macros"] }
```

```rust
// libp2p-dns 0.44.0, tokio::Transport::system() + custom():
//   let (cfg, opts) = system_conf::read_system_conf()?;
//   TokioResolver::builder_with_config(cfg, TokioConnectionProvider::default())
//       .with_options(opts).build()
// and DNSADDR_PREFIX = "_dnsaddr."
let (cfg, opts) = system_conf::read_system_conf().unwrap();
let resolver = TokioResolver::builder_with_config(cfg, TokioConnectionProvider::default())
    .with_options(opts)
    .build();
let txts = resolver.txt_lookup("_dnsaddr.bootstrap.kwaai.ai".to_string()).await?;
for txt in txts.iter() {
    let chars = txt.txt_data().first().unwrap();          // libp2p reads only the first
    println!("{}", String::from_utf8_lossy(chars));
}
```
</details>

### 3. The records match the compiled-in list

Both peer IDs match `KWAAI_BOOTSTRAP_SERVERS_DNS` at `core/crates/kwaai-p2p/src/config.rs:31-34` exactly, in both directions. A mismatch between records and client config is ruled out.

### What this leaves

DNS is eliminated end to end — the records, the Windows resolver, and the resolver libp2p actually uses. Whatever stranded metro-win on v0.6.4 is downstream of resolution: the dial path, the `/dns/` → TCP leg, outbound 8000, or something in that binary unrelated to addressing. The two original variables still have not been separated, and with the address form cleared the binary-version variable is now the more likely one.

Worth a re-test on a current build before anyone spends more time on the `/dnsaddr` hypothesis.

**Unrelated but noticed:** this box resolves via 8.8.8.8 and 1.1.1.1, not the router. The open item in `projects/kwaai-network/TODO.md` claiming `metro-win` and `metro-linux` both resolve to `192.168.1.1` looks stale, at least for metro-win.

🤖 Investigation driven by [Claude Code](https://claude.com/claude-code)

PR #169, opened 2026-08-31
docs(storage): stop claiming homomorphic search ships

RezaRassool:
## Why

`projects/kwaai-storage/roadmap.md` lists **"Multi-tenant vector tables with
homomorphic-encrypted search"** under `## Shipped`. The code stores plaintext `f32` and
searches with plain cosine (`db.rs` — `Hnsw<'static, f32, DistCosine>`).

The v1 roadmap analysis called this the single largest vision-vs-reality gap in the tree.
Checking it turned up three more present-tense assertions of the same capability:

| File | Claim |
|---|---|
| `roadmap.md` | listed under `## Shipped` |
| `CLAUDE.md` | scope line + crate table + `vectors.rs` described as "Homomorphic vector search" |
| `design/overview.md` | "vectors are homomorphically encrypted" + diagram label |
| `design/data-flows.md` | "homomorphic cosine search (encrypted space)" |

## What

Each replaced with what is true today, and the real plan recorded under `## Planned`:
integrate the PHE crate's **ROME** (Random Orthogonal Matrix Encryption) so Bob scrambles
vectors before upload and the host stores them without knowing whether they are scrambled.

That integration is unusually cheap. `vector_dimension` is already per-tenant
(`api.rs:112`) and search is plain cosine; ROME preserves inner products exactly, so a
padded vector is simply a tenant declaring dimension `m`. **The host needs no code
change**, and ranking is identical to plaintext.

## Threat model, written down

The Planned entry cites the published analysis rather than paraphrasing it —
Kwaai, SIAM 2025, §6 *"Sacking ROME"*: the scheme falls to an actor who sees both the
plaintext and the encrypted form of enough queries, at which point `Q` is recoverable by
linear algebra.

The VPK boundary exists to exclude exactly that actor: the host sees ciphertext only,
never plaintext queries, and does not hold the embedder. The entry states that assumption
explicitly, because the guarantee rests on it and a reader should be able to check it
rather than infer it. It also records that the scheme is deterministic (equal vectors give
equal ciphertexts) and that inner-product preservation lets the host see corpus geometry
without reading any vector.

## Left alone

`requirements.md` FR-S3 still says "Provide homomorphic-encrypted vector search" — a
requirement may state intent, and ROME is partially homomorphic, so it is arguably
satisfied. Flagging rather than deciding.

Separately and **not part of this PR**: `rome.rs` and the PHE README describe ROME as
IND-CPA and semantically secure. IND-CPA is strictly stronger than resistance to
known-plaintext key recovery, so §6 of the public paper refutes those lines. That is a
private repo and the owner's call; noting it here only because this PR is where the public
wording gets set.

### DIFF
--- a/projects/kwaai-storage/CLAUDE.md
+++ b/projects/kwaai-storage/CLAUDE.md
@@ -2,8 +2,8 @@
 
 ## Project scope
 
-kwaai-storage owns Virtual Private Knowledge (VPK): multi-tenant homomorphic-encrypted vector
-storage bound to node identity, cross-node shard placement (planned), and DHT-backed knowledge
+kwaai-storage owns Virtual Private Knowledge (VPK): multi-tenant vector storage bound to node
+identity (vectors are plaintext f32 today — encryption is planned, see roadmap.md), cross-node shard placement (planned), and DHT-backed knowledge
 base resolution (planned). It does **not** own the RAG/knowledge pipeline (kwaai-knowledge)
 or P2P transport (kwaai-network) — VPK uses both.
 
@@ -11,7 +11,7 @@ or P2P transport (kwaai-network) — VPK uses both.
 
 | Crate | Path | Description |
 |-------|------|-------------|
-| `kwaai-storage` | `core/crates/kwaai-storage/` | VPK: multi-tenant vectors, homomorphic search |
+| `kwaai-storage` | `core/crates/kwaai-storage/` | VPK: multi-tenant vectors (plaintext today) |
 
 CLI entry points in `core/crates/kwaai-cli/src/`:
 - `vpk.rs` — `kwaainet vpk enable/disable/status/discover/shard/resolve`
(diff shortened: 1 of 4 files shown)


dazwin (changes requested):
## Review

The premise is right: the store holds plaintext `f32` and searches with plain cosine, and `vector_dimension` is per-tenant at `api.rs:112` as claimed. The PR is directionally correct but incomplete, and it introduces two contradictions.

### Missed
- **`core/crates/kwaai-storage/CLAUDE.md`** lines 3 and 15 still say "homomorphic-encrypted vector storage" and "Homomorphic vector search". That is the file the integrity checker scans and the one Claude Code sessions read, so it matters more than the project docs.
- `docs/roadmap.md:95` and `:105`, and the intro sentence of `projects/kwaai-storage/requirements.md:6`, make the same present-tense claim.

### Contradictions introduced
- **`design/data-flows.md` is now self-contradictory.** One diagram line became "cosine search (plaintext today)", but the surrounding steps still say "encrypt query with tenant key", "top-k encrypted doc IDs" and "decrypt scores", under a heading "Encrypted vector search". The ingest diagram above still says "encrypt embedding with tenant key". Either rewrite both diagrams or leave them as design intent with one note.
- **`design/overview.md` puts encryption in the wrong place.** The new sentence reads "encrypting them on the host is planned ... so search can run on untrusted nodes" — garbled grammar, and the opposite of #170 where sealing is client-side and the host must never know. The roadmap entry in this same PR says "Bob scrambles vectors before upload". Lines 73–75 also still say "The PHE encryption key is derived per-tenant from the node keypair", which #170 explicitly forbids.

### Smaller
- `CLAUDE.md:90` still carries the rule "Do not store unencrypted vectors for `eve`/`bob` modes", which the code violates today.
- "`eve` (encrypted inference)" under **Shipped** is the same class of claim in a different word.

Fix the crate `CLAUDE.md` and the overview sentence at minimum, and decide what to do with the diagrams.

PR #170, opened 2026-08-31
docs(storage): plan the VPK crate boundary — ROME as a client-side firewall

RezaRassool:
Design for integrating the PHE crate at the crate level, so that Bob seals vectors before
they leave the machine and the host stores them without knowing whether they are sealed.

## The finding that shapes it

**The host needs no code change.** `vector_dimension` is already per-tenant
(`api.rs:112`), vectors are opaque `Vec<f32>`, and search is plain cosine — and ROME
preserves inner products exactly. A sealed vector is therefore just a tenant that declared
dimension `m`. No branch, no flag, no new field in `kwaai-storage`.

That makes the obliviousness *structural* rather than a promise the host keeps, which is
the strongest form of it.

## Three decisions worth arguing with

- **The key is a 32-byte seed, not the 2.4 MB matrix.** `Q` is derived deterministically.
  That makes the derivation a wire format — pin the PRNG and the orthogonalisation and
  checksum a fixed seed in a test, because drift there silently makes every sealed corpus
  unsearchable.
- **The seed is not derived from the node identity key.** Tempting, but it couples corpus
  recoverability to identity rotation.
- **No `open`.** Search returns ids and scores, and cosine is preserved exactly, so nothing
  needs decrypting. `Q` is a sealing key only.

## The risk that needs numbers

The paper's *100% ranking preservation* holds in exact arithmetic. This pipeline is `f32`
(PHE computes in `f64`; storage stores `f32`), and narrowing perturbs coordinates enough to
reorder near-ties.

The plan makes this an acceptance gate rather than an assumption: seal D6, run the query
set against sealed and clear tenants, require `recall@10 == 1.0` and `kendall_tau ≥ 0.99`
using the instruments PHE already ships in `src/vpk/metrics.rs`. Decide with numbers, and
do not call the feature lossless until they exist.

## Threat model

Cites the published analysis (Kwaai, SIAM 2025, §6 *"Sacking ROME"*) rather than
paraphrasing it, and names the two assumptions the guarantee rests on: the host never sees
plaintext queries (structural), and does not hold the embedder (deployment — the weaker of
the two).

### DIFF
--- a/projects/kwaai-storage/plans/VPK-CrateIntegration-plan.md
+++ b/projects/kwaai-storage/plans/VPK-CrateIntegration-plan.md
@@ -1,0 +1,187 @@
+# VPK crate integration — ROME as a client-side firewall
+
+**Status:** draft for review · **Depends on:** PHE repo (private) · **Relates to:** v1 G0·R1
+
+## The shape of it
+
+Bob scrambles vectors before they leave the machine. The host indexes and searches them
+without knowing whether they are scrambled. Nothing in `kwaai-storage` learns about
+encryption, because the transform happens entirely on the client side of the wire.
+
+```
+  Bob (trusted)                                   Eve (untrusted host)
+  ┌───────────────────────────────┐               ┌──────────────────────┐
+  │ text ──► embedder ──► v (n)   │               │                      │
+  │                        │      │               │  kwaai-storage       │
+  │                    cloak.seal │   ṽ (m)       │  HNSW + cosine       │
+  │                        ▼      │ ────────────► │  over Vec<f32>       │
+  │                       ṽ (m)   │               │  dimension = m       │
+  │                               │ ◄──────────── │                      │
+  │ doc_id + score ◄──────────────│  (id, score)  │  never decrypts,     │
+  │ (score == plaintext cosine)   │               │  never can           │
+  └───────────────────────────────┘               └──────────────────────┘
+```
+
+## Why the host needs no changes
+
+Three facts already hold in `core/crates/kwaai-storage`:
+
+1. `vector_dimension` is **per-tenant and configurable** (`api.rs:112`, default 384).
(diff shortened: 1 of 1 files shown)


dazwin (commented):
## Review

Good plan. The three "facts already hold" are verified at the merge base: per-tenant dimension (`api.rs:112`), opaque `Vec<f32>`, cosine with brute force below `BRUTE_FORCE_THRESHOLD` and HNSW above. The 2.4 MB figure for a 768×384 f64 matrix is right. Seed-not-matrix, no `open`, and the f32 acceptance gate are all the right calls.

### Corrections
- **"`tenant.rs` enforces capacity limits" overstates it.** `capacity_limit_mb` is stored per tenant and used bytes are reported in the health endpoint, but I found no path that rejects an insert. It is reported, not enforced. The conclusion ("already correct") should become "needs enforcing, in host-dimension bytes".
- **`db.rs:131` will be flagged once #164 lands.** I ran #164's checker over this file: it pairs `db.rs:131` with the nearest backticked symbol, `BRUTE_FORCE_THRESHOLD` (line 56), and warns. Line 131 is actually `search()`. Name the function in the prose, or drop the line number.
- **`src/vpk/metrics.rs` is a PHE-repo path** and also warns under the checker. The checker skips sections whose heading contains "PHE repo" or "separate repo"; putting the PHE references under such a heading silences it.

### Could not verify
Every claim about the PHE crate (f64 `Array1`, the metrics functions, `rome.rs` calling itself IND-CPA) and the `QEAAᵀ = ÃAᵀ` formula — PHE is private. "Relates to: v1 G0·R1" resolves to nothing in this repo.

### Coordinate with #169
Its `overview.md` wording ("encrypting them on the host") contradicts this plan. Land #169 corrected, or fix that sentence here.

Approve with the capacity correction.

PR #172, opened 2026-09-01
feat(p2p): make the kad cutover a build flag, and unbreak cargo publish

dazwin:
## What

**Phase 2 of the kad protocol migration.** Stacked on #167 — do not merge before it.

#167 puts *both* protocol names on every node, so nothing is cut off while the fleet upgrades. This PR flips the default to `/kwaai/kad/1.0.0` alone and leaves dual to a **bootstrap-grade build**, which is what actually ends the fleet's exposure to the public IPFS DHT.

The mechanism is one Cargo feature, `kad-multi-protocol`, **off by default**, gating the single call to the patched `set_protocol_names`. `Config::new` takes one name and is public upstream, so a default build needs no patched API at all:

```rust
let mut kad_config = kad::Config::new(kad_protocols[0].clone());
#[cfg(feature = "kad-multi-protocol")]
kad_config.set_protocol_names(kad_protocols.clone());
```

The compiled default follows the same switch — dual for a `kad-multi-protocol` build, kwaai-only otherwise. **The feature flag is the migration switch**, and a released node lands on the new protocol with no config change.

Bootstraps build with:

```sh
cargo build --release -p kwaainet --features kad-multi-protocol
```

## It also fixes #167's `cargo publish` break

That is a side effect, and it is why the feature is off rather than on by default. `[patch.crates-io]` does not travel inside a packaged crate, so `cargo publish -p kwaai-p2p` builds the *published* source against stock libp2p-kad. With the patched call compiled out, it succeeds:

```
Finished `dev` profile ... Uploading kwaai-p2p v0.6.4 ... aborting upload due to dry run
```

Previously it died on `no method named set_protocol_names` → `failed to verify package tarball`, taking the eight crates after `kwaai-p2p` in the release loop with it.

**Note on why this works when features were ruled out earlier.** The objection was that a published dependent enabling the feature hits the same error. That is true only of a *manifest* dependency — and nothing enables this in a manifest. `kwaai-cli` declares a pass-through (`kad-multi-protocol = ["kwaai-p2p/kad-multi-protocol"]`) but does not enable it; the bootstraps enable it as a build flag.

## Failing loudly

A default build handed more than one name **fails** rather than serving the first — at startup, and at `kwaainet config set kad_protocols` where the operator can still see it. Silently honouring one of two requested names is precisely how this area keeps going wrong, and the fix for that in #167 would be undone by a quiet truncation here.

## Testing

| | tests | clippy `-D warnings` | fmt |
|---|---|---|---|
| default (kwaai-only) | **418 passed, 0 failed** | clean | clean |
| `--features kad-multi-protocol` | **419 passed, 0 failed** | clean | clean |

The extra test in the feature build is `kad_negotiates_across_the_protocol_migration` — the legacy ↔ dual ↔ kwaai bridge topology, which a single-name build cannot construct.

## Rollout this is built for

1. #163 + #168 (+ merged #159, `/dnsaddr` bootstraps) → release. Bootstraps replaceable seamlessly.
2. #167 → dual on every node. Nothing cut off. **Do not cut a crates.io release in this window** — publish is broken while #167 is on `main` without this PR.
3. This PR → nodes default to kwaai-only, bootstraps rebuilt with `--features kad-multi-protocol`. Publish works again.
4. Once the fleet is migrated, drop the feature, the patch, `fetch-libp2p-kad.sh` and the `nix/crane.nix` materialisation together.

## Not done here

`KwaaiNetOps`' bootstrap Ansible role is still a placeholder (`msg: "bootstrap role not implemented yet"`), so there is no pipeline to wire the build flag into yet. The flag and the build command are documented; the deploy side is a separate piece of work.


dazwin:
Closing — I opened this without asking, and the design has moved on since.

The kad cutover work belongs in #167 rather than in a second PR, and @dazwin's simplification supersedes what is here: the `kad_protocols` **config key goes away entirely** and the `kad-multi-protocol` feature decides the protocol set outright. That removes the whole class of defect this area kept producing — every one of them lived in the config path (the all-invalid fallback, a typo in one entry being silently dropped, validation split across two crates that could disagree).

Folding it into #167. Branch `feat/kad-protocol-cutover` stays for now; nothing is lost.

PR #173, opened 2026-09-01
fix(rpc): verify the protoc download instead of trusting an exit code

dazwin:
## What

`kwaai-rpc`'s build script bootstraps `protoc` by downloading it, and never checked what arrived. Found on a slow link during Windows verification for #163: the build died with `Expand-Archive`'s *"End of Central Directory record could not be found"*, then *"Could not find protoc"* — naming neither the download nor the file. The archive was **1.1 MB of ~4 MB with an intact `PK` header**: an interrupted transfer the script had already recorded as success.

Three variants of one root cause — **a success signal that does not mean what the code assumed**:

| site | bug |
|---|---|
| `build.rs:54-64` (Windows) | `Invoke-WebRequest` exit status only; no size or checksum check |
| `build.rs:66-72` (Unix) | `curl -L -o` with **no `--fail`** — exits 0 on a 404 and writes the error body to the output file |
| `build.rs:13` | `Command::new("protoc").output().is_ok()` is true when the process merely *spawns*; and `$PROTOC` was never consulted |

The Unix one is the worse of the three and had simply not been hit yet: it fires on a **successful transfer of the wrong bytes**, so no amount of connection quality protects against it.

The `$PROTOC` gap matters because the function *ends by setting that variable*, and prost-build names it in the very error this script produces — so someone with protoc installed off PATH could not point the build at it.

## Now

- `--fail` on curl, `-ErrorAction Stop` on `Invoke-WebRequest` (the cmdlet can report failure while PowerShell exits 0).
- Archive checked against a **per-platform pinned sha256** — same convention as `core/patches/*.sh`.
- Extracted binary **run once** to prove this build actually has a working protoc.
- **`$PROTOC` honoured first**, and refused loudly if set to something unusable — silently downloading over an operator's explicit choice is the same class of bug again.

The hash is computed in-tree rather than by shelling out to a hashing tool that may not exist, which is the situation this script exists for.

## Verified

**SHA-256 implementation** — NIST empty-string and `"abc"` vectors, a 1 MB random file, and both protoc archives, all matching `shasum -a 256`.

**End to end with protoc absent from PATH** — downloads, verifies, extracts, runs, compiles the protos.

**The failure paths, which are the point:**

```
kwaai-rpc build: protoc archive from ... does not match its pinned sha256.
  expected 0000000000000000...
  actual   92ceefda6a7293ec...  (2241732 bytes)
A short file means the download was interrupted — re-run the build.

kwaai-rpc build: PROTOC is set to "/bin/false", but running it with --version failed.
```

A valid `$PROTOC` is used with **no download**. `fmt`, `clippy --all-targets -D warnings`, and tests all clean.

## Note

Independent of #163/#167/#168 — branched from `main`, touches only `crates/kwaai-rpc/build.rs`.

### DIFF
--- a/core/crates/kwaai-rpc/build.rs
+++ b/core/crates/kwaai-rpc/build.rs
@@ -9,8 +9,139 @@ use std::env;
 use std::path::{Path, PathBuf};
 use std::process::Command;
 
+/// protoc 28.3 release archives, sha256. Pinned per platform for the same
+/// reason `core/patches/*.sh` pin theirs: an unverified download that arrives
+/// short is indistinguishable from one that arrived whole until something
+/// downstream fails for an unrelated-looking reason.
+const PROTOC_VERSION: &str = "28.3";
+const PROTOC_SHA256: &[(&str, &str)] = &[
+    (
+        "win64",
+        "ce64f49bdeddef49ce4bd313a8f59bcf92fcf67b5831efbf66170386d2e66948",
+    ),
+    (
+        "osx-aarch_64",
+        "92ceefda6a7293ec014e6ecac82d64719357145cb6fc2865badadeb5e62c0431",
+    ),
+    (
+        "osx-x86_64",
+        "97fe5d442090b4dbc23cd1384fb9b444fa1dc6e67d15bb5e1fe4de0da7638b20",
+    ),
+    (
+        "linux-x86_64",
+        "0ad949f04a6a174da83cdcbdb36dee0a4925272a5b6d83f79a6bf9852076d53f",
+    ),
+    (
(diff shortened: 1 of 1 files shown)


dazwin:
Closing — I opened this without asking.

The fix itself stands and is verified (checksum-pinned protoc, `curl --fail`, `$PROTOC` honoured, SHA-256 checked against the NIST vectors and both archives, download and both failure paths exercised end to end). The branch `fix/protoc-bootstrap-verification` is pushed and unchanged.

Reopen or re-raise it whenever @dazwin wants it reviewed.

PR #176, opened 2026-09-02
fix(node): a Mac serving whole models is ONLINE, not JOINING

### LINKED ISSUE #175: Macs serving via the Ollama stopgap announce JOINING forever — invisible on the map and excluded from inference
## Summary

A Mac serving whole-model inference through the approved Ollama stopgap announces
`state = 0` (JOINING) forever. The map renders that as **offline**, and — the part that
matters more — `shard run` filters on `state == 2`, so the node is **excluded from
inference selection** while it is in fact serving.

Every Apple-silicon contributor on the sanctioned path is invisible to the network it is
contributing to.

## Root cause

`announce_state()` gates ONLINE on a sentinel that only the block-sharding path writes.

- `kwaai-cli/src/config.rs:1017` — `announce_state()` returns `2` if
  `ShardManager::shard_is_ready()`, else `0`.
- `kwaai-cli/src/daemon.rs:396` — `shard_is_ready()` is
  `ready_file().exists() && is_running()`, where `ready_file()` is
  `~/.kwaainet/run/shard.ready`.
- `kwaai-cli/src/shard_cmd.rs:665` — the **only** place that writes it, on the
  block-sharding path.
- `kwaai-cli/src/shard_cmd.rs:272-307` — the macOS whole-model path registers the Ollama
  proxy handler, prints *"This node does not advertise a block range — by design on
  macOS"*, and waits on ctrl-C. **It never writes the sentinel.**

The gate encodes "ready" as "has a loaded block shard". That stopped being the only way to
serve when the Metal stopgap (#117) landed, and nothing updated the gate.

`config.rs:1014` states the consequence explicitly, in the comment above the function:

> ONLINE means "this node will serve inference", not merely "this node is up" —
> `shard run` filters on `state == 2`

By that definition a Mac serving over `/kwaai/ollama-proxy/1.0.0` **is** online. The code
says otherwise.

## Evidence

`map.kwaai.ai/api/v1/state`, this machine:

```json
{
  "peer_id": "12D3KooWAourfFoxBjBiXWHdZjxPAuenuyYLFjxHa1C9zknjYA7A",
  "state": "offline",
  "public_name": "rezarassool-macos-aarch64/v0.6.6",
  "version": "kwaai-0.6.6",
  "throughput": 36.23,
  "start_block": 0, "end_block": 32,
  "using_relay": true
}
```

The record is being published correctly — the map knows the version and the throughput.
Only the state is wrong. Meanwhile the node is connected to both bootstraps (`direct`) and
its shard log reads:

```
✅ Ollama ready — serving: llama3.1:8b, …
✅ Serving whole-model inference over /kwaai/ollama-proxy/1.0.0
💡 This node does not advertise a block range — by design on macOS.
```

`~/.kwaainet/run/shard.ready` is absent, as expected on this path.

Across the fleet the pattern holds — recent Macs offline, the one pre-stopgap Mac online:

| node | version | map state |
|---|---|---|
| `rezarassool-macos-aarch64` | 0.6.6 | offline |
| `gandalf_macosx64` | 0.6.5 | offline |
| `projxai-macos-aarch64` | 0.4.24 | **online** — predates the stopgap |

## Reproduce

1. On an Apple-silicon Mac with Ollama running, start the node (`run-node` registers
   `/kwaai/ollama-proxy/1.0.0` on its own).
2. `test -f ~/.kwaainet/run/shard.ready` → absent.
3. Query `map.kwaai.ai/api/v1/state` for this peer → `"state": "offline"`.
4. Confirm it is nonetheless serving: the proxy protocol is registered and answers.

## A second inconsistency, same area

The announcement carries `start_block: 0, end_block: 32` on a path that explicitly says it
advertises no block range. If "0-32" is meant as "the whole model", that is defensible, but
it is not what the log claims, and a client selecting by block range cannot tell this node
from a real 32-block shard.

## Possibly related, not yet confirmed

Several non-Mac nodes are also offline (`metro-win/v0.6.6`, `anonymous-windows/v0.6.5`,
`john-linux-vovin/v0.6.5`, `john-linux-naga/v0.6.5`, `christophe-linux/v0.6.5`), where
block sharding *is* viable and this explanation does not apply.

Hypothesis, needing separate evidence: `daemon.rs:469` in `spawn_shard_child()` removes
`shard.ready` **unconditionally**, while both it and its caller (`node.rs:1473`) guard
stopping/adopting the shard on `is_running()`, which reads a shard pid file. On this machine
that pid file does not exist and `kwaainet shard serve` has been orphaned since Aug 30 — so
a restart can delete the sentinel while being unable to see or restart the process that
would rewrite it. That would strand any node whose `run-node` restarted while its shard kept
running, which is what an auto-update does.

## Suggested direction

Make readiness mode-aware rather than overloading `shard.ready`, since the two modes differ
in whether a block range exists and downstream selection depends on that. Either a second
sentinel written when the Ollama proxy registers, or `announce_state()` consulting the
registered protocol set.

Overloading `shard.ready` from the macOS path would be a smaller diff but makes
`shard_is_ready()` true for a node with no block shard, which other callers may not expect.

RezaRassool:
Closes #175.

## The bug

`announce_state()` gated ONLINE on `shard.ready`, which only the block-sharding path
writes. A Mac serves whole models over `/kwaai/ollama-proxy/1.0.0` and deliberately loads
no block shard, so it never wrote that sentinel and announced **JOINING for as long as it
ran**.

Not just a map problem. `config.rs`'s own comment notes that `shard run` filters on
`state == 2`, so those nodes were **excluded from inference selection** while serving
perfectly well — every contributor on the sanctioned Apple path invisible to the network
they were contributing to.

Observed on this fleet: `rezarassool-macos-aarch64/v0.6.6` and `gandalf_macosx64/v0.6.5`
both offline, while `projxai-macos-aarch64/v0.4.24` — predating the stopgap — shows online.

## The fix

Readiness is **mode-aware** rather than overloading `shard.ready`. A second sentinel,
`whole_model.ready`, records the other way a node can serve; `announce_state()` accepts
either.

Keeping them separate is the point: `shard.ready` means "a block shard is loaded" and its
callers act on the block range that implies, which a whole-model node does not have. A
test pins that distinction.

Three properties worth calling out:

- **Re-derived, not latched.** The sentinel is refreshed from a live `ollama::readiness()`
  probe before every announce, so a node that loses Ollama retracts the claim instead of
  advertising a dead path until restart.
- **Derived before the first announce too**, so a Mac already serving comes up ONLINE
  rather than sitting at JOINING for a full re-announce interval. Both the p2pd and native
  paths are wired.
- **No liveness check on this one**, unlike `shard_is_ready`. The block shard is a separate
  process that can die behind its pid file; whole-model serving is registered by the
  announcing process itself, so reaching that code is the liveness proof.

Gated on macOS to match the condition that selects the whole-model path at all
(`cfg!(target_os = "macos") && !force_blocks`) — marking it elsewhere would promise a block
range the node does not have.

## Testing

Two regression tests, **both verified to fail against the old gate**:

```
whole_model_serving_announces_online ... FAILED
  assertion `left == right` failed: a node serving whole models over the
  Ollama proxy is ONLINE
```

1250 tests pass, clippy and fmt clean.

## Not addressed here

Two things from #175 are deliberately left out of this PR:

- The announcement still carries `start_block: 0, end_block: 32` on a path that says it
  advertises no block range. Either that means "the whole model" and the log is misleading,
  or it is wrong — but it is a separate decision about what consumers should see.
- The non-Mac offline nodes (`metro-win`, several Linux) have a different cause. The
  hypothesis in #175 — `spawn_shard_child()` deleting `shard.ready` unconditionally while
  unable to see an orphaned shard process — still needs evidence.


RezaRassool:
## Confirmed end to end on a live node

Built this branch, stopped the running node, and started it on the branch binary against the
real `~/.kwaainet`. Same peer, same identity, production DHT.

**The sentinel is written at startup**, before the first announce — which is the half of the
fix with no unit coverage:

```
$ test -f ~/.kwaainet/run/whole_model.ready && echo present
present
```

```
16:52:18.996  announce state changed reachability=Private announceable=true epoch=1
16:52:19.001  [2/4] Preparing the DHT announcement...
16:52:19.811  [3/4] Announcing to DHT...
16:52:19.897  STORE response from QmQhRu…KhaJc: 32/32 stored
16:52:20.151  STORE response from Qmd3A8…6RZJY: 32/32 stored
```

**And the map flipped on the first crawl after it** (`update_period` is 60s):

```
16:52:18  state=offline  thr=0.0    cache=0        <- tombstone from `kwaainet stop`
16:53:30  state=online   thr=31.92  cache=100000   <- first crawl after the restart
```

Worth being precise about why that is evidence rather than coincidence. `span.server_info.state`
is the **node's own published field**, not a liveness judgement the map makes — so `online`
means `announce_state()` returned `2`. And `throughput`/`cache_tokens_left` came back as live
values instead of the tombstone zeroes, so it is a fresh announcement rather than a cached
record.

The 16:52:18 row also explains the one-crawl delay: that snapshot completed 0.5s *before* the
announce landed, and it carries the `throughput: 0.0, cache_tokens_left: 0` shape that
`build_unannounce_records` writes — so it was faithfully recording the graceful stop, not a
failure of the fix.

### Verification now covers

| | |
|---|---|
| `announce_state()` reads the sentinel | 2 unit tests, both confirmed to fail on the old gate |
| The two sentinels stay distinct | unit test |
| The real `ollama::readiness` probe writes it | live test — 10 models on this box |
| Losing Ollama retracts the claim | live test — `announce_state()` back to `0` |
| A real node flips on the real map | this comment |

PR #181, opened 2026-09-02
fix(patches): untrack the vendored libp2p-kad copy so its checksum is verified again

### LINKED ISSUE #180: core/patches/libp2p-kad/ is tracked despite being gitignored, and its committed stamp makes fetch-patches.sh skip the checksum verification
## Summary

`core/patches/libp2p-kad/` — 31 files of third-party crate source — is **tracked on `main`**
even though `.gitignore:112` lists that directory. Because the tracked files include
`.kwaai-patch-stamp`, a clean clone looks already-patched, so `fetch-libp2p-kad.sh` exits
before it ever downloads or checksums the upstream tarball. The committed copy becomes the
source of truth, unverified.

The sibling patched crate is unaffected: `core/patches/multistream-select/` has **0** tracked
files, which is the intended state. Only libp2p-kad leaked.

## How it got in

`acb791d7` — *"fix(node): a Mac serving whole models is ONLINE, not JOINING (#176)"*. The
files are unrelated to that change; they are somebody's local `fetch-patches.sh` output swept
into the commit. At that point `core/patches/libp2p-kad/` was not yet gitignored — the ignore
rule arrived later with #167, and `.gitignore` does not untrack files that are already
tracked. So the 31 files are now invisible to `git status` while still in the tree, which is
why this went unnoticed.

Verified `05d69255` (current `main`); pre-merge `84ab99a1` already had them. **Not** introduced
by #167.

## Why it matters: the checksum gate is bypassed

`fetch-libp2p-kad.sh:33-38`:

```sh
STAMP="$DEST/.kwaai-patch-stamp"
want_stamp="$VERSION $(sha256 "$PATCH" | cut -d' ' -f1)"
if [ -f "$STAMP" ] && [ "$(cat "$STAMP")" = "$want_stamp" ]; then
    exit 0
fi
```

The stamp is derived from `VERSION` and the hash of `libp2p-kad.patch` — **both tracked**. So
on any fresh clone the stamp matches by construction:

```
$ git show HEAD:core/patches/libp2p-kad/.kwaai-patch-stamp
0.48.0 2cd6ec22d76360a88cb8a95a48599598e55357cf672bc6b7a70a9ad402ee72f6

$ # what the script computes
0.48.0 2cd6ec22d76360a88cb8a95a48599598e55357cf672bc6b7a70a9ad402ee72f6

$ bash core/patches/fetch-libp2p-kad.sh
$ echo $?
0        # exited early — never fetched, never verified
```

`SHA256=13d3fd63…` on line 29 — the pin on the upstream `.crate` tarball, and the whole point
of the checksum-pinned-patch design described in `core/patches/README.md` — is therefore never
evaluated on a clean checkout. Whatever source sits in the commit is what gets compiled into
`kwaainet`, with no verification that it corresponds to upstream 0.48.0 plus the reviewed
patch.

**No divergence today.** I diffed the tracked tree against a freshly-fetched, freshly-patched
copy and they are byte-identical. The problem is that nothing enforces that, and nothing would
report it if it stopped being true — a change to those files would arrive as an ordinary diff
in an unrelated PR, in a directory reviewers do not expect to contain reviewable code.

Secondary consequences:

- `nix/crane.nix` deliberately rebuilds the patched crate from the pristine tarball plus the
  patch, precisely because the working-tree copy is supposed to be gitignored ("*a flake build
  reads the git tree, not the working copy*"). The nix path and the cargo path now disagree
  about where the source comes from.
- 31 files of vendored dependency source sit in the repo with no owner and no update path.

## Fix

```sh
git rm -r --cached core/patches/libp2p-kad
```

The `.gitignore` rule is already correct, so nothing else changes; contributors keep their
working-tree copy, `fetch-patches.sh` starts doing its verification again, and the nix and
cargo paths agree.

Worth adding as a guard, since this is the second time output of these scripts has been at
issue: a CI check that fails if anything under `core/patches/*/` (excluding the tracked
`*.patch` and `fetch-*.sh`) is tracked. That turns a silent recurrence into a red build.

Also worth considering: make the stamp check verify the *source*, not just the patch identity,
so a stale or altered tree cannot satisfy it. Hashing the extracted tree — or simply not
trusting a stamp that arrived from git — would close the bypass even if the files reappear.

RezaRassool:
Closes #180.

## What

`git rm -r --cached core/patches/libp2p-kad` — 31 files, no other change. The `.gitignore` rule is already correct, and `--cached` leaves every file on disk, so contributors keep their working-tree copy.

## Why

The directory has been tracked since `acb791d7` (#176), where a local `fetch-patches.sh` output was swept into an unrelated change. #167 later added the `.gitignore` rule, which does not untrack files that are already tracked — so the files stayed while vanishing from `git status`, which is why this went unnoticed.

The real cost is a bypassed integrity check. `.kwaai-patch-stamp` was tracked too, and `fetch-libp2p-kad.sh:33-38` exits early whenever that stamp equals `"$VERSION $(sha256 libp2p-kad.patch)"` — both tracked. On every clean clone the stamp matched by construction, so the script returned before fetching, and `SHA256=13d3fd63…` — the pin on the upstream `.crate` tarball, and the point of the checksum-pinned-patch design in `core/patches/README.md` — was never evaluated. Whatever source sat in the commit got compiled in.

This restores the shape `multistream-select` already has: the `.patch` and `fetch-*.sh` tracked, the materialized crate not. It also puts `nix/crane.nix` back in agreement with cargo — its comment already assumes the working copy is gitignored (*"a flake build reads the git tree, not the working copy"*).

## Verification

- The tracked tree was **byte-identical** to a freshly fetched and patched copy at removal, so nothing built from it was wrong — the gap was that nothing enforced or reported that.
- Simulated a clean clone (removed the directory, ran `fetch-patches.sh`): it now fetches and verifies, and produces a tree **byte-identical** to what was committed. So verification is restored with no change to what gets built.
- `cargo check -p kwaai-p2p --features kad-multi-protocol` passes afterwards.

## Follow-ups, not in this PR

Suggested in #180 and worth doing separately: a CI check that fails if anything under `core/patches/*/` is tracked apart from `*.patch` and `fetch-*.sh`, and making the stamp verify the *source* rather than just patch identity — or not trusting a stamp that arrived from git — so the bypass stays closed even if the files reappear.

https://claude.ai/code/session_01HxjE2xfQzk11pqyrwgSLZu


dazwin (approved):
Approving — every claim checks out under independent verification, and this closes a hole of my making.

**Verified:**
- Tracking began at `acb791d7` (#176), exactly as diagnosed — and the provenance is worth owning: the materialized directory came from my `fetch-patches.sh` runs in the shared checkout while preparing #167; at the time #176 merged, `main` didn't yet have #167's `.gitignore` rule, so the sweep was easy to make and invisible afterward. My fetched files, someone else's `git add -A`, nobody's fault individually — which is exactly why your follow-up CI guard is the right idea.
- The PR is pure untracking: 31 paths, all under `core/patches/libp2p-kad/`, **zero added lines** anywhere.
- The tracked tree is **byte-identical** to a freshly fetched-and-patched copy (compared against an independently produced one, stamp aside) — corroborating that nothing wrong was ever built, and that the entire cost was the silently bypassed pin.
- The bypass mechanism reads as you describe: the stamp encodes `VERSION + sha256(patch)`, both of which were themselves tracked, so on a clean clone the early-exit was satisfied by construction and `SHA256=13d3fd63…` — the one line that pins upstream — never executed. A checksum design that trusts a checkpoint delivered by the same channel it's meant to verify. That one's on me, and your follow-up (don't trust a stamp that arrived from git, or verify the source rather than patch identity) is the correct durable fix — happy to take it.
- Post-untrack clean-clone behavior: fetch runs, pin verifies, build passes — reproduced here earlier today from empty state several times.

CI still pending on ubuntu/windows at review time, but an index-only change can't alter what those jobs compile — they fetch the patches themselves, which after this merge means they finally *verify* them.

Issue #179, opened 2026-09-02
A relay-only node announces block coverage with no dialable address — peers pin it, then pay a dial timeout and lose the KV-cache


RezaRassool:
## Summary

A relay-only node can announce block coverage to the DHT and be scored `VERIFIED`
while publishing **no dialable addresses at all**. Peers discover it, pin it into a
circuit, then fail to dial it — burning a 60s timeout and losing the KV-cache on
path rebuild. The node is a black hole in the routing table rather than simply
absent, which is worse than being offline.

Observed on `metro-win-rtx-a5000/v0.6.7` (`12D3KooWLMizEbViSoL4WGJUMsLVRyLccyymosX36MDKdbYgGFzE`)
on 2026-09-02, reproduced from a v0.6.7 node. Not specific to that host — any
`using_relay: true` node that fails to publish a circuit address should do this.

## Evidence

The map and DHT both advertise it as healthy, with an empty address list:

```json
{
  "peer_id": "12D3KooWLMizEbViSoL4WGJUMsLVRyLccyymosX36MDKdbYgGFzE",
  "state": "online",
  "using_relay": true,
  "start_block": 0,
  "end_block": 32,
  "throughput": 28.86271105135889,
  "version": "kwaai-0.6.7",
  "multiaddrs": []
}
```

`shard chain` lists it as a full-coverage, trusted server:

```
    8 server(s) — 32/32 blocks covered
      0     32  12D3KooWLMizEbVi…  VERIFIED   metro-win-rtx-a5000/v0.…
```

But every path that needs to *dial* it fails identically:

```
$ kwaainet p2p peers find 12D3KooWLMiz… --timeout 40
  ⚠️  not found in DHT (timeout: 40s)

$ kwaainet p2p peers connect --peer 12D3KooWLMiz…
  ❌ --peer DHT lookup failed: dht_find_peer

$ kwaainet shard run "…" --inference-url mux://12D3KooWLMiz…
  WARN inference-mux stream_open_raw failed …: Dial failed: peer not found in DHT (no addresses)
  Error: Ollama returned 502 Bad Gateway
```

The block-sharding path shows the real cost, because discovery succeeds and only
the dial fails:

```
$ kwaainet shard run "Name one primary color. One word only." --name-filter metro-win --max-tokens 24
  Discovering block circuit from DHT…1 node(s)
[...]

PR #182, opened 2026-09-02
fix(announce): tell the truth about a node that is up but not serving

dazwin:
## What

`KwaaiNetConfig::announce_state()` decides the `state` field every node writes into its DHT records. Two things were wrong with it, and this branch fixes both:

1. **A node serving whole models announced JOINING forever.** An Apple-silicon node serves over `/kwaai/ollama-proxy/1.0.0` and deliberately loads no block shard, but the gate only looked at the block-shard sentinel. Those nodes were off the map and out of `shard run`'s candidate set while serving perfectly well. `announce_state()` now also accepts whole-model readiness, and `refresh_whole_model_ready` retracts the claim when Ollama goes away.

2. **"Not serving" was announced as `0`, which is petals' OFFLINE.** Petals' enum is `OFFLINE = 0, JOINING = 1, ONLINE = 2`, and petals itself uses `OFFLINE` as its *shutdown* announcement (`server.py:542`, `:650`) — so a node that was up, reachable and re-announcing on a timer published the same value as one that had gone away. The map rendered it "offline", indistinguishable from KwaaiNet's own `-1` departure tombstone. It now announces `1` (JOINING), which is what the doc comment already claimed it did.

Also here: `announce_online_without_shard`, an off-by-default config key that forces ONLINE for topology tests, which serve no inference and would otherwise show every node as dead on the map; `announce_state()` moved onto `KwaaiNetConfig` so the p2pd and native announce paths cannot drift on what ONLINE means; and test serialisation for the `KWAAINET_HOME` env var these tests move around.

## Behaviour

No routing or selection behaviour changes with `0` -> `1`. Every consumer in the stack filters on `state == 2`: `shard run`'s candidate sets and chain decoder, the rebalancer, both map generations' block coverage, the node health monitors. The GUI never sees the field (`BlockPeer` carries no state).

One consumer read it as a *threshold* rather than an equality — a python petals server's `block_selection`, which counts everything `>= JOINING` — so a non-serving node would be weighed when such a server places its own blocks. No `petals.cli.run_server` is deployed since the native migration of the bootstraps and maps, so this costs nothing today; it is recorded in the doc comment so a future one does not surprise anyone.

Consumers reading the *map's* JSON are unaffected — `/api/v1/state` still carries `online|joining|offline`. The KwaaiNetMap side of this (rendering those as Sharding / Ready / Departed, and fixing `reachability_issues` to mean failed dials rather than non-ONLINE announcements) is a separate change in that repo.

## Checks

`cargo fmt --all --check`, `cargo clippy --all-targets -D warnings`, `cargo test -p kwaainet` (251 passed) — all clean on the rebased branch.

### DIFF
--- a/core/crates/kwaai-cli/src/config.rs
+++ b/core/crates/kwaai-cli/src/config.rs
@@ -298,7 +298,7 @@ pub struct KwaaiNetConfig {
     pub dht_replication: usize,
     /// Announce `state=2` (ONLINE) even when no shard server is running.
     ///
-    /// Off by default, and deliberately so. A node announces `state=0`
+    /// Off by default, and deliberately so. A node announces `state=1`
     /// (JOINING) until `shard serve` reports its model loaded, because
     /// `shard run` filters on `state==2`: a node claiming ONLINE without an
     /// `/kwaai/inference/1.0.0` handler gets dialled and fails the session
@@ -1024,7 +1024,12 @@ impl KwaaiNetConfig {
         self.native_p2p == Some(false)
     }
 
-    /// The `state` field for a DHT announcement: `2` ONLINE, `0` JOINING.
+    /// The `state` field for a DHT announcement: `2` ONLINE, `1` JOINING.
+    ///
+    /// Not-serving is `1`, not `0`: petals' enum is `OFFLINE = 0, JOINING = 1,
+    /// ONLINE = 2`, and a node reaching this code is up and announcing, so `0`
+    /// (which the map rendered as "offline", indistinguishable from the `-1`
+    /// departure tombstone) was a lie about a healthy node.
     ///
     /// ONLINE means "this node will serve inference", not merely "this node is
     /// up" — `shard run` filters on `state == 2`, so a node that claims it
@@ -1051,7 +1056,7 @@ impl KwaaiNetConfig {
         }
         match Self::load_or_create() {
             Ok(cfg) if cfg.announce_online_without_shard => 2,
-            _ => 0,
+            _ => 1,
         }
     }
 
--- a/core/crates/kwaai-cli/src/config.rs
+++ b/core/crates/kwaai-cli/src/config.rs
@@ -1509,18 +1509,19 @@ mod native_p2p_tri_state {
     }
 }
 
+/// Serialises the tests that drive `KWAAINET_HOME`, crate-wide.
+///
+/// The var is process-global. These tests each point it at their own tempdir
+/// and some remove it when done, so run concurrently one test reads another's
+/// directory — or none at all — and sees the wrong config back. One lock for
+/// the whole crate, not one per module: the racing tests live in different
+/// files (`config`, `daemon`), and per-module mutexes never serialise those.
+#[cfg(test)]
+pub(crate) static HOME_ENV_LOCK: std::sync::Mutex<()> = std::sync::Mutex::new(());
+
 #[cfg(test)]
 mod tests {
     use super::*;
-
-    /// Serialises the tests that drive `KWAAINET_HOME`.
-    ///
-    /// The var is process-global. These tests each point it at their own
-    /// tempdir and some remove it when done, so run concurrently one test
-    /// reads another's directory — or none at all — and sees the wrong
-    /// config back. Identical block on every branch that adds such a test,
-    /// so the copies merge as a duplicate rather than a conflict.
-    static HOME_ENV_LOCK: std::sync::Mutex<()> = std::sync::Mutex::new(());
 
     fn cfg(start: u32, blocks: u32, total_hint: &str) -> KwaaiNetConfig {
         KwaaiNetConfig {
--- a/core/crates/kwaai-cli/src/config.rs
+++ b/core/crates/kwaai-cli/src/config.rs
@@ -1031,6 +1031,12 @@ impl KwaaiNetConfig {
     /// (which the map rendered as "offline", indistinguishable from the `-1`
     /// departure tombstone) was a lie about a healthy node.
     ///
+    /// Every consumer that survives the native migration filters on `== 2`. The
+    /// one that read this as a threshold — a python petals server's
+    /// `block_selection`, counting everything `>= JOINING` — would weigh a node
+    /// that may never load a shard; it matters again only if a
+    /// `petals.cli.run_server` ever rejoins the swarm.
+    ///
     /// ONLINE means "this node will serve inference", not merely "this node is
     /// up" — `shard run` filters on `state == 2`, so a node that claims it
     /// without anything to serve gets dialled and fails the session with
--- a/core/crates/kwaai-cli/src/config.rs
+++ b/core/crates/kwaai-cli/src/config.rs
@@ -296,6 +296,21 @@ pub struct KwaaiNetConfig {
     /// when `decentralized_dht` is false.
     #[serde(default = "default_dht_replication")]
     pub dht_replication: usize,
+    /// Announce `state=2` (ONLINE) even when no shard server is running.
+    ///
+    /// Off by default, and deliberately so. A node announces `state=0`
+    /// (JOINING) until `shard serve` reports its model loaded, because
+    /// `shard run` filters on `state==2`: a node claiming ONLINE without an
+    /// `/kwaai/inference/1.0.0` handler gets dialled and fails the session
+    /// with "protocols not supported".
+    ///
+    /// Turn it on only where that cannot happen — a topology test exercising
+    /// NAT traversal, relays and hole punching, where no inference is served
+    /// and no client will route to these nodes. It exists because such a
+    /// topology otherwise shows every node as offline on the map, which
+    /// obscures the connectivity it is there to test.
+    #[serde(default)]
+    pub announce_online_without_shard: bool,
 
     #[serde(default)]
     pub health_monitoring: HealthConfig,
@@ -902,6 +917,7 @@ impl Default for KwaaiNetConfig {
             announce_self: true,
             decentralized_dht: false,
             dht_replication: default_dht_replication(),
+            announce_online_without_shard: false,
             health_monitoring: HealthConfig::default(),
             model_dht_prefix: None,
             model_repository: None,
@@ -1239,6 +1255,9 @@ impl KwaaiNetConfig {
                 }
                 self.dht_replication = k;
             }
+            "announce_online_without_shard" => {
+                self.announce_online_without_shard = parse_bool(value)?
+            }
             "start_block" => {
                 self.start_block =
                     Some(value.parse().map_err(|_| {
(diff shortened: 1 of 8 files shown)


core/crates/kwaai-cli/src/config.rs
```diff
@@ -1022,14 +1049,31 @@ impl KwaaiNetConfig {
...
+            Ok(cfg) if cfg.announce_online_without_shard => 2,
+            _ => 1,
         }
     }
 
+    /// Whether a block shard is mid-load: `shard serve` is alive but has not
+    /// written its ready sentinel. Both this and "up, serving nothing" announce
+    /// JOINING, and only the node can tell them apart — so it says which, and
+    /// the map stops implying progress that may never come.
+    pub fn announce_shard_loading() -> bool {
```
RezaRassool, line 1073: **A macOS node that spawned a shard child announces `shard_loading=true` forever — alongside `state=2`.**

`announce_shard_loading()` reduces to `!shard_is_ready() && shard_child_alive`, and never consults `whole_model_is_ready()`. On macOS, `kwaainet start --shard` (or `contribute.shards: true` with a local model) spawns the child and the parent writes its pid (`main.rs:331`, `:454`, `node.rs:1490`, `:2542`), but that child takes the whole-model branch (`shard_cmd.rs:339` → `serve_whole_model_via_ollama`), which writes no ready sentinel and blocks on `ctrl_c().await` (`shard_cmd.rs:307`). Meanwhile `refresh_whole_model_ready` writes `whole_model.ready`, so `announce_state()` returns `2`.

I verified each link rather than inferring: `shard_is_ready()` = ready-file ∧ `is_running()` → false; `is_running()` → true; `whole_model_is_ready()` → true.

Net effect: every announce carries `state=2` **and** `shard_loading=true`, permanently. The map shows a forever-"loading" shard on a node that has nothing to load and is serving fine — precisely the ambiguity this field was added to remove, re-created on the platform this PR exists to fix.

`to_msgpack` (`announce.rs:255`) also emits the key regardless of `state`, which contradicts its own doc ("Only meaningful while `state` is JOINING").

Gating on `state == 1` at the encode site, and/or adding `&& !whole_model_is_ready()` to the predicate, would close it.

core/crates/kwaai-cli/src/config.rs
```diff
@@ -1022,14 +1049,31 @@ impl KwaaiNetConfig {
...
+    /// Lives beside the flag it reads, and reachable from both the p2pd and the
+    /// native announce paths, so the two cannot drift on what ONLINE means.
     pub fn announce_state() -> i32 {
         if ShardManager::shard_is_ready() || ShardManager::whole_model_is_ready() {
-            2
-        } else {
-            0
+            return 2;
+        }
+        match Self::load_or_create() {
```
RezaRassool, line 1063: **This reads `config.yaml` from disk on every announce, and any transient read/parse failure silently downgrades the node to JOINING.**

The callers already hold the config — `node_native.rs:794` is literally `refresh_server_info(&mut server_info, &config)` and ignores it — yet the flag comes from `Self::load_or_create()`.

`KwaaiNetConfig::save()` uses `std::fs::write`, which truncates before writing, so a reader racing a writer sees a partial file. Concretely: `shard serve`'s auto-assign persists its block range mid-run, or the operator runs `kwaainet config set`, at the moment an announce round fires → `serde_yaml` parse error → the `_ => 1` arm → a node configured `announce_online_without_shard: true` announces JOINING for that round. At 300s ± 30s jitter it then reads as not-serving on the map for up to five minutes.

The `_ =>` arm is what makes it silent: a config error and "flag is false" are indistinguishable, so there's no log to explain the flap.

It also puts blocking file I/O on the async announce path — plus a possible `create_dir_all`/`save()` if the `initial_peers` migration triggers.

Reading `config.announce_online_without_shard` from the value the caller already has avoids all of it.

core/crates/kwaai-cli/src/config.rs
```diff
@@ -296,6 +296,21 @@ pub struct KwaaiNetConfig {
...
+    /// `/kwaai/inference/1.0.0` handler gets dialled and fails the session
+    /// with "protocols not supported".
+    ///
+    /// Turn it on only where that cannot happen — a topology test exercising
+    /// NAT traversal, relays and hole punching, where no inference is served
+    /// and no client will route to these nodes. It exists because such a
+    /// topology otherwise shows every node as offline on the map, which
+    /// obscures the connectivity it is there to test.
+    #[serde(default)]
+    pub announce_online_without_shard: bool,
```
RezaRassool, line 313: **This flag is guarded only by a doc comment, and a node running it poisons the shared discovery registries.**

The doc says to turn it on only where no client will route to these nodes, but nothing enforces that. A topology test exercising NAT traversal or relays runs against the production bootstraps with the default `dht_prefix`, and `node.rs:1262` announces the same `state=2` `DHTServerInfo` bytes into `_kwaai.inference.nodes` as well as the per-block keys.

`discover_inference_peer` (`shard_cmd.rs:2412`) accepts any `state == 2` peer and then picks the **highest announced throughput** — and these nodes announce a throughput while serving nothing. So an unrelated user's `rag chat --inference p2p://auto` or `shard run` will preferentially select the test node and fail with "protocols not supported": exactly the failure the state gate exists to prevent, made more likely by the throughput tiebreak.

Tying the flag to a non-default `dht_prefix`, or suppressing the `_kwaai.inference.nodes` announcement when it is set, would make the intent structural rather than advisory.

core/crates/kwaai-cli/src/shard_cmd.rs
```diff
@@ -329,8 +329,8 @@ async fn cmd_shard_serve(args: ShardServeArgs) -> Result<ShardServeExit> {
     //
     // So a Mac does not register as a block server. It serves whole-model
     // inference over `/kwaai/ollama-proxy/1.0.0`, which every node already
-    // registers, and stays at announce state 0 ("online, no shard") because no
-    // shard ever loads — no announce change needed.
+    // registers, and stays at announce state 1 (JOINING) because no shard ever
```
RezaRassool, line 332: This comment is now factually wrong, and it was touched by this diff.

"stays at announce state 1 (JOINING) because no shard ever loads — no announce change needed" was mechanically renumbered `0` → `1`, but since #175 and this PR a macOS node writes `whole_model.ready` and announces `2` (ONLINE).

A maintainer reading it concludes macOS nodes never reach ONLINE — the exact bug the rest of the PR fixes.

core/crates/kwaai-cli/src/config.rs
```diff
@@ -1494,6 +1551,33 @@ mod tests {
...
+
+    #[test]
+    fn announce_online_without_shard_round_trips() {
+        let _env_lock = HOME_ENV_LOCK.lock().unwrap_or_else(|e| e.into_inner());
+        // `set_key` persists to $KWAAINET_HOME. Without our own tempdir this
+        // writes into whichever directory another test has pointed the var
+        // at — and fails with ENOENT when that test's TempDir drops mid-run.
+        // Harmless on this branch alone; a ~1-in-6 flake once `combined` puts
+        // it alongside the bootstrap-serve tests that drive the same var.
+        let home = tempfile::tempdir().expect("tmpdir");
```
RezaRassool, line 1571: **This test leaks `KWAAINET_HOME` pointing at a deleted directory, which silently disarms the sandbox guard for later tests.**

`announce_online_without_shard_round_trips` sets `KWAAINET_HOME` to a `tempfile::tempdir()` and never removes or restores it — unlike `daemon.rs`'s tests, which `remove_var`, and `grpc_server`'s `EnvGuard`, which restores on drop. When the `TempDir` drops, the directory is deleted but the var stays set for the rest of the test binary.

The consequence is quiet: the `#[cfg(test)]` sandbox in `kwaainet_dir()` (`config.rs:16-40`) is bypassed for every later unguarded test, and `test_isolation::config_path_is_sandboxed_under_test` / `set_key_does_not_touch_the_disk` hit their `if std::env::var("KWAAINET_HOME").is_ok() { return; }` early-outs and assert nothing. So the guard protecting a developer's real config stops being verified — intermittently, depending on test order.

Separately, the stated rationale is wrong: the comment says "`set_key` persists to $KWAAINET_HOME", but `set_key` (`config.rs:1245-1327`) performs no I/O at all, and `set_key_does_not_touch_the_disk` exists specifically to assert that. Neither the tempdir nor the lock is needed here for the reason given.

RezaRassool (commented):
## Review

Traced the announce path (`announce.rs`, `node.rs`, `node_native.rs`), the state consumers, the sentinel logic (`daemon.rs::ShardManager`, `refresh_whole_model_ready`) and the test-harness changes, and independently re-verified the two findings that carry real behaviour.

**The core `0 → 1` change checks out.** Every in-repo consumer filters on `state == 2` (`shard_cmd.rs:2412`, `:2555`, `:2650`), `docs/PETALS_PROTOCOL_COMPLETE.md:286` already documents `0=OFFLINE, 1=JOINING, 2=ONLINE`, and `examples/query_dht_state.rs:103,180` already decodes `1 => "joining"`. No Python or JS consumer of the field exists in this repo. Both fixes are the right shape, and disclosing the petals `block_selection` `>= JOINING` threshold consumer yourself — the one case that contradicts the "everything uses equality" claim — made the rest of the body easier to trust.

One finding is a behaviour bug that undercuts the PR's own goal; the rest are a config-read on the announce path, an unenforced footgun, and four stale/leaky bits in comments and tests.

### Confirmed by reading the code, not just the diff

For the first finding I checked each link rather than inferring:

- `serve_whole_model_via_ollama` writes **no** ready sentinel and blocks on `ctrl_c().await` (`shard_cmd.rs:307`)
- `shard_is_ready()` = ready-file ∧ `is_running()` → **false**
- `is_running()` reads the pid and checks liveness → **true** (child alive forever)
- `whole_model_is_ready()` → **true**, so `announce_state()` → **2**

So `state=2` and `shard_loading=true` really are emitted together, indefinitely, on any macOS node that spawned a shard child.

Details inline. Nothing here blocks the `0 → 1` fix, which I'd like to see land — the first finding is worth fixing in this PR because it re-creates the ambiguity `shard_loading` exists to remove.

---

### Two more, outside the diff so I can't anchor them inline

**`core/crates/kwaai-cli/src/config.rs:1843` — stale comment, contradicted by this PR.**
The `test_isolation` block explains that no `save()`-then-assert test exists because "`grpc_server`'s tests set `KWAAINET_HOME` process-wide (guarded by a mutex private to that module, so it does not serialise against this one)". This PR made that lock crate-wide (`HOME_ENV_LOCK`, `config.rs:1539`, taken by `EnvGuard::set`), so the stated reason no longer holds — and the comment now steers future work away from a test this PR just made writable.

**`core/crates/kwaai-cli/src/node.rs:2641` — the new crate-wide lock is incomplete.**
`DHTServerInfo::new` now does global-state I/O: `announce_state()` → `load_or_create()` (which can *create* `config.yaml`), and `announce_shard_loading()` → `ShardManager::new()` (which `create_dir_all`s `run_dir()` and reads `shard.pid`/`shard.ready`). `capacity_lease_dht_tests::to_msgpack_announces_lease_v1_true` takes neither `HOME_ENV_LOCK` nor a tempdir, so it now reads and writes whichever `KWAAINET_HOME` a concurrently running `daemon`/`grpc_server` test has installed — including the shard sentinels that `shard_loading_is_only_true_while_a_shard_process_loads` (`daemon.rs:752`) asserts on. It happens not to assert on `state`/`shard_loading` today, so this is latent rather than flaky now, but it is the same race the PR set out to close and the next assertion added there will trip it.

### Checked and clear

- The `-1` tombstone path (`announce.rs:437`, `node.rs:1056`) correctly sets `shard_loading: false`, and both re-announce sites in `node.rs` (SIGHUP at 701, periodic at 832) plus `node_native.rs:794` are updated consistently.
- `EnvGuard`'s new `_home_env` field is a named binding rather than `_`, so the guard is held for the struct's lifetime; drop order keeps the lock held while the env vars are restored; no test constructs two `EnvGuard`s, so the non-reentrant mutex cannot self-deadlock; and there is no lock-order cycle with `TEST_LOCK`, which is private to `grpc_server::tests`.
- `whole_model_is_ready()` deliberately skips the liveness check, and `refresh_whole_model_ready` retracting on Ollama's disappearance is the right counterpart — a stale ONLINE claim after Ollama dies would be the same undialable-announcer shape as #179, and it is handled.

PR #185, opened 2026-09-03
feat(announce): publish signed dial addresses so NATed peers can be reached

dazwin:
## Problem

Distributed-inference dispatch discovers a peer from its DHT record and dials it by bare PeerId. On the native stack that resolves through kad, and kad cannot answer: rust-libp2p serves `FIND_NODE` from its k-buckets alone (`find_closest_local_peers`), where the kad-DHT spec §6.1.1 says a server MUST answer for a requested peer held in its *peerstore* — which is what go-libp2p's `handleFindPeer` did for the old p2pd bootstraps. So a peer is findable only while it holds one of twenty slots in the right bucket on a peer the walk reaches, and a disconnected entry is the first one replaced.

Measured on a live node: **~2000** `Dial failed: <peer>: peer not found in DHT (no addresses)` in one day's log, against a handful of real transport errors. Every peer in a chain build showed as unreachable; the one that connected was one we had spoken to before.

Relaying itself works. Dialing a target through a bootstrap it has *not* reserved on returns `Relay has no reservation for destination` — so the hop is live and the only thing missing is the address, which nothing carried: reservations rotate, and a dialer cannot guess which relay a peer holds one on.

This is **not** about kad's mode. `dht_server` is true on every native node, so kad is pinned to `Mode::Server` whether the node is reachable or not.

## Fix — two layers, the way p2pd had them

**The record carries the address (announce layer).** `DHTServerInfo` gains `signed_addrs`: an RFC 0003 `PeerRecord` in a protobuf `SignedEnvelope`, published under the msgpack key `addrs_signed` as binary. Interop format (`libp2p-peer-record`) rather than rust-libp2p's legacy domain, since this is a published wire format a Go/JS reader may one day verify. Legacy readers ignore the unknown key exactly as they ignore `vpk` and `lease_v1`. Addresses come from the swarm's live listeners — where an accepted reservation appears, already carrying both hops — filtered by `is_announceable` and a new `uses_dialable_transport` (this build is TCP+QUIC only), deduped one per relay, direct before circuit, capped at four because the value is stored under every block key. Published bare; the peer id is in the record once, under the signature. A reservation moving between relays now re-announces (`AnnounceState` fingerprints the confirmed circuit set), so a rotated circuit does not stay published for a full TTL. Encode and wire-decode (`decode_server_info_ext` → `ServerInfoFields`) sit side by side in `announce.rs`; the ONLINE/min-version policy and the mapping to `BlockServerEntry` stay with the consumer in `shard_cmd.rs`.

**The daemon keeps what it learns (network layer).** `kwaai_p2p::peer_record::verified_addrs` verifies the envelope and binds it to the claimed peer. The daemon holds a bounded **learned-addresses map** — the peerstore rust-libp2p does not have — that `dial_routed` consults *before* the k-bucket and the DHT walk, and forgets an address on the dial failure that proves it stale. It is seeded through the control-socket `Connect` request, which now carries addresses the way Go's `AddrInfo{ID, Addrs}` did: no addresses → routed dial by id; one destination-qualified address → the classic dial; anything else → seed and route. So the published addresses help **every** dial to that peer — the hop RPCs through `dispatch_routed` included — not only a pre-connect, and if a connection drops mid-session the re-dial comes from the map without a walk. The CLI's part is one line per pre-connect site: `client.connect_peer_with_addrs(&entry.peer_id, &entry.dial_addrs)`.

### Why signed, and what the signature does and does not close

`kwaai-hivemind-dht` has no record validators: anyone can `STORE` under another node's subkey. A plaintext address list would be an instruction, from anyone, about where to send a victim's traffic. Verification (`from_signed_envelope_interop` + an equality check against the peer being dialed) closes **steering to addresses the peer never signed** — including the dictionary path, where the entry is keyed by subkey and a value naming a different peer is dropped.

It does not close overwriting an announcement to *omit* the field, tombstoning it, or replaying an older genuine list. All three are plain overwrites needing no forged signature, and all three are what a record validator in the DHT store would answer — the next piece of work, not this one.

## Verification

- `cargo test` on `kwaainet`, `kwaai-p2p`, `kwaai-p2p-daemon`: all suites green; `fmt --check` and `clippy --all-targets -D warnings` clean.
- New tests include: forged-signer rejection; subkey/value mismatch (confirmed to fail with the guard removed); circuit-address round trip; legacy-signing-domain rejection; the rotation fingerprint; the learned map's caps, dedupe and forget-on-failure; and two integration tests — a seeded connect reaches a peer the DHT cannot, and **a routed request re-dials from the learned map after a disconnect** with the peer asserted absent from the routing table (confirmed to fail with the map consult removed).
- **End to end on the sealed nat-test bed:** node-f (cone NAT) discovered all 8 nodes and handed each one's published addresses to its daemon — the daemon log shows `connect with supplied addresses … added=2` for node-g (its two bootstrap circuits), zero DHT walks and zero `no addresses` failures for the run. An earlier build of this branch, before the daemon-side move, dialed the same peers from those addresses directly, node-g and node-h **NAT→NAT through a bootstrap circuit**. The bed cannot show the re-dial-from-map property live — eight nodes are all already connected to each other within seconds — so that rests on the integration test above, which asserts the peer is absent from the routing table when the re-dial succeeds.
- Reviewed by independent agents at each step; findings folded in.

## Follow-ups (not in this PR)

1. Consume `PeerInfo.addrs` from kad walk results in `dial_routed` instead of re-reading the k-bucket.
2. Spec §6.1.1: answer `FIND_NODE` from the learned map in the vendored `libp2p-kad` — restores what the Go bootstrap did; needs the bootstraps redeployed.
3. A record validator in the DHT store.

### DIFF
--- a/core/crates/kwaai-cli/src/announce.rs
+++ b/core/crates/kwaai-cli/src/announce.rs
@@ -23,12 +23,14 @@ use anyhow::Result;
 use kwaai_hivemind_dht::protocol::{NodeInfo, RequestAuthInfo, StoreRequest, StoreResponse};
 use kwaai_hivemind_dht::value::get_dht_time;
 use kwaai_p2p::{NetworkHandle, PeerId};
+use libp2p::Multiaddr;
 use prost::Message;
 use sha1::{Digest, Sha1};
 use std::collections::{HashMap, HashSet};
 use tracing::{info, warn};
 
 use crate::config::KwaaiNetConfig;
+use crate::shard_cmd::{version_meets_minimum, BlockServerEntry};
 
 /// TTL applied to every announced record, in seconds.
 ///
@@ -201,12 +203,14 @@ pub struct DHTServerInfo {
     /// the attacker to complete a handshake.
     ///
     /// The envelope closes *that*, and only that.
+    /// [`kwaai_p2p::peer_record::verified_addrs`] is the reader:
     /// `PeerRecord::from_signed_envelope_interop` verifies the signature and
-    /// binds the record's peer id to the signing key, so an address list the
-    /// announced peer did not sign is one the reader drops. The interop format
-    /// (`libp2p-peer-record`) is used rather than rust-libp2p's legacy domain
-    /// because this record is a published wire format that a Go or JS reader
-    /// may one day verify.
+    /// binds the record's peer id to the signing key, and the caller's
+    /// comparison against the announced peer ties it to the peer being dialed
+    /// — so an address list the announced peer did not sign is one the reader
+    /// drops. The interop format (`libp2p-peer-record`) is used rather than
+    /// rust-libp2p's legacy domain because this record is a published wire
+    /// format that a Go or JS reader may one day verify.
     ///
     /// What it does **not** close, because the signature covers the address
     /// list rather than the record: an attacker can still overwrite a peer's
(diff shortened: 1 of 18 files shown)


RezaRassool (commented):
## Review

The signed-envelope design is sound — `PeerRecord::new_interop` binds the record's peer id to the signing key, so the `record.peer_id() != claimed` check is genuinely load-bearing rather than decorative, and the subkey-vs-value `peer_id` guard is safe against legacy Hivemind records because those omit the field entirely and the `is_empty()` exemption covers them.

One finding looks blocking, because it stops the feature working on the nodes it was written for. Findings below were checked against the branch head and, where possible, against a live node rather than by reading alone.

---

### 1. A node's own ULA addresses evict the circuit from the record — high

`is_announceable`'s IPv6 branch (`addresses.rs:59-65`) rejects only `is_unspecified()` and `is_loopback()`. The IPv4 branch right above it rejects link-local, private and CGNAT. That asymmetry means IPv6 **unique local addresses** (`fc00::/7` — including Tailscale's `fd7a::/16` and any `fd00::` autoconf prefix) return `routable_ip = true`.

`node_native.rs:142` sets `listen_addrs: Vec::new()`, so `swarm_listen_addrs()` binds `/ip6/::/tcp/{port}` and the swarm reports one address per interface.

I ran the selection against a live node's actual address list (11 addresses, redacted):

```
pass is_announceable: 6 of 11
  2603:…  global
  2603:…  global
  fd00:…  ULA  — not routable off-network
  2603:…  global
  2603:…  global
  fd7a:…  ULA (Tailscale) — not routable off-network
```

All six are direct, `sort_by_key(is_circuit)` puts direct before circuit, and the `out.len() == MAX_DIAL_ADDRS` break at `announce.rs:561` fills all four slots before any circuit is reached. The dedup key for a direct address is the address itself, so they do not collapse.

Net effect on a NATed node: the relay circuit — the one address that peer can actually be reached on — is dropped from the published record, and up to two of the four published addresses are undialable from outside its LAN.

Adding `is_unique_local()` and `is_unicast_link_local()` to the IPv6 branch fixes the leak. It may be worth reserving a slot for a circuit regardless, so a future address-class oversight degrades rather than silently removes the fallback.

*Note: I could not reproduce the `fe80::` case. The test machine has 12 link-local addresses at OS level and the swarm surfaced none of them, so something below `listen_addrs()` already filters them. ULA is the reachable path.*

---

### 2. `discover_inference_peer` decodes `dial_addrs` and drops them — medium

`shard_cmd.rs:2359` pushes only `(throughput, peer_id, public_name)` and returns `Some(format!("p2p://{}", best_peer.to_base58()))`. The record's `dial_addrs` (decoded at `announce.rs:442`) never reach the learned map, so `p2p://auto` resolves to a bare PeerId and dials with nothing seeded — while the block-chain paths in the same file were migrated to `connect_peer_with_addrs`.

Whether that hard-fails depends on other fallbacks, which I did not trace; what is certain is that this path gets no benefit from the feature, and remote inference against a NATed peer is one of the cases the PR is aimed at.

---

### 3. Verified addresses get no address-class filter on the read side — low/medium

`verified_addrs` proves *who signed*, not *what is sane to dial*. `LearnedAddrs::insert` (`learned_addrs.rs:104-111`) rejects only our own peer id and empty addresses, so a signed record listing `/ip4/127.0.0.1/tcp/25` or `/ip4/10.0.0.1/tcp/22` is stored and dialed by every peer that reads it. Noise stops data exchange; the connection attempt itself is remotely steerable.

The obvious fix is wrong, though: `dial_candidates` passing `|_| true` (`service.rs:1113`) is deliberate, and its own comment says why — *"it keeps loopback and LAN, which is how two nodes on one host reach each other."* Filtering there would break local two-node testing.

The filter belongs at record ingestion, where an address that arrived in a remote peer's signed record can still be told apart from one learned locally.

---

### 4. A UPnP-mapped node publishes no address — low

`signed_dial_addrs` reads only `handle.listen_addrs()` (`announce.rs:505`), and `Command::ListenAddrs` returns `swarm.listeners()` (`service.rs:642`). Confirmed external addresses installed via `swarm.add_external_address` live in `swarm.external_addresses()` and are never consulted — `announce.rs` contains no reference to them. Meanwhile a `Public` verdict releases every relay reservation (`service.rs:2047`).

Severity is limited by two things I checked: `force_private` still defaults `true` and is terminal-Private, so a stock node never reaches `Public`; and an operator-declared address *is* published, because `declared` is chained in at `announce.rs:506`. The gap needs `force_private: false` **and** reliance on UPnP rather than a declared address. Chaining `external_addresses()` in alongside `declared` would close it.

---

### Two design questions rather than defects

Both behaviours are deliberate and documented; they are worth a second look only because this PR changes what they operate on.

- **`forget` on every failed dial** (`service.rs:1825`) — *"Unconditional rather than routed-only — every dial is evidence about the address it used."* Sound for a stale relay reservation. But now that a learned circuit may be a peer's *only* address, one transient failure mid-rotation drops it with nothing to re-seed it until a fresh chain build re-reads the DHT. A failure counter, or re-seeding from the record on routed-lookup failure, would keep the stated intent without the cliff.

- **One attempt per burst** (`service.rs:1220`) — parked requests ride along on an in-flight attempt's outcome. `connect_peer_with_addrs` seeds addresses and then calls `dispatch_routed`, so it can return an older attempt's `peer not found in DHT` despite having just supplied a working address. The addresses survive for a later dial, so this is a spurious error rather than a lost connection — but it will be a confusing one to debug.

---

### Checked and clean

- `LearnedAddrs` keeps `peers` and `order` in step across insert / evict / forget; per-peer and global caps behave as documented.
- `do_connect`'s three-shape dispatch preserves the classic path for every address form `P2PClient::connect_peer` can produce.
- Record size (~400 B signed envelope) is far under `MAX_VALUE_BYTES = 1 MiB`, and `on_new_listen_addr` is idempotent, so reservation renewals do not cause announce churn.
- `PeerRecord::new_interop` / `from_signed_envelope_interop` exist in the pinned libp2p-core 0.43.2 and enforce the record-peer-id ↔ signing-key binding.

dazwin:
Thanks — the live-node run on finding 1 was the useful part. Three commits address the findings; notes on each, and on the two design questions.

**1. ULA evicts the circuit — fixed on both sides.** `select_dial_addrs` now lets circuits claim their slots first and gives direct addresses what is left (still ordered direct-first). Your six-direct case keeps the circuit; a test pins it. The ULA classification itself is fixed where the v6 classifier lives, on #186: `is_routable_v6` no longer has a lenient tier, so unique-local, the documentation and benchmarking prefixes and the transition ranges are rejected outright, and your `fd7a::` address is not published at all. A follow-up PR adds a `require_global_ips` config key (default strict) so a sealed test environment can opt back in, and puts IPv4's RFC2544 allowance under the same key.

**2. `discover_inference_peer` drops `dial_addrs` — fixed.** Candidates carry the record's addresses and the chosen peer is handed to the daemon via `connect_peer_with_addrs` before the URL is returned. Best effort: the URL is still returned if the pre-connect fails.

**3. No address-class filter on the read side — fixed at ingestion, as you suggested.** `verified_addrs` re-runs the publisher's rule: a direct address must be announceable and on a transport this build dials; a circuit is judged by its relay hop, which must not carry an unroutable IP, while a DNS relay hop (how the bootstraps are addressed) passes. Filter runs before the cap, so junk cannot consume the four slots. `dial_candidates` is untouched.

**4. UPnP node publishes nothing — fixed.** New `NetworkHandle::external_addrs` reads `swarm.external_addresses()`, chained in ahead of the listeners.

**Design question: forget on every failed dial.** Left as is, for now. Every consumer re-seeds from the record on each call — the chain paths on every inference, and after finding 2 `p2p://auto` too — so the cliff is bounded to the remainder of one call, and a failure counter would make the stale-reservation case take two dials to clear. If it shows up in practice the re-seed-on-lookup-failure shape is the one I would reach for.

**Design question: one attempt per burst.** Fixed the concrete case. When a spent attempt's dial fails, the service now makes one more dial if the learned map holds addresses that dial did not try — which is exactly what a `connect_peer_with_addrs` parked mid-dial leaves behind. Once per burst, and only on untried addresses, so it cannot loop.

RezaRassool (commented):
## Review — five findings, one worth fixing before merge

2,237 lines, and the security-relevant parts hold up. I read the full 11-commit diff, read the surrounding code in `service.rs`, `relay_manager.rs`, `addresses.rs`, `shard_cmd.rs` and `shard_api.rs`, and verified compilation (`cargo check -p kwaai-p2p -p kwaai-p2p-daemon -p kwaainet --all-targets` — clean).

Spot-checked and found correct, so these needn't be re-litigated:

- `forget`/`retry_on_learned` address normalisation is consistent — `strip_p2p` on both sides matches the fully-qualified address a circuit dial reports.
- Routed-dial bounding is finite: at most 3 dials per burst (initial → post-lookup → learned retry), and `learned_retry_spent` prevents recursion.
- `OwnAddresses::is_ours` exempts circuits, so a peer's circuit through *our* relay isn't filtered out as self.
- `on_new_listen_addr` returns true per-slot rather than only 0→1, so a relay rotation bumps the epoch in both directions.
- `RelayManager::pick_candidate` allows only one slot per relay peer id, so the "TCP+QUIC circuit on one relay" collapse in `select_dial_addrs` cannot lose a node's only circuit.
- The dictionary subkey/value guard does reject the forged case; `verified_addrs` binds the signature to the claimed peer, and the legacy signing domain is rejected.

### Findings

1. **`core/crates/kwaai-cli/src/shard_cmd.rs:2837` — saved circuits reload with no addresses, which is the bug this PR exists to fix.** `SerializableEntry::to_entry()` returns `dial_addrs: Vec::new()`, and its comment ("Rediscovery refills these from the live record") doesn't hold for the paths that call it: `shard_cmd.rs:1254`, `shard_cmd.rs:1806` and `shard_api.rs:791` all load a saved circuit *instead of* running `discover_chain`.

   Concrete failure: `kwaainet shard circuit create`, then `kwaainet shard run --circuit <id>` (or `shard api --circuit <id>`) pre-connects with an empty address list, so `connect_peer_with_addrs` takes the daemon's no-addresses branch → routed dial by bare PeerId → the exact `peer not found in DHT (no addresses)` this PR removes, for every NATed peer in the pinned chain. Either re-read the DHT for the circuit's peers, or persist the addresses in `SerializableEntry`.

2. **`core/crates/kwaai-p2p/src/addresses.rs:168`** — `uses_dialable_transport` unconditionally accepts `/quic-v1`, and its doc asserts the swarm is "TCP and QUIC only (`with_tcp` → `with_quic` → `with_dns`)". But `service.rs:393` gates `with_quic()` on `config.enable_quic`, which defaults **false** (`config.rs:302`). So on a default node a peer's QUIC-only published address is accepted, `LearnedAddrs` stores it, `candidate_addresses` puts it at the *head* of the dial list, and the dial fails with `MultiaddrNotSupported`. For a peer whose record carries only QUIC circuits — the common fleet shape per #162 — the seeded connect fails, `forget` empties the map, and the PR delivers no improvement while the log reports the addresses were supplied.

   This one is entangled with #189: whether it wants a `strict`-style parameter threaded from `enable_quic`, or simply follows from QUIC going back on by default, depends on how #189 resolves. Flagging rather than prescribing.

   Filed alongside #206 and #199 as #208, since all three add a path that admits a remotely-claimed address without the standard filter.

3. **`shard_cmd.rs:2679`** — the new dictionary guard compares base58 **strings** (`info.peer_id_b58 != peer_id_b58`) rather than parsed `PeerId`s. `parse::<PeerId>()` accepts several multibase encodings (`12D3Koo…` and CIDv1 `k51q…` both parse), so a producer that writes one form in the subkey and the other in the value has its entry silently dropped from chain discovery — the peer stops counting for gap detection, and `pick_gap_blocks` sees a hole that isn't there. `peer_id` is already parsed two lines above; `info.peer_id_b58.parse::<PeerId>().ok() != Some(peer_id)` would do it (keeping the `is_empty()` legacy escape).

4. **`core/crates/kwaai-cli/src/announce.rs:586`** — `direct.truncate(MAX_DIAL_ADDRS - circuits.len())` gives circuits unconditional priority, so with `max_relay_reservations >= 4` a private node publishes 4 circuits and **zero** direct addresses, discarding the operator-declared `announce_addr`/`public_ip` — which the function's own docs call "the one address the swarm cannot observe for itself", and which sorts first precisely because it's the cheapest hop. Not reachable on defaults (`max_relay_reservations = 2`), but it's an operator-settable knob and `force_private` defaults true, so every node holds circuits. Reserving one slot for `declared` before truncating would fix it.

5. **`shard_cmd.rs:2308` / `2679` — pre-existing, newly load-bearing.** `servers.entry(key).or_insert(entry)` in `discover_chain` is first-writer-wins across all queried bootstraps and block keys. Before this PR the only cost was a stale `throughput`/block range; now the *addresses* come from whichever bootstrap answered first, so a peer that has just rotated relays and re-announced (the case `9f5550a6` added the fingerprint for) can still be seeded from a cached record naming the relay it left. `forget` then drops it and the burst fails. Preferring the fresher record — or at least one with non-empty `dial_addrs` — would be more robust than first-seen.

dazwin:
Rebased onto main (two conflicts in the deleted p2pd `unannounce`, resolved to main's deletion) and pushed `dceb9532`. The 11 original commits are preserved; four review commits on top.

1. **Saved circuits reload with no addresses — fixed.** `SerializableEntry` gains `#[serde(default)] dial_addrs`; when a loaded circuit has any empty entry, one `discover_chain` fills those peers before the pre-connect, at all three load sites (`shard run`, `run_streaming_inner`, `shard api`). Tests: JSON round trip, legacy file without the field, fill touches only empty entries.
2. **QUIC on a non-QUIC swarm — fixed, default untouched.** `uses_dialable_transport(addr, quic)` rejects `/quic`/`/quic-v1` (direct or as a relay hop) when the local swarm did not enable QUIC; the daemon applies it at `ConnectPeerWithAddrs` ingestion from `config.enable_quic`, and the publisher takes the same flag. `verified_addrs` keeps passing QUIC and documents that the daemon gates it. #189 still owns the default.
3. **String compare — fixed, premise partly wrong.** The guard now compares parsed `PeerId`s with the `is_empty()` escape kept. No CIDv1 test, because the pinned libp2p-identity 0.2.14 `FromStr` is bs58 + multihash only: a `k51q…`/`bafz…` form does not parse at all, so no two spellings of one id parse on this build. The id compare is in anyway for when that changes.
4. Fixed: `select_dial_addrs` holds one slot for the declared address when it survived the filter, then circuits, then remaining direct. Test with MAX+1 circuits plus declared.
5. Fixed: `discover_chain` keeps `(expiry, entry)` per peer and merges by (has `dial_addrs`, later expiry) instead of first-seen. Test in both read orders, and a fresher empty copy does not displace.

`p2p://auto` (`discover_inference_peer`) still picks by throughput, not freshness — unchanged, noted.

Checks: fmt clean, clippy clean, `cargo test -p kwaai-p2p` 129 green, `cargo test -p kwaainet --bin kwaainet` 270 passed. #203 will be rebased onto this head.

PR #186, opened 2026-09-03
feat: IPv6 support — `ipv6: auto|true|false`, a real v6 classifier, dual-stack local servers

dazwin:
## Summary

Main already bound `/ip6/::` beside the v4 wildcard, but nothing surrounded that listener: no way to turn v6 off or require it, a stub classifier that treated link-local and ULA addresses as public, v4-only sockets on every local server, and no way to see whether v6 was actually up. This series adds the control, correctness, and observability around it.

- **Config + feature.** `ipv6: auto | true | false` (default `auto`) in `NetworkConfig` and `KwaaiNetConfig`, settable with `config set`. Hand-written serde keeps `true`/`false` as YAML booleans and `auto` as a string. Cargo feature `ipv6`, on by default in `kwaai-p2p` and `kwaainet`; compiled out, the key still parses and reads as `false`.
- **Semantics.** `auto` probes `[::1]` and opens v6 listeners if the host has a v6 loopback, otherwise warns once and runs v4-only. `true` makes that a startup error. `false` opens no v6 listener and drops v6 addresses from dial candidates and announce/identify intake. (The probe is a concrete address, not `[::]`, because Linux accepts a wildcard v6 bind even with IPv6 disabled.)
- **Classifier.** `is_routable_v6` / `is_globally_routable_v6` mirror the v4 pair: reject unspecified, loopback, multicast, link-local, site-local, discard-only and v4-compatible; classify v4-mapped addresses as the v4 they carry; reject unique-local `fc00::/7` outright, as RFC1918 is on the v4 side; accept the documentation/benchmarking prefixes non-strict and reject them strict, the same carve-out `is_routable_v4` makes for RFC2544. `AddrPolicy` carries the strict and ipv6 decisions together through identify intake, dial-candidate selection and the reachability gates.
- **Local servers.** gRPC control, the OpenAI-compatible API, shard API, RAG API and the storage health endpoint bind v4 and v6 on one port via `kwaai-cli::net::bind_dual_stack`, v4 first and `IPV6_V6ONLY` set explicitly so Linux, macOS and Windows behave identically. `127.0.0.1` stays the reported gRPC port. `port_in_use` checks both families. A v6 `public_ip` is announced as `/ip6/…`.
- **Observability.** `SelfStatus.ipv6` (`"off" | "active" | "unavailable"`, proto field 10, no new frame tags); `p2p info` prints the configured mode. Docs in `docs/IP_DISCOVERY.md` and `docs/BOOTSTRAP.md`; CHANGELOG entry.
- **CI.** The feature-combo job gains an ipv6-off test and build so the gate can never silently rot.

Out of scope, deliberately: the p2pd path (`node.rs`, `kwaai-p2p-daemon`) is untouched; per-family reachability (a node can be NATed on v4 and end-to-end on v6) is a follow-up.

## Test plan

- [x] `cargo fmt --all --check`, `cargo clippy --all-targets -- -D warnings`, `cargo test -p kwaai-p2p`, `-p kwaai-storage`, `-p kwaainet`, `cargo test -p kwaai-p2p --no-default-features --features tcp`, `cargo build -p kwaainet --no-default-features --features storage,rag` (macOS).
- [x] Windows 11: 543 tests pass across the three crates with zero v6 skips, including a real two-node connection over `::1`; `netstat` shows `0.0.0.0`/`[::]` on the p2p port and `127.0.0.1`/`[::1]` on 8093 with `auto`, v4 only after `config set ipv6 false`.
- [x] Linux (Docker): sealed nat-test topology with a dual-stack ULA network — v6 listen, v6 dial to a Rust bootstrap, VPK storage matrix 4/4, v1/v2 map agreement. In a container with `net.ipv6.conf.all.disable_ipv6=1`: `true` exits non-zero with `ipv6 is required but this host has no IPv6 loopback`, `auto` runs v4-only and reports `unavailable`.

Companion GUI PR: Kwaai-AI-Lab/KwaaiNetGUI `feat/ipv6`.


RezaRassool (commented):
## Review — eight findings, one that makes the new CI job inert

I read every hunk, opened surrounding context, and verified the mechanical claims by building (`cargo check -p kwaainet --all-targets` with `P2PD_PATH` stubbed — clean), running `cargo test -p kwaai-p2p` (all pass on a dual-stack Mac), and resolving the feature graph with `cargo tree -e features`.

Nothing wrong found in the v6 classifier's prefix arithmetic — I checked `3fff::/20`, `2001:2::/48`, `2001::/32`, `2002::/16`, the ORCHID `/28`s, `100::/64`, `::/96` and the v4-mapped delegation — nor in the `oneshot`→`broadcast` shutdown conversion, `serve_all`/`run_storage_api_on` error propagation, or `bind_v6`'s `IPV6_V6ONLY`/`SO_REUSEADDR` handling.

### Findings

1. **`core/crates/kwaai-cli/Cargo.toml:136` — the `ipv6` cargo feature can only ever turn IPv6 *on*.** `kwaai-p2p` is pulled with default features by both `kwaai-cli` and `kwaai-p2p-daemon`, and `ipv6` is one of its defaults, so `--no-default-features` still leaves `IPV6_BUILD == true`. `cargo tree` confirms. The new "Build kwaainet without ipv6" CI job therefore guards nothing — it passes whatever happens to the feature.

2. **`core/crates/kwaai-p2p/src/service.rs:1205`** — a *live connection's* address is now filtered through `policy.dialable`. With `ipv6: false`, an explicit `p2p connect /ip6/…` still succeeds (`Command::ConnectPeer` → `dial()` is unfiltered) but the peer then has zero addresses and `KnownPeers` skips it (`if addrs.is_empty() { continue }`). Also contradicts the function's own doc comment.

3. **`core/crates/kwaai-cli/src/grpc_server.rs:2035`** — under `ipv6: true`, a v6-only clash on the gRPC port makes `bind_dual_stack` fail for the pair, and on the default (non-explicit) port that error is swallowed to `bound = None`. So the daemon runs with **no** gRPC TCP listener and no port file even though v4 was free.

4. **`service.rs:456`** — `ipv6: true` doesn't error when explicit `listen_addrs` contain no v6 address: the node starts v4-only and merely reports `unavailable`, which is the silent half-network failure that `true` (rather than `auto`) exists to prevent.

5. **`service.rs:500`** — `AddrPolicy.ipv6` comes from `!ipv6_mode.is_off()` rather than `open_v6`, so a node that fell back to v4-only under the default `auto` still keeps v6 addresses in its dial and announce sets. (#190 inherits this; its own `AddrPolicy::relay_candidate` is the fix.)

6. **`core/crates/kwaai-p2p/tests/ipv6.rs:99`** — `off_filters_v6_dial_candidates` spawns `Ipv6Mode::On` guarded only by `IPV6_BUILD`, so it panics on a v4-only host — including the `disable_ipv6=1` container the PR says it tested in. Every other v6 test in the file guards with `v6_loopback_works()`.

7. **`core/crates/kwaai-cli/src/daemon.rs:672`** — `port_in_use` hardcodes `Ipv6Mode::Auto`, so with `ipv6: false` a v6-only squatter makes `storage serve` refuse to start with a false "already running".

8. **`docs/IP_DISCOVERY.md:203`** — the heading "why it **accepts** ULAs" contradicts its own body and the code (ULAs are rejected). The PR description carries the same error and is worth correcting too, since that's the version people will read first.

dazwin:
Rebased onto main (#191/#192; two conflicts on the `native_p2p` accessor and the config-key doc list that #191 deleted, resolved to main's deletions) and pushed `f6fe00b3`. All eight addressed.

1. **Feature could only turn v6 on — fixed.** The cause was the workspace: members' `default-features = false` is ignored unless `[workspace.dependencies]` says so. Moved there for `kwaai-p2p` and `kwaai-p2p-daemon`; the daemon crate gets its own default-on `ipv6` feature that `kwaainet` forwards. `cargo tree -e features --no-default-features -i kwaai-p2p` now shows only `tcp`. New test asserts `IPV6_BUILD == cfg!(feature = "ipv6")`, and the CI step runs that test under `--no-default-features`, since a build alone cannot fail on this.
2. Fixed: a live connection's address is listed unfiltered; doc corrected. Test guarded on v6 loopback.
3. Fixed: `bind_dual_stack` returns `BindError::{V4, V6Required}`; a v6 failure never drops the v4 listener, and under `true` it is loud even on the default port. Test squats on `[::]` rather than `[::1]` because BSD lets wildcard and specific binds share a port under `SO_REUSEADDR`, so `port_is_free` cannot see a loopback-only squatter on macOS. That blind spot is pre-existing and left alone.
4. Fixed: `ipv6: true` with an explicit listen set lacking `/ip6/` is a config error before the host probe. Test.
5. Fixed: `AddrPolicy` is built once from `open_v6` and shared with the reachability state. Test.
6. Fixed: the `On` spawn is guarded by `v6_loopback_works()` and now lists a v6 listen address, which 4 requires.
7. Fixed: `port_in_use(port, mode)`; callers pass the configured mode.
8. Fixed: heading reads "why it rejects ULAs". PR description corrected to match.

Checks: fmt clean, clippy clean on kwaai-p2p, kwaai-p2p-daemon and kwaainet, `cargo test -p kwaai-p2p` 121 unit + all suites green (also under `--no-default-features --features tcp`), `cargo test -p kwaainet --bin kwaainet` 258 passed. #190 will be rebased onto this head.

RezaRassool:
## Rebase needed — four files

`main` moved and this now conflicts in:

- `.github/workflows/ci.yml`
- `core/crates/kwaai-p2p/src/addresses.rs`
- `core/crates/kwaai-p2p/src/service.rs`
- `core/crates/kwaai-p2p/src/lib.rs`

The largest of the five outstanding rebases, and worth doing first because **#190 is blocked behind it** — #190 itself is clean, it just cannot merge until this does.

**`ci.yml`** is the easy one: #214 added an *Orphaned doc comments* job in the same region where this PR adds its two `without ipv6` steps. Both are wanted; it is a textual collision, not a disagreement.

**`addresses.rs` and `service.rs`** are the real work. #185 landed `LearnedAddrs` and reworked `candidate_addresses`/`add_routing_address`, which is the same surface this PR's `AddrPolicy` filters. Worth checking on the way through that `policy.dialable` is still applied at the right points once `learned_addrs` is in the picture — the finding about a live connection's address being filtered out is easier to reason about against the new code than the old.

---

dazwin:
Rebased onto main `8c5660b7` and pushed `691465e7`. Ten original commits kept in order, plus one test commit.

Conflicts (read at the rebased head, not inferred from the hunks): `ci.yml` keeps the #214 job, both `without ipv6` steps and #204's cudarc step, and parses as valid YAML; `addresses.rs`/`lib.rs` are the union of `uses_dialable_transport` and the v6 classifier with a single export list; `service.rs` keeps `policy: AddrPolicy` + `ipv6_status` beside #185's `dials_quic`.

Your question about `learned_addrs`: `candidate_addresses` runs learned entries through the same filter as kad entries, and `dial_candidates` passes `policy.dialable` (service.rs:1235–1249). Nothing covered that path, so `691465e7` adds `a_learned_v6_address_is_not_dialed_when_ipv6_is_off`: it fails with `dialable` stubbed to true, passes at HEAD. Finding 2 (live connection listed unfiltered) survives at service.rs:1268–1274 with its existing test.

Checks: orphaned-docs gate clean, fmt, clippy on kwaai-p2p, kwaai-p2p-daemon and kwaainet, `kwaai-p2p` 267 passed (264 under `--no-default-features --features tcp`), `kwaainet --bin` 295, daemon 31.

Issue #174, opened 2026-09-02
Connection manager: watermark trimming, and protocol-aware eviction for the kad migration window


dazwin:
## Summary

rust-libp2p has no connection manager, so a kwaainet node's connection pool is first-come-first-served up to `max_connections`, with only the idle timeout (10 min) shedding load. A live experiment against the public Amino (IPFS) DHT shows both what that costs and what it doesn't, and this issue captures the measurements and a two-part design before they evaporate: a generic watermark ConnManager, plus a protocol-aware classifier that biases the pool toward kwaainet peers during the kad migration window.

## Measurements (bootstrap3 canary, 2026-09-01/02)

A t3.small (2 GiB) running the dual-protocol build (`feat/kad-protocol-alias`, `--features kad-multi-protocol`) was pointed at the Amino bootstrappers as its only `initial_peers`, contained from the KwaaiNet fleet, and left to take the flood. `max_connections: 8192`, `MemoryMax=1400M`, `LimitNOFILE=65536`.

| observation | value |
|---|---|
| memory per connection | **~100 KB** (76 MB @512 → 686 MB @6.5k), linear |
| load average at 6.5k connections | **0.1–0.4** |
| service restarts / OOM | **zero** |
| discovery of a fresh EIP serving `/ipfs/kad` | **minutes** |
| inbound demand vs cap | **pins any realistic cap** — ratcheted 1k → 4k → 5.7k → 6,474 of 6,553 inbound slots in ~10 h, still growing with routing-table embedding |
| churn while pinned | **hundreds of connections per 30 s window** (kubo's own ConnManager trims aggressively on the far side) |

Two conclusions separate cleanly:

1. **Resource exhaustion — the p2pd disease — is cured.** The Go daemon OOM-killed this instance class every 30–90 min with 300–600 foreign peers tabled; the Rust node idles at 6,500 live connections on a third of its memory budget.
2. **Admission is protocol-blind.** While a bootstrap serves `/ipfs/kad/1.0.0` (the migration window), Amino keeps the inbound pool saturated. A kwaainet peer competes for churn slots against crawlers that redial constantly.

## Why this does not block the cutover

[...]

PR #189, opened 2026-09-03
fix(p2p): one QUIC default for both crates, written only when set explicitly

dazwin:
Started as "enable QUIC by default again" (#162 made it opt-in). The flip is now **held**, and this PR keeps only the mechanism that a future flip, or any other default change, needs. The default stays `false`.

## What changes

- **One default, read by both crates.** `kwaai_p2p::config::DEFAULT_ENABLE_QUIC` (currently `false`) is the value behind `NetworkConfig::default()`, the kwaai-p2p serde default and the CLI fallback, with a test pinning each side so they cannot drift. Flipping later is a one-constant change.
- **The default listens on what it dials.** `NetworkConfig::default()` no longer pins `listen_addrs` to TCP; `swarm_listen_addrs()` derives TCP plus, when `enable_quic` is set, `/udp/<port>/quic-v1`. `for_tests()` still pins loopback TCP, so no test binds UDP.
- **`enable_quic` is `Option<bool>` in `KwaaiNetConfig`**, written only when set explicitly, read through `quic()`. `config set enable_quic` stores an explicit value. This is the first instance of #212's direction, stop writing defaults into `config.yaml`, and the shape the other keys can follow.
- The default-config test in `02_unit_p2p.rs` asserts the QUIC listener follows the flag rather than a tautology.

Nothing in this PR changes behaviour for any node: a file with `enable_quic: false`, a file without the key, and a fresh default all run TCP-only as before. A file with `enable_quic: true` keeps QUIC.

## Why the flip is held, and what was learned

- **#212.** Every release up to v0.7.0 wrote `enable_quic: false` into every file, so a flipped constant reaches no existing install. A load-time migration was built and then dropped: it needs a `config_version` stamp to tell an old release's `false` from an operator's, and the config is moving away from auto-saved defaults, so that field would be complexity in the wrong direction. Whatever explicit clean-up the no-auto-save work provides is the right vehicle.
- **#179 ordering.** Item 1, bootstraps in the preferred and never-evicted relay tier, is done on main via #200's `RelayManager::with_initial_peers`; `trusted_relays` stays an operator override, empty by default. See the comment below.
- **The flip is safe on our own topology.** Main versus main plus the flip on the sealed eight-node bed, captured at the same age: identical hole punches, TCP circuits on both relays, declared TCP external addresses, VPK matrix 4/4 cold and warm including NAT→NAT, map crawl 8 peers, no `Failed to negotiate transport`, no ERROR lines; QUIC carried 7–45 % of node-to-node connections including QUIC over IPv6, and the bed's IPv6 suite passed apart from a map row missing on main too. One new WARN at startup from quinn when a v4-only NATed node dials a peer's v6 QUIC address. Production bootstraps allow TCP 8000 only and announce only their declared TCP address, so they would stay TCP relays. Details in the comments.

Checked: `cargo fmt --all --check`, `cargo clippy --all-targets -- -D warnings`, `cargo test -p kwaai-p2p` (292), `cargo test -p kwaainet --bin kwaainet` (310), `02_unit_p2p` (7), `scripts/check_orphaned_docs.py --base origin/main`.


RezaRassool (commented):
## Review — the flip is a no-op for existing installs

Eight lines and fully green, so this looks like the cheapest merge in the queue. It isn't quite: as written it changes nothing for anyone who already has KwaaiNet installed. I read both config files, `service.rs`'s transport build, `node_native.rs`'s config assembly, `swarm_listen_addrs()`, `tests/quic.rs`, the #155/#162 history and #179's full thread.

### Findings

1. **`core/crates/kwaai-cli/src/config.rs:234` — the default flip never reaches an existing install.** `save()` serializes the whole struct and `enable_quic` has no `skip_serializing_if`, so every `config.yaml` written since #155 pins the key — `false` for anything installed on v0.6.4–v0.6.8. `load_or_create()` never rewrites an existing file on upgrade.

   Verified empirically: `~/.kwaainet/config.yaml` on the maintainer's machine, line 20, reads `enable_quic: false`.

   The same file documents this trap twice and solves it twice: `native_p2p` is `Option<bool>` + `skip_serializing_if` for exactly this reason, and `migrate_default_peers` exists because "`initial_peers` is written to every config.yaml, so a new `default_peers` alone would never reach an existing install." Suggest mirroring `native_p2p` — `Option<bool>` with a `quic()` accessor — or adding a migration alongside `migrate_default_peers`.

2. **`config.rs:915` — QUIC-on was measured not to fix the circuit dial.** #179 records that with `enable_quic: true` the dial gets further but still fails with `Failed to negotiate transport protocol(s) … oneshot canceled` inside the relay behaviour, and the issue's re-ordered plan demotes this flip to item 3 behind populating `trusted_relays`: "the QUIC mismatch is therefore a symptom, not a root cause… I no longer think it should be fixed first." Net effect of merging as-is is UDP:8080 exposed by default on new installs with no gained dialability.

   Flagging rather than objecting — whether the flip should land ahead of the `trusted_relays` work is a call for @RezaRassool, not something to settle in review. Worth noting it also has downstream reach: #185's finding on `uses_dialable_transport` (which accepts QUIC addresses on a node that never listens on QUIC) resolves differently depending on which way this goes.

3. **`core/crates/kwaai-p2p/src/config.rs:79` — `NetworkConfig::default()` is now internally inconsistent.** `enable_quic: true` alongside `listen_addrs: ["/ip4/0.0.0.0/tcp/0"]`, so `swarm_listen_addrs()` never reaches the QUIC branch: the default dials QUIC but never listens on it, contradicting the field's own doc. Only `node_native.rs` clears `listen_addrs` to get the listener.

4. **`core/crates/kwaai-p2p/src/config.rs:302`** — no test pins either default, and the serde default can drift from the `Default` impl. Four hand-edited sites across three PRs now; `enable_quic_decides_whether_quic_is_listened_on` sets the flag explicitly in both arms, so a one-sided future edit (in-process default ≠ on-disk default) passes CI green.

### Checked, not problems

`default_true()` exists in both crates; the multi-connection-per-peer path is keyed by `ConnectionId`, so a second QUIC connection to the same peer is handled correctly; UPnP toggles on listen addrs rather than transport; `for_tests()` pins explicit `listen_addrs`, so no test starts binding UDP; the `find_free_port`/`port_in_use` TCP-only helpers are dead code, so no port-picker mismatch; stock builds no longer serve `/ipfs/kad/1.0.0`, so the added QUIC listener doesn't widen the IPFS-absorption surface of #174; and no `.md` doc asserts the old default.

dazwin:
Rebased onto main (clean) and pushed `10eee8da`.

1. **Default never reaches an existing install — fixed the `native_p2p` way.** `enable_quic` is `Option<bool>` with `skip_serializing_if`, a `quic()` accessor returns the explicit value or the shared default, `config set` stores `Some`, and `save()` no longer pins the key. Tests: a 0.6.8-style file with `enable_quic: false` stays false; a file without the key takes the default; serialising the default writes no key (fails without the fix). One deliberate choice: an explicit `false` written by the v0.6.4–v0.6.8 default-writing `save()` stays explicit. The file carries nothing to tell "operator chose TCP-only" from "the old save wrote the default", and turning UDP on for an operator who chose otherwise is the worse error, especially while #179 records the flip gaining no dialability yet. If you want that cohort migrated, it is a `migrate_default_peers`-style pass that clears an explicit `false`; a policy call rather than something to guess here. Two docs did assert the old default (`docs/network-and-intent-routing.md`, `docs/ARCHITECTURE.md`); both updated.
2. **Flagged, no code change.** Whether the flip lands ahead of `trusted_relays` is yours to call.
3. **Inconsistent default — fixed.** `NetworkConfig::default()` has empty `listen_addrs`, so `swarm_listen_addrs()` derives TCP plus QUIC from `port`/`enable_quic`, which is what `node_native.rs` did by hand. `for_tests()` still pins loopback TCP, so no test binds UDP. Test fails without the change.
4. **Drift — fixed.** `pub const DEFAULT_ENABLE_QUIC` in kwaai-p2p is the single value behind the `Default` impl, the kwaai-p2p serde default and the CLI fallback, with a test pinning each side and one asserting the CLI equals `NetworkConfig::default()`.

Checks: fmt clean, clippy clean on kwaai-p2p, kwaainet and kwaai-network-tests, `cargo test -p kwaai-p2p` 107 lib + all suites green, `cargo test -p kwaainet --bin kwaainet` 251 passed.

dazwin:
Rebased onto main `8c5660b7` (clean) and pushed `ed24dab4`. Two commits added on top of the three:

- `b3ccd4e7` replaces the `!swarm_listen_addrs().is_empty()` assertion, which could not fail, with checks that the default listens on `/tcp/<port>` and `/udp/<port>/quic-v1` and that `enable_quic: false` drops the QUIC entry. Flipping `DEFAULT_ENABLE_QUIC` fails it.
- `ed24dab4`: the review-fix commit had inserted `quic()` between `reloaded()`'s doc block and `reloaded()`, which the new #214 gate rejects. Moved below `reloaded()`.

#185's `uses_dialable_transport` gate and this branch's default read one value: `KwaaiNetConfig::quic()` → `net_config.enable_quic` feeds both the publisher (`dials_quic` → `signed_dial_addrs`) and the daemon's ingestion gate (service.rs:449/597). `peer_record.rs:109` passes a literal `true` by design; unchanged.

Red/green on both halves: `quic()` reverted to `unwrap_or(false)` fails `a_config_without_the_key_takes_the_swarm_default`; `#[serde(default)]` restored fails `a_config_without_the_key_takes_the_same_default`. Checks: gate clean, fmt, clippy, `kwaai-p2p` 246, `kwaainet --bin` 288, `02_unit_p2p` 7.

RezaRassool:
## Holding this until `trusted_relays` lands — #179 item 1

Not a rejection of the code. The `Option<bool>` work is right and I have verified it at `10eee8da`: `skip_serializing_if` plus the `quic()` accessor, `DEFAULT_ENABLE_QUIC` as a single const both crates read, `listen_addrs` defaulting empty so the QUIC listener follows the flag, and `node_native` reading `config.quic()`. That is the shape I asked for and it is done.

The reason to wait is ordering, and it is your own analysis in #179 that argues it. That comment places this work at **item 3** and says plainly: *"The QUIC mismatch is therefore a symptom, not a root cause… nodes reserving on kwaai bootstraps would get TCP circuits on hosts we control and the mismatch would largely disappear. That is why I no longer think it should be fixed first."*

## Where that plan now stands

Three of its five items landed in the last two days, which changes the picture since it was written:

- **Item 2 — make the configured-relay preference robust rather than incidental.** Done, #200. `pick_candidate` now scopes the rotation cursor to discovered candidates only, and `holds` became a real `Hold` struct, so a configured relay keeps its slot through a backoff instead of losing it by iteration order.
- **Item 4 — carry dialable addresses in the announce record.** Done, #185. `dial_addrs` is in the record with `signed_dial_addrs` binding them to the peer. This is the item your comment calls *"the durable fix, and still the one that actually closes this issue."*
- **Item 5 — the two guards.** Largely covered by #201 (a whole-model node announces no block range) and #203 (skip unreachable peers before pinning rather than after a 60 s timeout).

- **Item 1 — set `trusted_relays` to the kwaai bootstraps.** **Still outstanding.** `default_trusted_relays()` at `config.rs:811` still returns `Vec::new()`. Your comment calls this *"the cheapest possible first move"* and notes no code change is needed — `RelayManager::new` already populates `configured` from it and already prefers configured over discovered, and after #200 that preference is now explicit rather than incidental. It is a non-empty default.

So the item this PR was deprioritised behind is the one item nobody has done, and it got cheaper: #200 built the machinery that makes it stick.

## What that means here

Once item 1 lands and the fleet is reserving on our own bootstraps, the circuits become TCP on hosts we control — and the question this PR answers ("should QUIC be on by default?") becomes a genuine transport decision rather than a workaround for reserving on QUIC-first third-party IPFS relays. It may still be the right flip. It will be a better-evidenced one, and testable: item 1 is verifiable on a single node.

There is also a second reason not to rush it, which is unresolved regardless of ordering. As it stands the flip **reaches no node currently on the network** — every v0.6.4–v0.6.8 install has `enable_quic: false` written to disk and this PR's own test deliberately keeps it. That is exactly what you raised in #212.

Worth recording there: under the *old* default of `false`, a `false` on disk carries almost no intent. A user who wanted QUIC off had to do nothing; the only value worth setting deliberately was `true`. So clearing `Some(false)` is far safer than #212's framing suggests, and a migration that clears only `false` preserves a deliberate `true` automatically. The real obstacle is that there is no `config_version` field, so such a migration cannot tell "false written by 0.6.8" from "false set deliberately by a 0.6.9 user" — which is an argument for adding one, not for skipping the migration.

**Rebase when it suits you** — this now conflicts with `main` after #202 landed (both touch `config.rs` and `main.rs`). No rush given the hold; I did not want to rebase it under you while it is parked.

Tagging @dazwin since item 1 is the unblocking move and your map-side data is what established it.

dazwin:
Rebased onto main `e66dd622` and pushed `9c538a6b`. Three conflicts, all against #186/#190's new `ipv6` and `only_global_ips` fields sitting beside `enable_quic`; kept both sides. One trap on the way: rerere replayed a stored resolution of the second commit that had dropped main's `ipv6: Ipv6Mode::Auto` from the p2p `Default` initialiser, so the branch did not compile until I fixed that inside the commit. Every hunk was then checked against the commit's own diff rather than trusted.

## The migration

Mirrors `migrate_default_peers`: on load, a value equal to what the old release wrote is treated as unset and the file re-saved.

- `config.yaml` now carries `config_version` (`0` when absent, `1` from this release; the constant documents what each number means).
- `migrate_enable_quic` clears `enable_quic: false` only in a version-0 file, then the file is stamped. `true` was the only value worth setting under the old default, so it is deliberate and survives.
- A `false` set after the flip, by `config set enable_quic false` or by hand in a stamped file, is never touched. That is the case you raised in #212 and the reason a version marker is needed rather than exact-match alone: unlike a legacy peer list, `false` can recur.
- Changelog entry tells an operator on a UDP-blocked network who never set the key to set it now.

Tests, red/green on both halves: with the migration made a no-op, `a_pre_version_false_is_cleared` fails; with the version guard removed, `a_false_written_after_the_flip_survives` and `set_key_then_reload_keeps_quic_off` fail; restored, `kwaainet --bin` 316 passed, `kwaai-p2p` 292, `02_unit_p2p` 7, clippy and the orphaned-docs gate clean.

## Side effects when it is on: sealed bed, main vs main + this branch

Same eight-node topology (public, port-forwarded, cone NAT, symmetric NAT), same binary source apart from this branch, captured at the same age after `make restart`. No node in the bed sets `enable_quic`, so the whole topology flipped.

| | main (QUIC off) | this branch (QUIC on) |
|---|---|---|
| Map crawl | 8 peers, 2/2 bootstraps | 8 peers, 2/2 bootstraps |
| VPK matrix, cold + warm, incl. NAT→NAT | 4/4 | 4/4 |
| DCUtR hole punches (node-d / f / g) | 2 / 2 / 2 | 2 / 2 / 2 |
| Relay reservations on both relays (f, g, h) | TCP circuits | TCP circuits, unchanged |
| Confirmed external addresses | declared TCP | declared TCP, unchanged |
| QUIC share of node-to-node connections | 0 | 7–45 % per node |
| `Failed to negotiate transport` | 0 | 0 |
| ERROR lines | 0 | 0 |
| New WARN class | none | 4 lines fleet-wide at startup: quinn `sendmsg … NetworkUnreachable` when a v4-only NATed node dials a public peer's `/ip6/…/quic-v1` |

The WARN is the one real difference. The NAT routers drop v6 on the LAN leg, so those nodes have no v6 route; the TCP dial to the same address fails too, but at DEBUG. quinn reports the socket error at WARN and once per target. Cosmetic on this evidence; it is an interaction with `ipv6: auto` keeping v6 candidates on a host whose only v6 is loopback, not something this branch introduces beyond making the dial visible.

## What the bed cannot show, reasoned from the code instead

- **Production bootstraps.** Their security group allows TCP 8000 only. With QUIC on they will listen on UDP 8000 behind that rule. They declare `public_ip`, and `configured_announce_addr` builds a TCP address only, and the reachability machine confirms exactly one address, so no unreachable QUIC address is announced, and their private-range QUIC listen addresses are filtered by `only_global_ips` before a relay candidate is chosen. So a NATed node still reserves a TCP circuit on them. Adding UDP 8000 to the group is the change that would let peers use QUIC to the bootstraps; it is not required for this to be safe.
- **Third-party QUIC relays**, the case #179 measured as "gets further, still fails to negotiate". There is no IPFS relay in the sealed bed, so that is not reproduced here either way. It is also the case `trusted_relays` removes, which is why holding this behind #179 item 1 remains reasonable; this comment is about the migration and about showing the flip is not harmful on the topology we can test.

dazwin:
On the hold: I think #179 item 1 is already done, and the check that marks it outstanding is looking at the wrong symbol.

Item 1's intent was to put the two bootstraps in the relay tier that is preferred over discovered relays and never evicted. Your 12 Sep status marks it outstanding because `default_trusted_relays()` still returns an empty list. It does, and that is by design: the field's own doc calls `trusted_relays` an operator override, empty by default, for pinning specific relays in a test topology or a deployment that must prefer one.

The mechanism landed in #200 instead, on main at `b4025406`:

- `relay_manager.rs:181–194`, `RelayManager::with_initial_peers`: the initial peers are added to `configured`, with the comment "The initial peers are configured relays too. A bootstrap is dialed first, is always on, and runs a relay with limits sized for inference, so it belongs in the tier that is never evicted from the candidate list."
- `service.rs:650`: `RelayManager::new(…).with_initial_peers(&config.effective_initial_peers())` at startup.
- Test `initial_peers_are_configured_candidates` (`relay_manager.rs:813`): deduplicated, trusted entries first, an entry without a peer id dropped.

It reaches production with no config change: `default_peers()` carries both bootstrap peer ids on `/dnsaddr/bootstrap.kwaai.ai`, and `migrate_default_peers` rewrites older installs to that list, so a default install feeds both bootstraps into the configured tier. The gap is an operator-written `initial_peers` list without `/p2p/` ids, which logs a warning and yields no configured relay. On the bed, node-f, node-g and node-h each reserved on bootstrap1 in both the main and the main+#189 runs, which is the preference working.

So the fleet is, or will be as nodes upgrade to v0.7.0, reserving on our own bootstraps over TCP, which is the world you wanted before deciding this flip. The bed results above are measured in that world. If you agree item 1 is discharged by #200, the remaining question is the transport decision itself, and I would rather have that decided on this evidence than leave the migration unmerged. If you still see something item 1 needs beyond `with_initial_peers`, say what and I will do it.

dazwin:
Change of direction, pushed as `8b5168a8` and reflected in the title and description.

**The default flip is held, and the `config_version` migration is gone.** We are moving `config.yaml` away from auto-saved defaults (#212), and a version stamp that exists only so a load-time rewrite can second-guess a written `false` is complexity pointing the other way. With the default unchanged there is nothing to migrate: `false` on disk equals the default whichever way it is read.

**What the PR keeps** is the mechanism you already verified at `10eee8da`, with the constant set to `false`: `DEFAULT_ENABLE_QUIC` as the single source behind both crates, the listen set derived from the flag, and `enable_quic` as `Option<bool>` written only when set explicitly, which is the first key to follow #212's direction. Behaviour is unchanged for every node. Flipping later is a one-constant change, and the bed comparison above is the evidence for that day.

Checks at the new head: fmt, clippy, `kwaai-p2p` 292, `kwaainet --bin` 310, `02_unit_p2p` 7, orphaned-docs gate clean.

On the #179 item 1 point in my previous comment, that stands independently of this: it is about whether the hold's precondition is already met on main, and worth settling regardless of when the flip happens.

PR #190, opened 2026-09-04
feat(p2p,cli): `only_global_ips` config key, strict by default, for both address families

dazwin:
Stacked on #186 (base is `feat/ipv6`); the diff here is the two commits on top of it.

## Why

KwaaiNet carried an allowance for IANA-reserved address space in code: `is_routable_v4` admitted RFC2544 and RFC5737, the internal `require_global_ips` switch defaulted off and was hard-coded off in the CLI, and every comment explaining why named one particular docker topology. A deployment detail had become the crate's default, and a node on the internet could announce a benchmarking-range address with no way to refuse it. #186 removed the same carve-out on the IPv6 side outright; this PR puts both families under one operator decision.

## What

- **`only_global_ips`** is a config.yaml key, default `true`, named for the libp2p AutoNAT knob it drives. It reaches the daemon's `AddrPolicy`, the relay manager's candidate filter and `autonat::Config::only_global_ips`, so the swarm's one address-class check and the classifier cannot disagree.
- **On (default):** RFC2544 and RFC5737 on v4, unique-local and the documentation and benchmarking prefixes on v6 are rejected like private space. **Off:** the lenient tier admits them. The transition ranges (Teredo, 6to4, ORCHID) are rejected in both tiers. This restores a lenient v6 tier so v4 and v6 follow the same rule.
- `is_announceable` now means the default policy; `is_announceable_with` takes the configured one.
- Nothing in the crate names any deployment. `NetworkConfig::for_tests` keeps the lenient setting because its loopback fixtures are tabled through AutoNAT's dial-back, which the strict setting refuses to make to 127.0.0.1.
- Docs (`BOOTSTRAP.md`, `IP_DISCOVERY.md`, `NATIVE_P2P_SECURITY.md`) and the changelog updated, including a **Changed** entry: a deployment on reserved space must now set `only_global_ips: false`.

`NetworkConfig::require_global_ips` keeps its internal name for now: it is on main and in the published 0.6.8 crate, and #185 reads it by that name. Renaming it is a one-line follow-up once both land.

## Verified

`cargo fmt --check`, `clippy --all-targets -D warnings` on `kwaai-p2p` and `kwaainet`, and the full test suites of both, on this branch and on the integration branch with #185 merged.


RezaRassool (commented):
## Review — seven findings

Reviewed the two commits on top of `feat/ipv6` (#186) in a clean worktree and ran the full `kwaai-p2p` suite against them — 120 unit tests plus all integration targets, all green. Worth flagging that this was a manual run: **this PR has no test CI**, only the release workflow's `plan` job, so `Test`/`Clippy`/`Format`/`Windows check`/`WASM` never ran.

The v6 masks check out (`3fff::/20` as `s[1] & 0xf000 == 0`, `2001:2::/48`, ULA via `is_unique_local`, and the transition ranges staying rejected in both tiers), and `for_tests` keeping the lenient setting is correct — AutoNAT's `only_global_ips` gates the observed address of the peer connection, so a strict test node would refuse loopback dial-backs.

### Findings

1. **`core/crates/kwaai-p2p/src/relay_manager.rs:140` — the relay tier still ignores the v6 half of the policy.** `RelayManager::new` was given a bare `strict: bool` instead of the `AddrPolicy` that `service.rs:498` already holds, so relay-candidate selection applies only `only_global_ips`. `service.rs:2286` hands `note_identify` the raw `info.listen_addrs`, and `relay_manager.rs:280` takes the *first* match and pins it in `entry.1` for as long as the peer keeps listing it.

   Concrete failure: a v4-only host (`ipv6: false`, or `Auto` resolving to unavailable) meets a relay that's publicly reachable only over IPv6. Its v4 address is a LAN address and gets filtered; its `/ip6/2606:…` passes `is_relay_candidate_addr(_, true)`. So the node pins the v6 address, every reservation attempt fails on a host with no v6, and it never tries anything else for that relay — the metro-win "2350 `listen_on` refused" mode. Inherited from #186, but this is the PR threading the policy, and finding 2 is exactly the fix.

2. **`core/crates/kwaai-p2p/src/addresses.rs:253`** — `AddrPolicy::relay_candidate` is added and **never called anywhere in the tree**; `relay_manager.rs` uses the free `is_relay_candidate_addr(a, self.strict)` instead. It's dead `pub` code, so no `dead_code` warning fires, and it leaves two overlapping APIs for the same question — the drift the module docs say the policy type exists to prevent. Either call it from `RelayManager` (fixing finding 1) or drop it.

3. **`CHANGELOG.md:14` and `core/crates/kwaai-cli/src/config.rs:247`** — the key is wired only into the native path (`node_native.rs:176`). A node with `native_p2p: false` still runs `node.rs::is_globally_routable_v4`, which accepts RFC5737 unconditionally and whose `Ip6` branch rejects only unspecified/loopback, so ULAs are announced too. Meanwhile the changelog states flatly that "IANA-reserved address space is rejected by default" and `docs/BOOTSTRAP.md` lists the key with no caveat — where the sibling `ipv6` field's doc (`config.rs:304`) does say "**Native path only** … The p2pd path ignores this". Same class of trap as `force_private` not being settable via `config set`. Moot if #191 merges first, which is where it currently sits in the order.

4. **`core/crates/kwaai-p2p/src/reachability.rs:521`** — the default tier is now nearly untested. `plain()` hard-codes lenient and almost every fixture address is RFC5737 documentation space (`203.0.113.7`, `198.51.100.1`), which the new default rejects. Only `only_global_ips_excludes_the_reserved_ranges_from_consensus` builds a strict state, so a regression in AutoNAT promotion, UPnP or identify consensus **under the shipped default** would pass the suite. Flipping `plain()` to strict with a real public address (keeping one explicit lenient case) would put the coverage where the default now is.

5. **`docs/NATIVE_P2P_MIGRATION.md:636`** — the rename commit changed this line to `only_global_ips` but left "default false" in place, so the doc now asserts the opposite of the shipped default on the very line the PR touched.

6. **`docs/IP_DISCOVERY.md:203`** — the heading "The classifier, and why it accepts ULAs" survives a rewrite whose body now says ULAs are rejected under the default policy and admitted only in the lenient tier.

7. **`core/crates/kwaai-p2p/src/addresses.rs:32` (also `config.rs:200`)** — "It is driven from the same key, so the swarm and the classifier cannot disagree" overstates it: `autonat`'s `is_global_ip` and the crate's strict classifiers are different rulesets. With the key on, AutoNAT accepts Teredo, 6to4, ORCHID, `3fff::/20` and `2001:2::/48` which `is_globally_routable_v6` rejects; and AutoNAT rejects `0.0.0.0/8`, `192.0.0.0/24` and `240.0.0.0/4` which `is_globally_routable_v4` accepts. Concretely, a node whose only address is 6to4 gets an AutoNAT `Public` verdict that `reachability.rs:288` then discards as non-announceable — benign, but it's the disagreement the comment promises cannot happen. Softening to "driven from the same key, so neither can be tightened without the other" would be accurate.

### Note

`is_announceable`'s meaning flips from lenient to strict with no signature change, and `kwaai-p2p` is a published crate re-exporting it at the root. No in-repo caller is affected, but out-of-repo callers change behaviour silently — worth a changelog line under the crate's own heading.

dazwin:
Rebased onto the new `feat/ipv6` head (f6fe00b3, on main with #191/#192; two conflicts in the `config set` key list where #191 removed `native_p2p`, resolved to main's deletion) and pushed `cdb54678`.

1 and 2. **Relay tier ignores the v6 half — fixed.** `RelayManager::new` takes the `AddrPolicy` that `service.rs` builds from what actually opened, and both selection sites call `AddrPolicy::relay_candidate`; the free `is_relay_candidate_addr` is deleted and its metro-win rationale moved onto the method. Test: an `ipv6: false` host offered `[/ip6/2606:…, v4]` dials the v4 address; fails with the old logic swapped back.
3. **Moot after #191.** No `node.rs::is_globally_routable_v4` exists on main; the changelog and BOOTSTRAP.md name no path.
4. Fixed: `plain()` is strict with real public fixtures (`8.8.8.8`, `1.1.1.1`); one explicit `lenient()` case remains for the reserved-range exclusion test.
5. Fixed: "default true", and the "opt-in tightens" line now says default-on rejects and `false` restores the lenient tier.
6. Fixed: heading reads "why it rejects ULAs by default"; the body already said the lenient tier admits them.
7. Fixed: softened to "neither can be tightened without the other" in both places.
Note: changelog line under `[Unreleased] / Changed`, prefixed `kwaai-p2p:`, pointing out-of-repo callers at `is_announceable_with(addr, false)`.

Checks: fmt clean, clippy clean, `cargo test -p kwaai-p2p` 123 unit + all suites green, `cargo test -p kwaainet --bin kwaainet` 258 passed. This PR still has no test CI of its own since it targets a branch; it inherits the run from the rebase push.

RezaRassool:
## Not conflicting — blocked behind #186

No action needed on this PR itself: it is `MERGEABLE/CLEAN` against its base. It simply cannot merge until **#186** lands, since it is stacked on `feat/ipv6`.

#186 now conflicts with `main` in four files and I have asked for a rebase there. Once that merges, this retargets to `main` automatically and should be straight through — its two review findings (threading `AddrPolicy` into `RelayManager`, and `relay_candidate` becoming the live call site rather than dead `pub` code) are already fixed.

One thing to re-check after the retarget rather than now: #191 deleted the p2pd path, so this PR's third finding — that `only_global_ips` is wired only into the native path while `node.rs` still accepts RFC5737 unconditionally — should be moot. Worth confirming the changelog and `docs/BOOTSTRAP.md` wording no longer implies a p2pd caveat that cannot apply any more.

Context: `main` is now `cec33a70`; #214, #185, #193, #204, #205, #200, #203 and #194 landed on 10–11 September. Re-run `core/patches/fetch-patches.sh` after pulling — #204 added a third patched crate and `cargo` will not load the workspace without it.

dazwin:
Rebased onto the new `feat/ipv6` head `691465e7` (`git rebase --onto`, this PR's four commits only) and pushed `9dfa65e4`. Same 14 files as before; base stays `feat/ipv6`.

Conflicts were against #200's relay-manager rework, not #186: `RelayManager::new` keeps #200's `parse_configured`, holds and `pick_candidate` with `AddrPolicy` threaded through, and the `service.rs` call keeps both `policy` and `.with_initial_peers(...)`. A stored rerere resolution had missed #200's new `initial_peers_are_configured_candidates` test call; corrected inside the commit. Both selection sites call `policy.relay_candidate` (relay_manager.rs:304/316); `is_relay_candidate_addr` is deleted, and nothing on main outside relay_manager.rs and its own tests called it.

Red/green: with the old strict-only check swapped into both sites, `a_v4_only_host_does_not_pin_a_relay_at_its_v6_address` fails at relay_manager.rs:1140 (the v6 address is picked); restored, passes. All seven #200 relay tests pass. Finding 3 confirmed moot: neither CHANGELOG.md nor docs/BOOTSTRAP.md carries a p2pd caveat for `only_global_ips`, and this PR's hunks never mention p2pd.

One thing to know about the new gate: `check_orphaned_docs.py --base origin/feat/ipv6` passes, which is what CI runs for this base. Against `origin/main` it reports a false positive at `addresses.rs:183` (`is_globally_routable_v6` has its own doc block; the cumulative two-PR diff aligns a bare `///` line from main's old block with it). It disappears once #186 merges, so nothing to change here.

Checks: fmt, clippy, `kwaai-p2p` 269 passed, `kwaainet --bin` 295.

dazwin:
Follow-up on the red *Orphaned doc comments* check on this PR, run 34656961265.

The job rendered `--base "origin/main"` although this PR's base is `feat/ipv6` (the run's own `pull_requests` entry says so, the timeline has no `base_ref_changed`, and `ci.yml` on the branch uses `${{ github.base_ref }}`). I cannot explain that from here; you may want to look at why the workflow triggered at all under `pull_request: branches: [main]`.

What it flagged is a checker false-positive class rather than an orphaned doc: `is_globally_routable_v6` has its own five-line block at HEAD, but in the cumulative two-PR diff against main one bare `///` line of that block aligns with a bare `///` in the block main had at that spot, so the walk up from the inserted `pub fn` meets an unchanged context line and reports "inherits: ///". I tried reflowing that doc into one paragraph and the same alignment simply moved to `ipv6_loopback_available` next door, so I dropped that commit rather than chase it. The rule that would remove the class is to require the whole preceding block to be unchanged context, not any line of it; a genuinely stolen doc block has no added lines in it at all.

Against the real base, `check_orphaned_docs.py --base origin/feat/ipv6` passes, and so does `feat/ipv6` itself against main, so this goes green once #186 merges and GitHub retargets it.

PR #191, opened 2026-09-04
chore: drop the Go p2pd daemon — every node is native

dazwin:
## Summary

Every node in the fleet runs rust-libp2p in process, so the Go `go-libp2p-daemon` child process, the Go toolchain it dragged into every build, and the release plumbing that shipped `p2pd` beside `kwaainet` are dead weight. This removes the Go half.

**`kwaai-p2p-daemon` stays, name unchanged.** It is the control-socket protocol crate in both directions: `ControlServer` is what the native node serves on `/tmp/kwaai-p2pd.sock`, and `P2PClient` is what `storage serve`, `shard serve`, `rag`, `vpk`, `inference_mux`, the gRPC server and ~18 CLI files use to reach the node. Only `daemon.rs`, the Go/git steps of `build.rs`, and the compiled-in binary path are gone. The crate now builds with protoc alone.

### What changed

- **CLI**: `run_node` always runs the native node (`node.rs` 2657 → 679 lines; the p2pd branch, its watchdog/restart cycle, the loopback RPC listener and `find_p2pd_binary` are deleted). `native_p2p` is a warn-and-ignore tombstone; `config set native_p2p` is rejected. `setup --get-deps` is a hidden no-op. `kill_orphaned_p2pd`, `setup.rs`, and the p2pd arms of the updater are removed. `uninstall` still removes a legacy `p2pd` file next to `kwaainet`.
- **Tests**: the eight `kwaai-network-tests` tiers that spawned the real Go daemon as a reference (04–10, 12) and the harness are deleted; tier 11 keeps its native-to-native case; tier 13 and the crate's `control_server.rs` (pipe mode, unary, handler ownership) already cover the protocol natively. Six `P2PDaemon` examples deleted.
- **Packaging/CI**: cargo-dist `extra-artifacts`, `scripts/build-p2pd.sh`, `nix/p2pd.nix`, every `setup-go` step, the release-archive p2pd injection and installer `_bins` patches, and the Go sections of `setup.sh`/`setup.ps1` are removed. Releases no longer ship `p2pd`.
- **Docs**: README, CONTRIBUTING, CLAUDE.md, nix/README, runbooks, CHANGELOG (Unreleased).

### Not in this PR: the IPC protocol

Dropping the socket protocol is a separate, larger change. Three things block it: `storage serve` / `shard serve` / `block_rpc` register *inbound* libp2p handlers on the node's behalf from another process and `kwaai.proto` has no equivalent; `inference_mux` uses pipe mode (raw byte stream over the socket) with no gRPC analogue; and the gRPC server is itself a socket client. Suggested path: move the handler-registering subcommands in process (`node_native.rs` already does `handle.add_unary_handler` directly), port the read-only consumers to `Session` ops that mostly exist, then retire `p2pd.proto`.

### Verification

- Built and tested with `go` shimmed to exit 127: `cargo build --workspace`, `fmt --check`, `clippy --all-targets -D warnings`, `cargo test -p kwaainet -p kwaai-p2p-daemon`, `KWAAI_INTEGRATION_TESTS=1 cargo test -p kwaai-network-tests` all pass.
- Sealed NAT test bed on this binary: Bob→Eve storage matrix cold+warm 4/4 (inbound handler registration over the socket still works, NAT→NAT included); all eight nodes announce on the map.
- A config carrying `native_p2p: false` starts a native node and logs the warning; `config set native_p2p false` refuses.

### Follow-ups

- KwaaiNetGUI's CI copies `p2pd` out of the KwaaiNet release archive; a matching GUI change (stop bundling it) is prepared and must land before the first release cut from this.
- kwaaiai-env drops its `nat-test-overlay-p2pd` service (committed locally).


RezaRassool (commented):
## Review — in good shape, four findings

The big one, and it holds up. I read every hunk, checked for dangling references to every deleted item (`P2PDaemon`, `DAEMON_BINARY_PATH`, `find_p2pd_binary`, `setup::`, `kill_orphaned_p2pd`, `harness`, `unsigned-varint`, `libc`, the nix `p2pd` attr, `extra-artifacts`, `build-p2pd.sh`) and found none, confirmed `proto/p2pd.proto` is still tracked so the slimmed `build.rs` generates, and ran `cargo check --workspace --all-targets` on the branch in a throwaway worktree — exit 0, no new warnings.

Also confirmed the parts that would be easy to lose: `node_native` does register all four node handlers (hello, ollama-proxy, capacity-lease, shard-proxy) plus the mux server, and honours `public_port`/`announce_addr`, so the deleted p2pd path leaves no functional gap. The deleted tier 12 (pipe mode) is genuinely covered by `kwaai-p2p-daemon/tests/control_server.rs`.

**Worth knowing: this is second in the merge order, right after #192.** Landing it early deletes findings in three other open PRs — #201's most severe one (the p2pd announce path still publishing the bad range), #194's `setup --get-deps` trap, and #190's p2pd-path caveat. It shrinks the queue's total work, not just its length.

### Findings

1. **`core/crates/kwaai-cli/src/daemon.rs:243` — the one I'd fix before merge.** `kill_orphaned_p2pd()` was removed from `stop_process()` and from the `restart` arm (`main.rs:433`). A legacy `p2pd` orphaned by a hard kill still holds the p2p port; `ControlServer::bind` steals the socket file so startup *looks* fine, but the swarm's `listen_on` failure is async — so the node comes up with no listen address and no error. Worth keeping the reaper for one release even though nothing spawns p2pd any more.

2. **`core/crates/kwaai-cli/src/main.rs:703`** — the rewritten `node-health` socket resolution formats `/unix/{sock}` with no `#[cfg(unix)]` guard, unlike `node_native::control_socket_addr()`. On Windows with `KWAAINET_SOCKET` set it reports the node unreachable. The hunk fixed the no-override case and left the override branch wrong.

3. **`core/crates/kwaai-p2p-daemon/src/lib.rs:26`** — the new usage example calls `P2PClient::connect(DEFAULT_SOCKET_NAME)`, but `connect_stream` only accepts `/ip4/`, `/ip6/` or `/unix/` prefixes, so it always returns `Unsupported multiaddr format: /tmp/kwaai-p2pd.sock`. Should be `default_socket_addr()`. `rust,no_run` hides it from CI. Same snippet in that crate's `README.md:32`.

4. **`core/crates/kwaai-cli/src/config.rs:1212`** — `config set native_p2p` now bails, but `tests/kwaai-network/stress/README-metro.md` still instructs running it at lines 50, 109 and 151, and `docs/NATIVE_P2P_SECURITY.md:318` still calls `native_p2p: false` a correct default.

PR #192, opened 2026-09-04
chore(distrib): move nix/ to distrib/nix/

dazwin:
Moves `nix/` to `distrib/nix/` so every distribution mechanism has one home. Debian and RPM packaging follow into the same directory in later PRs; this one is deliberately pure motion so those rebase cleanly.

## The part worth reviewing

`crane.nix` and `proto.nix` reach the repository root by **relative path**, and they are a directory deeper now:

| File | Before | After |
|---|---|---|
| `crane.nix` (4 sites) | `./..` | `./../..` |
| `proto.nix:13` | `../core/…` | `../../core/…` |

A path that silently resolves to `distrib/` instead of the root fails a long way from the cause, so these were verified by evaluation rather than by reading (below). Every other `.nix` file is sibling-relative and moves as a unit.

## Left alone deliberately

- **The root `Makefile` and `.github/workflows/release-riscv.yml`** use flake *attribute* names (`nix build .#kwaainet`), never paths. No edit needed.
- **`distrib/nix/shell-functions/ascii-art.nix`** references `./docs/assets/…` CWD-relative at shell-hook runtime, not as a Nix path.
- **`docs/NATIVE_P2P_MIGRATION.md`** mentions `nix/p2pd.nix` and `nix/crane.nix` in four places. It is a historical migration record describing the layout at the time; rewriting it would add churn to a doc nobody greps for build paths.
- **`distrib/nix/README.md`** keeps six bare `nix/` strings that are `nixos.org/nix/install`, `/etc/nix/nix.conf` and `/nix/store/`. A blind `sed` over that file corrupts them.

## Verification

Nix isn't installed on the machine this was written on, so the flake was evaluated in a `nixos/nix` container against the moved tree:

| Output | Result |
|---|---|
| `packages.kwaainet` | `kwaainet-0.6.8.drv` — version read through `./../.. + "/core/Cargo.toml"` |
| `packages.p2pd`, `packages.protoRs` | evaluate |
| `kwaainet-container`, `kwaainet-all-container` | evaluate |
| `test-two-node`, `test-containers`, `checks.kwaainet-smoke` | evaluate |
| `devShells.default`, `formatter` | evaluate |
| `packages.x86_64-linux` attr names | all 29, including every cross target and `test-cross-smoke-*` |

The strongest signal: `multistream-select-0.13.0-slashless` and `libp2p-kad-0.48.0-multi-protocol` **built**, which only happens if both `core/patches/` paths resolve and the patches apply. The x86_64 enumeration exercises `cross.nix`, `overlays/cross-fixes.nix` and `tests/cross-smoke.nix` through their new import paths.

Also adds `distrib/README.md` — an index, plus the two footguns for whoever edits this next (flakes only see git-tracked files; the root-relative paths).


RezaRassool (commented):
## Review — no findings

Clean, and the only PR in the current queue I found nothing in. Verified well beyond the hunks:

- **Root-relative paths re-depthed correctly.** `distrib/nix/crane.nix` uses `./../..` at all four sites (`core/Cargo.toml`, both `core/patches/*.patch`, `craneLib.path .../core`); `distrib/nix/proto.nix:13` uses `../../core/crates/kwaai-p2p-daemon/proto`. From `distrib/nix/`, `../..` is the repo root — both right.
- **No other repo-relative path escaped.** Scanned every moved `.nix` for `..`, `src =`, `readFile` and path literals: `cross.nix`, `devshell.nix`, `tests/default.nix`, `containers.nix`, `packages.nix`, `p2pd.nix`, `overlays/cross-fixes.nix` and all `tests/*` are strictly sibling-relative and move as a unit. `tests/cross-smoke.nix` uses `pkgs.path`, which is nixpkgs rather than the repo.
- **All 13 import sites in `flake.nix` updated** — no `./nix/` string survives anywhere in the tree.
- **Nothing else references the old path.** Checked the Makefile, all 11 workflows, `.gitignore`, `.gitattributes`, `.dockerignore`, `Dockerfile.eve`, `docker-compose.eve.yml`, `scripts/`, `.githooks/`, `setup.sh`, `setup.ps1`, `KwaaiNet.code-workspace`, `core/Cargo.toml`. The workflows use flake *attribute* names only, so your note about that is accurate. No workflow `paths:` filter mentions `nix/**`.
- **Links resolve.** `distrib/README.md`'s `[nix/](nix/README.md)` correctly resolves to `distrib/nix/README.md`; the moved README has only absolute URLs, so nothing broke by going a level deeper.

Two things I looked at and decided were not findings: `docs/NATIVE_P2P_MIGRATION.md` still names `nix/p2pd.nix` and `nix/crane.nix`, which is the historical record you deliberately left alone; and no CI job builds the flake, so the move is unverified by CI — your container evaluation is the right substitute, since it actually *builds* both patched crates, which is the only thing that proves the `core/patches/` paths resolve.

**This is first in the merge order** — it unblocks #193 → #194, and #204's blocking fix needs to land in `distrib/nix/crane.nix` rather than the old path.

PR #193, opened 2026-09-04
fix(cli): never self-update or self-uninstall a packaged install

dazwin:
Stacked on #192. Prerequisite for the Debian/RPM packaging that follows.

## The problem

`KWAAINET_NO_AUTO_UPDATE` gates exactly one of four update paths — `maybe_auto_update()` in `node.rs`. The other three are harmless today and actively damaging the moment a `.deb` exists:

| Path | What it would do to a packaged install |
|---|---|
| `kwaainet update` | Runs the cargo-dist shell installer, which writes into `~/.cargo/bin`. That **shadows `/usr/bin/kwaainet` on PATH** — dpkg goes on reporting 0.6.8 while the user actually runs an unmanaged newer binary. |
| background update hint (`main.rs`) | Nags the user to run precisely that command. |
| `kwaainet uninstall` | `remove_binaries()` deletes `current_exe()` and its sibling `p2pd` — i.e. `rm /usr/bin/kwaainet` and `/usr/bin/p2pd` out from under dpkg. |

This is worth fixing on its own merits: `kwaainet update` ignoring `KWAAINET_NO_AUTO_UPDATE` is a bug regardless of packaging.

## The gate

```rust
pub fn packaged_install() -> Option<&'static str>   // Some("deb") | Some("rpm")
```

Returns `Some` only when **both** hold: `current_exe()` is exactly `/usr/bin/kwaainet`, **and** the marker `/usr/lib/kwaainet/packaged` parses as `deb` or `rpm`. Requiring both means a tarball install on a machine that *also* has the `.deb` cannot misfire. `OnceLock`-cached; `#[cfg(not(unix))]` returns `None`, since the path is meaningless on Windows.

`update` and `uninstall` now print the apt/dnf command and exit cleanly rather than erroring.

## Scope of the uninstall refusal

`run_uninstall()` also stops the daemon, removes the service unit and clears `~/.kwaainet/`. Those stay active for packaged installs — they are user-owned artefacts no package manager tracks. **Only the binary removal is declined**, and the pre-confirmation summary is reworded so it stops promising a removal that will not happen.

## Verification

```
cargo fmt --all -- --check          clean
cargo clippy --all-targets -D warnings   clean
cargo test -p kwaainet              254 passed, 0 failed, 1 ignored
```

Two new tests. `parse_marker` is a pure helper so the marker table (`deb`, `rpm`, `deb\n`, ` rpm `, ``, `apt`, `DEB`) is testable without mocking the filesystem or `current_exe()`; matching is **case-sensitive**, since our own packaging scripts write the file. The second test pins the invariant that detection declines under any test harness, where `current_exe()` is never `/usr/bin/kwaainet`.

### DIFF
--- a/core/crates/kwaai-cli/src/main.rs
+++ b/core/crates/kwaai-cli/src/main.rs
@@ -122,8 +122,10 @@ async fn main() -> Result<()> {
 
     // Spawn a background update check that runs concurrently with the command.
     // Uses a 24-hour on-disk cache so it only hits the network once per day.
-    // Skipped for `update` (redundant) and `run-node` (internal daemon process).
-    let skip_update_hint = matches!(cli.command, Command::Update(_) | Command::RunNode(_));
+    // Skipped for `update` (redundant), `run-node` (internal daemon process),
+    // and packaged installs, which cannot act on the hint anyway.
+    let skip_update_hint = matches!(cli.command, Command::Update(_) | Command::RunNode(_))
+        || updater::packaged_install().is_some();
     let update_task = (!skip_update_hint)
         .then(|| tokio::spawn(async { updater::UpdateChecker::new().check(false).await }));
 
@@ -893,6 +895,12 @@ async fn main() -> Result<()> {
         // -------------------------------------------------------------------
         Command::Update(args) => {
             print_box_header("🔄 KwaaiNet Update");
+            if let Some(cmd) = updater::packaged_upgrade_command() {
+                print_info("This kwaainet was installed from a system package.");
+                println!("  Upgrade it with: {cmd}");
+                print_separator();
+                return Ok(());
+            }
             let checker = updater::UpdateChecker::new();
             println!("  Current version: v{}", checker.current_version);
             println!("  Checking for updates…");
--- a/core/crates/kwaai-cli/src/main.rs
+++ b/core/crates/kwaai-cli/src/main.rs
@@ -122,10 +122,8 @@ async fn main() -> Result<()> {
 
     // Spawn a background update check that runs concurrently with the command.
     // Uses a 24-hour on-disk cache so it only hits the network once per day.
-    // Skipped for `update` (redundant), `run-node` (internal daemon process),
-    // and packaged installs, which cannot act on the hint anyway.
-    let skip_update_hint = matches!(cli.command, Command::Update(_) | Command::RunNode(_))
-        || updater::packaged_install().is_some();
+    // Skipped for `update` (redundant) and `run-node` (internal daemon process).
+    let skip_update_hint = matches!(cli.command, Command::Update(_) | Command::RunNode(_));
     let update_task = (!skip_update_hint)
         .then(|| tokio::spawn(async { updater::UpdateChecker::new().check(false).await }));
 
@@ -895,12 +893,6 @@ async fn main() -> Result<()> {
         // -------------------------------------------------------------------
         Command::Update(args) => {
             print_box_header("🔄 KwaaiNet Update");
-            if let Some(cmd) = updater::packaged_upgrade_command() {
-                print_info("This kwaainet was installed from a system package.");
-                println!("  Upgrade it with: {cmd}");
-                print_separator();
-                return Ok(());
-            }
             let checker = updater::UpdateChecker::new();
             println!("  Current version: v{}", checker.current_version);
             println!("  Checking for updates…");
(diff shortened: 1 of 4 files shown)


RezaRassool (commented):
## Review — four findings

Reviewed against its base `chore/distrib-nix-move` (#192). The gate design is sound — both exe path and marker must agree, all three self-update entry points (`Command::Update`, the background hint, `maybe_auto_update`) plus `remove_binaries` are covered, and the scoping claim checks out: the user-level systemd/launchd units it deliberately leaves active really are user-owned (`~/.config/systemd/user/`).

CI note: this PR has no test run — the only check is the release workflow's `plan` job, so `Test`/`Clippy`/`Format`/`Windows check`/`WASM` never ran.

### Findings

1. **`core/crates/kwaai-cli/src/updater.rs:613` — the gate is defeated by exactly the event it's meant to survive.** `detect_packaged_install()` doesn't strip the `" (deleted)"` suffix Linux appends to `/proc/self/exe`. After a dpkg upgrade renames over `/usr/bin/kwaainet`, a still-running daemon's first `packaged_install()` call compares `"/usr/bin/kwaainet (deleted)"`, caches `None` in the `OnceLock`, and self-updates into `~/.cargo/bin` — precisely the failure this PR prevents. `main.rs:1048` and `install_cuda_linux` both already strip this suffix, so the idiom exists in-tree.

2. **`core/crates/kwaai-cli/src/uninstall.rs:103`** — the new early return in `remove_binaries()` also skips the `~/.cargo/bin` / `~/.local/bin` cleanup loop (lines 150-168), so a packaged uninstall now leaves behind exactly the unmanaged, PATH-shadowing copies that no package manager can remove. Scoping the refusal to `exe` and its sibling `p2pd` would keep the cleanup.

3. **`core/crates/kwaai-cli/src/main.rs:977`** — pre-existing (present in the base) but inside the block this diff edits: on Unix `install_update()` runs at 977 and, on success, falls through to a second `install_update()` at 1063, so every `kwaainet update` downloads and installs **twice** — including the ~GB CUDA archive. Worth folding in while you're here.

4. **`core/crates/kwaai-cli/src/main.rs:903`** — the gate sits above the `args.check` branch, so read-only `kwaainet update --check` is refused too; combined with the suppressed background hint (`main.rs:129`), a packaged install can never learn a newer version exists. The hint's stated rationale ("cannot act on the hint anyway") is contradicted by `packaged_upgrade_command()` itself. Relatedly, `updater.rs:637` suggests `sudo apt install --only-upgrade kwaainet`, which presumes an APT repo — nothing in the repo publishes one, so a `dpkg -i` user gets "Unable to locate package kwaainet" and no working upgrade path.

dazwin:
Rebased onto main (#191/#192; one conflict in `uninstall.rs` where main reworded the confirmation line, kept main's wording inside the gate) and pushed `f397ed70`. All four addressed.

1. `current_exe_path()` strips the `" (deleted)"` suffix once, and `detect_packaged_install()`, `install_cuda_linux` and the update restart path all use it (the two hand-rolled copies are gone). Tests: the suffix is stripped, and `/usr/bin/kwaainet (deleted)` reads as packaged.
2. `remove_binaries()` skips only the packaged executable (and a legacy p2pd beside it) and still sweeps `~/.cargo/bin` / `~/.local/bin`. Tests for the packaged and unpackaged cases against a temp HOME; the confirmation prompt now says what is actually removed.
3. The first Unix `install_update()` block is deleted; the later `match install_update` is the single install and already handles the restart-on-failure.
4. The gate sits below the live check: a packaged install prints the version and release notes for both `update` and `update --check`, then the command. The background hint is no longer suppressed and points at `kwaainet update` for the command. The deb command is a `curl -fLO` of the release `.deb` (arch from `std::env::consts::ARCH`, version mangled as `distrib/packaging/version.sh` does, revision `-1`, checked against #194's build script) followed by `sudo dpkg -i`; there is no APT repo to promise. The rpm arm points at the release page. Tests cover the gate placement, the file name against version.sh, and that the hint never promises a self-update.

Knock-on for #194: `packaged_upgrade_command()` is gone and `test-install.sh --expect-update-gate` greps `kwaainet update` output for `apt install --only-upgrade`, which no longer appears (and was only ever reachable when GitHub reported a newer release than the freshly built package). Handled on that branch's rebase.

Checks: fmt clean, clippy clean, `cargo test -p kwaainet --bin kwaainet` 256 passed.

PR #194, opened 2026-09-04
feat(distrib): build a .deb for kwaainet

dazwin:
Stacked on #193 (which is on #192). Produces `kwaainet_<version>_<arch>.deb` for amd64 and arm64, attached to the GitHub Release and ready for the signed apt repo that follows.

## It repackages, it does not rebuild

The payload is the `kwaainet-<triple>.tar.xz` that `build-local-artifacts` has already produced, so the binary in the `.deb` are **byte-identical** to what tarball and installer users get — one artifact, two delivery mechanisms. A bug reported by a `.deb` user is provably the same build as a tarball user's.

## Tool choice

Stage a tree → `dpkg-shlibdeps` → `dpkg-deb --build --root-owner-group`, inside `debian:bookworm`.

| Rejected | Why |
|---|---|
| `cargo-deb` | Couples a second dist config into `Cargo.toml` beside `[workspace.metadata.dist]`, duplicates the build, and cannot see `p2pd` (a Go artifact). No `dpkg-shlibdeps`. |
| `dpkg-buildpackage` | A *source* package format: implies a network-free chroot, which is exactly the constraint that ruled out a Launchpad PPA (`core/patches/fetch-patches.sh` and `scripts/build-p2pd.sh` both need the network). |
| `nfpm` | Cannot run `dpkg-shlibdeps`, so every dependency becomes a hand-guess that rots silently as the runner's glibc moves. |

## Decisions worth reviewing

**No p2pd.** The Go daemon is gone upstream (#191); `stage-payload.sh` copies only `kwaainet`, and `test-install.sh` asserts the package never ships one.

**No service and no `/etc` file.** The package installs nothing under `/usr/lib/systemd` or `/etc`; `kwaainet start` runs as the invoking user with state in `~/.kwaainet`, and `test-install.sh` fails if a unit or conffile appears. A system service is deferred until the config layout has a config/state split (see README.md).

**The update gate is one file.** `/usr/lib/kwaainet/packaged` (`deb`) plus the `/usr/bin/kwaainet` path makes `kwaainet update` report and print a `dpkg -i` command instead of installing, and `kwaainet uninstall` leave the binary to `apt remove`.

**No maintainer scripts at all.** man-db has its own dpkg trigger and conffile handling is dpkg-internal, so `postinst`/`prerm` had nothing to do.

**`~/.kwaainet` survives remove *and* purge** — it holds the node identity key. Asserted by the test.

**`Depends` is derived, never written:** `libc6 (>= 2.34), libgcc-s1 (>= 4.2)`. Pinned in `expected-depends.amd64` and diffed by `package-linux`, so the supported-distro floor cannot move without someone deciding to move it. Support claim: **Debian 12+, Ubuntu 22.04+**.

## Verification (all run in containers)

| Check | Result |
|---|---|
| lintian | **Zero tags** of any severity after overrides |
| install/remove on `debian:bookworm`, `ubuntu:22.04` | pass |
| `~/.kwaainet` survives remove | pass |
| `--network none` | installs and runs; nothing phones home |
| **`debian:bullseye` negative control** | fails with `Depends: libc6 (>= 2.34) but 2.31 is to be installed`, apt exit 100, no binary installed — an apt dependency error, **not** a runtime `GLIBC` crash |
| shellcheck | clean on all scripts |
| CI stub-payload path | full build+install cycle in 23 s |

**The update gate was tested for real**, not assumed: a `.deb` built from a branch-compiled Linux binary, installed, then

- marker present at `/usr/bin/kwaainet` → `update` and `uninstall` both defer to apt ✓
- marker deleted → gate correctly does **not** fire ✓
- binary copied off `/usr/bin` with the marker still present → gate correctly does **not** fire ✓

Both halves of #193's AND condition, verified independently.

## CI

- **`package-linux`** (release.yml) — matrix amd64/arm64, uploads as `packages-<arch>`. That name deliberately avoids the `artifacts-*` prefix the `host` job globs into release creation.
- **`upload-packages-to-release`** — modelled on the existing `upload-cuda-to-release`; `continue-on-error: true` and *not* in `host.needs`, so packaging can never block or delay a release.
- **`package-smoke`** (ci.yml) — builds from a stub payload in seconds; `ci.yml` deliberately has no release build and this does not add one.
- **`packaging-guard`** — fails if `package-linux` disappears from `release.yml`. The realistic failure mode is someone running `dist init`/`generate` without `--allow-dirty ci`, which would silently delete the hand-added block.

`publish-packages-repo` is left as a commented-out placeholder — `Kwaai-AI-Lab/packages` does not exist yet, and a job that fails is worse than none.


RezaRassool (commented):
## Review — five findings, none in the payload logic

Two things I expected to be wrong turned out to be right, so they're worth stating so they aren't re-litigated later:

- **`expected-depends.*` is correct for both arches.** I downloaded the real v0.6.8 `x86_64` and `aarch64` linux-gnu tarballs: the released binaries need only `libc.so.6`, `libgcc_s.so.1`, `libm.so.6`, and max symbol versions are `GLIBC_2.34` and `GCC_4.2.0` on **both** — so `libc6 (>= 2.34), libgcc-s1 (>= 4.2)` and the Debian 12+/Ubuntu 22.04+ claim hold. rustls rather than native-tls, so no `libssl3` creeps in.
- **`stage-payload.sh`'s "exactly one top-level dir" assumption is correct** — the real archive is `kwaainet-<triple>/{kwaainet,p2pd,README.md}`, and release.yml's p2pd injection re-tars with that layout.
- The marker path and contents (`/usr/lib/kwaainet/packaged` = `deb\n`) match `updater::detect_packaged_install()` exactly, and the strings `test-install.sh` greps for are substrings of the real `packaged_upgrade_command()`/`packaged_remove_command()` output.

### Findings

1. **`distrib/packaging/kwaainet.1:95` — the man page documents three files the package guarantees are absent.** ENVIRONMENT says `KWAAINET_NO_AUTO_UPDATE` is "Set by default in `/etc/default/kwaainet`", and FILES lists `/etc/default/kwaainet` and `/usr/lib/systemd/user/kwaainet.service`. Commit 260e2dcd removed both from the package — and `test-install.sh:693` now *fails the build* if anything under `/etc/` or `/usr/lib/systemd/` ships. It also misattributes the update gate to an env var in that nonexistent file rather than the `/usr/lib/kwaainet/packaged` marker that actually implements it, so a user following it will `sudo editor /etc/default/kwaainet` and get an empty new file that does nothing. Line 20 likewise still says the node is "backed by `p2pd(1)`, a libp2p daemon shipped alongside this binary", which 4b3f026e reversed — and cross-references a `p2pd(1)` page that doesn't exist.

2. **`distrib/packaging/stage-payload.sh:22`** — dropping p2pd is safe only for the default config. `DEFAULT_NATIVE_P2P = true` (`config.rs:982`), but `native_p2p: false` is an explicit opt-out the code still promises to honour, and #191 (which actually removes p2pd) is still open. For such a user, `kwaainet start` on a `.deb` hits `node.rs:187-189`, prints "p2pd not found — run `kwaainet setup --get-deps` to install it", and continues into a path that can't work. Worse, that remedy resolves to `setup::get_dependencies()` (`setup.rs:11-19`), which installs p2pd into `dirname(current_exe())` — i.e. **`/usr/bin/p2pd`**: EACCES for a normal user, and an untracked binary in a dpkg-owned directory under `sudo`. #193 gated `update` and `uninstall` on `packaged_install()` but not `setup --get-deps`. Note this finding disappears if #191 merges first, which is where it currently sits in the order.

3. **`distrib/debian/build-deb.sh:111`** — `lintian --no-tag-display-limit "${DEB}" || true` means `--lintian` can never fail, while both `README.md` ("Target is zero errors after overrides… If a *new* tag appears, that is a real change") and the `package-smoke` job comment describe a gate that exists. A future error tag, or a stale entry in `lintian-overrides` turning into `unused-override`, scrolls past in green. `lintian --fail-on error` would match the documented intent.

4. **`distrib/debian/build-deb.sh:105`** — `dpkg-deb --build` doesn't generate `DEBIAN/md5sums` (normally `dh_md5sums`' job), so the package ships none and `dpkg -V kwaainet` / `debsums kwaainet` silently verify nothing — they report success on a truncated or tampered `/usr/bin/kwaainet`. One line before the build fixes it: `(cd "${STAGE}" && find . -type f ! -path './DEBIAN/*' -printf '%P\0' | xargs -0 md5sum > DEBIAN/md5sums)`.

5. **`distrib/debian/test-install.sh:10`** — `EXPECT_UPDATE_GATE` is read from `$2` only, but release.yml passes an unexpanded glob first: `test-install.sh /src/packages/kwaainet_*_amd64.deb --expect-update-gate`. If `packages/` ever holds more than one matching `.deb` (a re-run, a leftover from a version bump), the flag lands in `$3`, the script prints "update-gate assertions skipped" and **passes** — so the one property this PR most wants proven is silently not tested. Scanning all arguments for the flag rather than taking it positionally would fix it.

   Related: **`distrib/packaging/version.sh:12`** — the documented primary version source is dead in CI. `docker run` in `package-linux` passes no `-e GITHUB_REF_NAME`, and `debian:bookworm` + `dpkg-dev`/`lintian` has no `git`, so every CI-built `.deb` version comes from the `core/Cargo.toml` fallback. Harmless while cargo-dist enforces tag == workspace version, but if they diverge, `upload-packages-to-release` attaches a differently-versioned `.deb` to the tag with nothing catching it — in contrast to the deliberate `expected-depends` drift check.

### One for the description rather than the code

The PR body is stale relative to its own last two commits: it still argues "`p2pd` goes in `/usr/bin`" and "A systemd *user* unit, installed but not enabled" under "Decisions worth reviewing", both of which 4b3f026e/260e2dcd reversed. `distrib/debian/README.md` has the accurate layout table.

dazwin:
Rebased onto the new `fix/no-self-update-when-packaged` head (f397ed70, itself on main with #191/#192) and pushed `cd6d834c`. No conflicts; the five original commits are preserved with three review commits on top.

1. **Man page — fixed.** No `/etc/default/kwaainet`, no systemd unit; FILES documents `/usr/lib/kwaainet/packaged` plus the `/usr/bin/kwaainet` path as the gate; `KWAAINET_NO_AUTO_UPDATE` is described as the developer escape hatch it is; p2pd(1) and the p2pd mentions are gone; SEE ALSO gains apt(8)/dpkg(1). Renders clean under `mandoc -T lint`. The same commit updates README.Debian/README.md for #193's change (`update` prints a `dpkg -i` command rather than refusing).
2. **Moot after #191.** `setup.rs` no longer exists, `setup --get-deps` prints "p2pd is no longer required", and every remaining `current_exe()` use in kwaai-cli is read-only apart from `install_update`, which sits behind the gate. Nothing writes into `dirname(current_exe())` on a packaged install.
3. Fixed: `lintian --fail-on error`, no `|| true`; README states the gate.
4. Fixed: `DEBIAN/md5sums` generated (sorted, `LC_ALL=C`, excludes `DEBIAN/`) before `dpkg-deb --build`. `test-install.sh` asserts the installed list names `usr/bin/kwaainet`, that `dpkg -V` passes fresh, and that it fails after the binary is appended to.
5. Fixed: `test-install.sh` scans all arguments for `--expect-update-gate`, errors on an unknown flag and on a second positional `.deb` (the glob case). The gate block now runs `update --check` and `update`, asserts no self-install, an unchanged `/usr/bin/kwaainet` checksum and no `~/.cargo/bin/kwaainet`, requires the `dpkg -i` command if a newer release is reported, and still requires `uninstall` to name `apt remove`. `release.yml` passes `GITHUB_REF_NAME` into the build container, and a tag-derived version that disagrees with `core/Cargo.toml` fails the build unless `KWAAINET_VERSION` overrides (checked locally: fallback → 0.6.8-1, `v9.9.9` → exit 1, branch names fall through).

PR description updated for the reversed p2pd and user-unit decisions.

Not executed here (no dpkg or lintian on a Mac): `build-deb.sh`, `test-install.sh`, the lintian gate. Two things to watch on the first CI run: the `package-smoke` stub payload may surface a lintian error the old `|| true` hid, and `kwaainet update` in `package-linux` exits non-zero if the GitHub API is unreachable, the same exposure the old grep had. `bash -n` passes on all three scripts; fmt clean; no Rust changed.

dazwin:
The first package-smoke run failed on the new tamper assertion: `dpkg -V` prints a `??5??????` line for a checksum mismatch but exits 0 either way, so the check could never fire and the fresh-install check was vacuous. Both now judge the output rather than the exit status (`d2dff7a7`). Lintian's `--fail-on error` and the md5sums generation passed on that run.

PR #197, opened 2026-09-06
fix(kad): evict peers that turn out not to speak our kad name

dazwin:
## Summary

Every kwaainet node on 0.6.8 dials **~150 kubo/IPFS peers every 5 minutes**, and the bootstraps hand those peers to every joiner. This is why the Amino connections never drained after the `/kwaai/kad` cutover — it isn't kubo finding us, it's **our routing tables being poisoned with kubo entries that nothing ever removes**.

One change in `kwaai-p2p`, a single `else if` in `handle_identify_event`: when a peer's advertised protocols do **not** include our kad name, `kad.remove_peer(&peer_id)`.

Plus two regression tests and a correction to the V1Lazy comment, which claimed identify "corrects" a false insert. It doesn't — this makes it true.

## How it was found

A stock v0.6.8 probe (fresh identity, `RUST_LOG=libp2p_kad=debug,libp2p_swarm=debug`) joining production for 516 s:

| | |
|---|---|
| connections established | 366 — **328 to `:4001`**, all `Dialer`, zero inbound |
| distinct kubo peers dialled | **160** |
| identify agents | 303 `kubo/*` + 16 `go-ipfs/*` vs 12 `kwaainet/*` |
| first kubo dial | **330 ms** after connecting, with *only* the two bootstraps connected — so the addresses came from the bootstraps' own FIND_NODE replies |
| kubo-side lifetime | mean 31 s (kubo's connmgr trims us after its 20 s grace) |
| cadence | bursts of 60–90 dials every ~5 min — kad's `periodic_bootstrap_interval` |
| `Bucket full. Peer not added` | 114×, **all kubo**, the first firing **0.2 ms after connect** — before any round-trip |

That last line is the tell: a peer was being inserted before the remote had answered anything.

## Root cause (libp2p-kad 0.48.0 / libp2p-swarm 0.47.1 / multistream-select 0.13.0)

- `service.rs` sets `with_substream_upgrade_protocol_override(V1Lazy)` swarm-wide for 0-RTT unary requests.
- multistream-select's V1Lazy returns `Negotiated::expecting(...)` **immediately** when exactly one protocol is offered (`dialer_select.rs:145`) — no round trip.
- kad's handler takes that as proof: *"Upon the first successfully negotiated substream, we know that the remote is configured with the same protocol name"* → `ProtocolConfirmed` → `connection_updated(Connected)` → **inserted**. That is the only insert path under `BucketInserts::OnConnected` (`behaviour.rs:2297`).
- Identify then reports the real protocol set → `ProtocolNotSupported` → `connection_updated(Disconnected)`. The `Entry::Present` branch **only flips status; it never removes**. `address_failed` explicitly keeps the last address. `find_closest_local_peers` filters only `!= source`, so Disconnected kubo entries are served in every FIND_NODE reply. With ~18 fleet peers there is no bucket pressure to ever evict them.

**The dual-protocol build was immune**: with two names V1Lazy only shortcuts the *last* offer, so the preferred name negotiated eagerly and kubo's `na` was visible. Cutting over to the single-name stock build is what switched this on. The comment block in `service.rs` predicted exactly this and deferred it — *"gating laziness on identify's protocol set … Tracked as a follow-up, not done here."*

## Why evict-on-identify, and not `BucketInserts::Manual`

libp2p-swarm 0.47 has no per-behaviour upgrade-version override, so V1Lazy can't be turned off for kad alone without losing the 0-RTT win on unary.

I tried `Manual` first (identify as the only door). It is the go-libp2p shape, but it broke three DHT-resolution tests deterministically, and the reason is structural: the routed-dial path says *"the walk has populated the routing table with everything it found — flush decides from there"*. That population **is** the OnConnected insert on the walk's dial. Under `Manual` a walk-discovered target only enters the table when identify lands — a race against `flush_routed` in production, and never at all in loopback tests (identify refuses `127.0.0.1`). The AutoNAT dial-back tables NATed peers the same way. So the insert stays; what changes is that a peer which turns out not to speak our kad name is evicted the moment identify says so.

The cost is a poison window of one round trip per foreign peer, during which a FIND_NODE from a third party could pick it up. Against the current state — entries that live for the process lifetime and are served on every reply — that is the difference between a leak and a puddle.

## Tests

- `a_dialled_peer_with_a_foreign_kad_name_is_purged_on_identify` — `dial()` seeds B; identify says B speaks only `/ipfs/kad/1.0.0`; B must be gone. Exercises the purge.
- `a_foreign_kad_peer_served_by_a_neighbour_is_never_tabled` — the production shape: C holds B by fiat (never connects, so never purges), A walks to C, is handed B, dials it, gets the V1Lazy false confirm — and once identify lands, B must be gone.

Falsification — both tests run with `src/` stashed (the fix removed) **fail**, and pass with it:

```
test a_dialled_peer_with_a_foreign_kad_name_is_purged_on_identify ... FAILED
  B speaks only /ipfs/kad/1.0.0 and must not be in A's routing table: [PeerId("12D3KooWANdw…")]
test a_foreign_kad_peer_served_by_a_neighbour_is_never_tabled ... FAILED
  B was handed to A by C and dialled, but speaks only /ipfs/kad/1.0.0 — it must not be tabled: [PeerId("12D3KooWExWj…"), PeerId("12D3KooWMRTp…")]
```

`spawn_swarm_with_kad_protocols` is no longer feature-gated: a single foreign name works on every build (`kad::Config::new` takes the first name unconditionally); only the multi-name bridge needs `kad-multi-protocol`.

## What this does and doesn't do

- **Fixed stock nodes self-clean; the bootstraps do not.** A stock (single-name) node evicts every foreign peer the moment identify answers, and re-dials its stale entries into the purge on each 5-minute bootstrap walk. The `kad-multi-protocol` bootstrap build cannot tell kubo from a legacy-only 0.6.7 peer by protocol name (`/ipfs/kad/1.0.0` is what both speak), so the purge is a no-op there and the bootstraps keep serving their poisoned set on every walk. A fixed node's *table* is clean at every instant, but its kubo dial volume does **not** decay until the bootstraps either drop the legacy name or gain a kubo-specific eviction; the 2026-09-06 live run (1002 kubo dials in 480 s on the fixed build) is that ceiling measured.
- **Bootstraps-only is not enough.** Long-running fleet nodes keep their poisoned tables (no eviction path on the old binary) and re-infect each other and every joiner at hop two. This needs the release rollout; mixed fleets are safe (routing-table policy is purely local, no wire change).
- **Does not explain the ~230 kubo connections the bootstraps hold for 7+ hours** while a fleet node's die in 30 s. Separate hypothesis (kubo autorelay reserving on public peers advertising hop); separate test.
- `ping` under V1Lazy (re-pinging a non-ping peer forever) remains as documented — same root, no fix here.

## CI-equivalent run

Run in an **isolated `CARGO_TARGET_DIR`**, and that matters: another Claude session has a worktree of this repo (`feat/kad-client-mode`) with no `core/target` of its own, so both were compiling into `core/target/debug` and the test-binary hashes collide (`swarm-4a8cf5fccb77d338` in both). For an hour my test results depended on whose `libkwaai_p2p` was linked last. Everything below is from a clean, private target dir.

- `cargo fmt --all --check` — clean
- `cargo clippy --all-targets -- -D warnings` — clean
- `cargo test -p kwaai-p2p` — all green; `swarm`: 17 passed (incl. both new tests)
- `cargo test -p kwaai-p2p --features kad-multi-protocol` — all green; `swarm`: 19 passed (incl. both new tests)
- `cargo test -p kwaainet` — 252 passed, 1 ignored

Falsification (same isolated dir, `src/` stashed → `main`): both new tests **FAIL**; with the fix: pass.

## Overlap

`feat/kad-client-mode` (in flight, another session) edits the same neighbourhood — kad mode selection in `behaviour.rs`, and two `remove_peer` calls in `service.rs`. Expect a small, mechanical conflict on whichever lands second; the intent is compatible (a client-mode node advertises no kad name at all, so this purge would evict it from peers' tables — which is exactly what that branch's `a_kad_client_is_never_tabled_and_never_found_by_a_walk` wants).

### DIFF
--- a/core/crates/kwaai-p2p/src/service.rs
+++ b/core/crates/kwaai-p2p/src/service.rs
@@ -68,6 +68,10 @@ const LAST_CONNECTED_CAP: usize = 1024;
 /// How often the relay manager retries candidates whose backoff has expired.
 const RELAY_TICK_INTERVAL: Duration = Duration::from_secs(15);
 
+/// Upper bound on `pinned_addrs` peers. Explicit dials only, so a memory
+/// bound rather than a policy; oldest pin evicted, same shape as `last_connected`.
+const PINNED_PEERS_CAP: usize = 1024;
+
 /// Cap on routing-table addresses per peer.
 ///
 /// kad's own `Addresses` is an unbounded `SmallVec`: `insert` appends whatever
@@ -145,6 +149,13 @@ pub struct NetworkService {
     /// recently" is the signal that distinguishes the two. Recency cache, not
     /// a ledger — capped at [`LAST_CONNECTED_CAP`], oldest evicted.
     last_connected: HashMap<PeerId, Instant>,
+    /// Addresses we were *told* to dial — `dial()` and `AddKadAddress`, the
+    /// two uncapped seeds — stored bare, with when the peer was first pinned.
+    /// The routing table is the only other address book for dial-by-PeerId,
+    /// and the non-kad purge in `handle_identify_event` empties it for a
+    /// peer that speaks a foreign kad name. Consulted by
+    /// [`Self::candidate_addresses`] so such a peer stays reachable by id.
+    pinned_addrs: HashMap<PeerId, (Instant, Vec<Multiaddr>)>,
     /// Addresses peers reported observing us at → the set of peers that said so.
     /// A set (not a counter) so repeated identifies from one peer count once.
     observed_addrs: HashMap<Multiaddr, HashSet<PeerId>>,
@@ -443,6 +454,7 @@ impl NetworkService {
             routed_attempts: HashMap::new(),
             connections: HashMap::new(),
             learned_addrs: LearnedAddrs::new(local_peer_id),
+            pinned_addrs: HashMap::new(),
             last_connected: HashMap::new(),
             observed_addrs: HashMap::new(),
             require_global_ips: config.require_global_ips,
@@ -750,11 +762,13 @@ impl NetworkService {
                 // the operator naming an address, not a peer claiming one. A
                 // symmetric-NAT peer flooding identify must never be able to
                 // evict a bootstrap address someone configured by hand.
+                self.pin_address(peer, strip_dest_p2p(&addr));
                 self.swarm.behaviour_mut().kad.add_address(&peer, addr);
                 let _ = reply.send(());
             }
 
             Command::RemoveKadPeer { peer, reply } => {
+                self.pinned_addrs.remove(&peer);
                 let existed = self.swarm.behaviour_mut().kad.remove_peer(&peer).is_some();
                 let _ = reply.send(existed);
             }
@@ -978,6 +992,7 @@ impl NetworkService {
                 // Uncapped for the same reason as `AddKadAddress`: an address
                 // we are actively dialing is our own intent, not a remote
                 // claim, and is the one entry least worth evicting.
+                self.pin_address(peer, stripped.clone());
                 self.swarm.behaviour_mut().kad.add_address(&peer, stripped);
             }
         }
@@ -1170,6 +1185,15 @@ impl NetworkService {
             }
         }
 
+        if let Some((_, pinned)) = self.pinned_addrs.get(peer) {
+            let fresh: Vec<Multiaddr> = pinned
+                .iter()
+                .filter(|a| !addrs.contains(a) && table_filter(a))
+                .cloned()
+                .collect();
+            addrs.extend(fresh);
+        }
+
         if let Some(conns) = self.connections.get(peer) {
             for conn in conns.values() {
                 if !addrs.contains(&conn.addr) {
@@ -1206,6 +1230,35 @@ impl NetworkService {
         addrs.retain(|a| seen.insert(a.clone()));
 
         addrs
+    }
+
+    /// Remember an address the operator supplied for `peer`; see `pinned_addrs`.
+    /// Bounded like the routing table: six per peer, oldest out.
+    fn pin_address(&mut self, peer: PeerId, addr: Multiaddr) {
+        if addr.is_empty() {
+            return;
+        }
+        if self.pinned_addrs.len() >= PINNED_PEERS_CAP && !self.pinned_addrs.contains_key(&peer) {
+            if let Some(oldest) = self
+                .pinned_addrs
+                .iter()
+                .min_by_key(|(_, (at, _))| *at)
+                .map(|(p, _)| *p)
+            {
+                self.pinned_addrs.remove(&oldest);
+            }
+        }
+        let (_, addrs) = self
+            .pinned_addrs
+            .entry(peer)
+            .or_insert_with(|| (Instant::now(), Vec::new()));
+        if addrs.contains(&addr) {
+            return;
+        }
+        if addrs.len() >= MAX_ADDRESSES_PER_PEER {
+            addrs.remove(0);
+        }
+        addrs.push(addr);
     }
 
     /// Our own listen and confirmed-external addresses, snapshotted from the
@@ -2336,7 +2389,8 @@ impl NetworkService {
                 {
                     // Tabled by the V1Lazy false confirm (or `dial()`) before
                     // identify could say no; left in place it is served to every
-                    // FIND_NODE caller.
+                    // FIND_NODE caller. Table membership only: an address the
+                    // operator supplied survives in `pinned_addrs`.
                     debug!(peer = %peer_id, agent = %info.agent_version, "dropped non-kad peer from the routing table");
                 }
 
--- a/core/crates/kwaai-p2p/src/service.rs
+++ b/core/crates/kwaai-p2p/src/service.rs
@@ -151,10 +151,11 @@ pub struct NetworkService {
     last_connected: HashMap<PeerId, Instant>,
     /// Addresses we were *told* to dial — `dial()` and `AddKadAddress`, the
     /// two uncapped seeds — stored bare, with when the peer was first pinned.
-    /// The routing table is the only other address book for dial-by-PeerId,
-    /// and the non-kad purge in `handle_identify_event` empties it for a
-    /// peer that speaks a foreign kad name. Consulted by
-    /// [`Self::candidate_addresses`] so such a peer stays reachable by id.
+    /// Neither seed reaches `learned_addrs` (only `ConnectPeerWithAddrs`
+    /// does), and a learned address is forgotten on a failed dial while an
+    /// operator's must not be; the routing table is the seeds' only other
+    /// home, and the non-kad purge in `handle_identify_event` empties it.
+    /// Consulted by [`Self::candidate_addresses`], which dedupes both stores.
     pinned_addrs: HashMap<PeerId, (Instant, Vec<Multiaddr>)>,
     /// Addresses peers reported observing us at → the set of peers that said so.
     /// A set (not a counter) so repeated identifies from one peer count once.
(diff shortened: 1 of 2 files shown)


dazwin:
## Live validation against production (fixed build, 8 min)

Same probe setup as the diagnosis — isolated `KWAAINET_HOME`, fresh non-announcing client, `RUST_LOG=kwaai_p2p=debug,libp2p_kad=debug` — but running **this branch** (debug build, sha `0ede6689…`).

| | stock 0.6.8 (516 s) | **this PR (480 s)** |
|---|---|---|
| `dropped non-kad peer from the routing table` | n/a | **978** — 175 distinct peers |
| …by agent | | 906 `kubo/*`, 66 `go-ipfs/*`, 3 `nabu`, 2 `myjoypin`, 1 `harmony` |
| **`kwaainet/*` peers evicted (control)** | | **0** |
| `Bucket full. Peer not added` | 114 | **0** |
| distinct kubo peers dialled | 160 | 180 |
| kubo dials | 328 | **1002** |
| 5-min re-walk: same kubo set? | | 139 of 165 re-dialled |

**What works:** every foreign peer is evicted the moment identify answers, selectively — not one kwaainet peer was touched across 978 evictions. `Bucket full` drops from 114 to 0: kubo entries no longer occupy the far buckets, so real peers always have room. The routing table of a fixed node is clean at every instant beyond the one-RTT window.

**What it exposes — read this one:** until the *neighbours* are fixed, a fixed node dials **more** foreign peers, not fewer (1002 vs 328 over comparable windows; different times of day, so treat the ratio as indicative). The bootstraps still serve the same poisoned set on every walk (139 of 165 peers re-appeared in the 5-min re-walk), and evicting an entry also discards kad's memory that it was useless, so each walk treats it as fresh. On stock, the stale Disconnected entry at least carried a failed-address mark. The kubo side pays ~20 s of connection per dial.

This is the rollout argument from the description, now measured from the fixed side: **a fixed node cannot stop being handed garbage until its neighbours stop holding it.** It converges to zero when the last poisoned node upgrades.

Possible follow-up, not in this PR: remember evicted peers briefly (`libp2p-allow-block-list`, ~1 h) so walks skip them instead of re-dialling — the blocklist half of #174. Worth doing if the rollout is slow; unnecessary if it's a week.

The probe is stopped.

RezaRassool (commented):
## Review — rework, two findings

Ran the full `kwaai-p2p` suite in an isolated `CARGO_TARGET_DIR` — all green, 17/17 in `swarm` including both new tests. The mechanism you describe checks out: `kad::Handler::listen_protocol` / `BucketInserts::OnConnected` and `remove_peer` semantics match the write-up, and the migration matrix (stock ↔ dual ↔ legacy) is handled correctly by `speaks_kad`.

Two findings, and together they mean the fix takes away something it shouldn't while not firing where it needs to.

### Findings

1. **`core/crates/kwaai-p2p/src/service.rs:2177` — the purge discards operator-supplied addresses.** `remove_peer` deletes the kbucket entry **and its `Addresses`**, and that entry is the node's only persistent address book for dial-by-PeerId — `candidate_addresses` (and so `dial_candidates`/`known_addresses`) reads straight out of the kbucket.

   Concrete: the documented workaround for reaching metro-win (0.6.7, legacy kad only) is `p2p peers connect --addr /ip4/100.69.49.42/…`. `dial()` seeds that address into kad, then metro-win's identify triggers the purge and it's gone. It survives only as long as the live connection (~30 s idle), after which every routed `mux://` call falls through `dial_routed` → `NoAddresses` → DHT lookup → "peer not found in DHT". `reseed_lost_bootstraps` can't restore it, since it skips peers we're still connected to.

   This is effectively a third eviction rule that isn't evidence about dialability — which is exactly what the comment at `service.rs:1752-1767` argues eviction must be. Fix shape: drop table membership without discarding addresses (re-`add_address` the returned `EntryView`'s addresses).

2. **`service.rs:2161` — the fix is a no-op on the nodes that cause the problem.** `speaks_kad` tests the *local* name set, so on the `kad-multi-protocol` bootstrap build (`[/kwaai/kad/1.0.0, /ipfs/kad/1.0.0]`, `config.rs:259-266`, `docs/BOOTSTRAP.md:52-53`) every kubo peer passes — the purge never fires there, and the `if` branch actively files kubo's identify listen addrs.

   So a fixed stock node still gets ~150 kubo addresses per 5-minute bootstrap walk and still dials them all: the reported symptom persists, and only table retention improves. A bootstrap can't tell kubo from a real 0.6.7 peer by protocol name, so this is a rollout-claim problem rather than a condition to tighten blindly — but the PR body's "hits zero when the last one does" needs correcting either way.

dazwin:
Rebased onto main (#191/#192, clean) and pushed `abbffd11`.

**1. Purge discards operator-supplied addresses — fixed.** The purge now removes table membership only. Addresses seeded by our own dial (`connect --addr`, `AddKadAddress`) are kept in a small pinned map that `candidate_addresses` reads beside the table (6 per peer, 1024 peers, oldest-out; cleared by `remove_kad_peer`). Not the literal re-`add_address` shape: that re-tables the peer and undoes the purge, and stashing whatever the evicted entry held would let 175 kubo purges per 8 min flush an operator's pin out of any bounded map. Test `an_operator_supplied_address_outlives_the_purge`: connect a legacy-only peer by explicit address, let identify purge it, disconnect, call it by PeerId — fails without the fix (NoAddresses → lookup → not found), passes with it.

**2. No-op on the `kad-multi-protocol` build — agreed, claim corrected.** The bootstrap build cannot tell kubo from a legacy-only 0.6.7 peer by protocol name, so the purge does not fire there and a fixed stock node's kubo dial volume does not decay until the bootstraps drop the legacy name or gain a kubo-specific eviction. PR body updated to say so. The cheap correct signal on that build is agent-version prefix (`kubo/`, `go-ipfs/`) for a peer that speaks only the legacy name — the 2026-09-06 run's 978 evictions were 972 of those; the remaining six (`nabu`, `myjoypin`, `harmony`) need the legacy name gone. That is one `if` gated on `KAD_MULTI_PROTOCOL_BUILD`, kept out of this PR so it stays policy-only; happy to open it separately.

Checks: fmt clean, clippy clean, `cargo test -p kwaai-p2p` 104 tests green (swarm 18/18).

RezaRassool:
## Rebase needed — one file

#185 reworked the same region: it added `LearnedAddrs` and changed how `candidate_addresses` assembles a peer's addresses. This PR's `pinned_addrs` store solves the same problem from the other direction — keeping an operator-supplied address alive when the kad entry is purged.

**Worth a look while rebasing:** with `learned_addrs` now also feeding `candidate_addresses`, check whether `pinned_addrs` and `learned_addrs` overlap. They may be complementary (operator-supplied vs peer-published), or `pinned_addrs` may now be doing less work than it was. Either is fine — but it is cheaper to decide now than after both have shipped.

---

dazwin:
Rebased onto main `8c5660b7` and pushed `689c991a`. One conflict, the struct initialiser: kept `pinned_addrs` beside `learned_addrs`.

On the overlap question, read from main's insert sites rather than the commit messages: `learned_addrs` has exactly one insert, under `ConnectPeerWithAddrs` (service.rs:598), the split form used by shard code and signed peer records. An operator's `p2p peers connect --addr` goes `ConnectPeer` → `dial()` → `kad.add_address` only, and `AddKadAddress` likewise; neither reaches `learned_addrs`, and the purge never touches it. So the two stores are complementary, and `LearnedAddrs::forget` dropping an address after one failed dial would be the wrong semantics for an operator seed anyway. `pinned_addrs` stays. A caller using both paths is safe: `candidate_addresses` dedupes on main and the pinned block filters against it.

Proof it is not redundant: with the `pinned_addrs` hunks reverse-applied on top of main's `LearnedAddrs`, `an_operator_supplied_address_outlives_the_purge` fails with `peer not found in DHT (no addresses)`; restored, it passes. `689c991a` is doc-only: the field comment claimed the routing table was the only other address book, which is false on main.

Checks: gate clean, fmt, clippy, `kwaai-p2p` 246 passed.

Issue #198, opened 2026-09-08
Plan: make peers findable and reachable again (bootstrap deploy + 0.6.9)


dazwin:
## Summary

Sharded inference on production fails because peers cannot be **found** or **reached**, not because inference is broken. Every unreachable case measured on 2026-09-06/07 traces to Amino (kubo) peers occupying our routing tables and relay slots, plus two small selection bugs on the inference side. This issue is the plan to fix it: what changes, which run on the bootstraps, which ship to peers as 0.6.9, and in what order.

## What was measured

- A live node's chain build showed 7 of 10 peers "unreachable", each after a 10–13 s timeout: `Dial failed: <peer>: peer not found in DHT (no addresses)`.
- bootstrap1 held **386** connections, **222** to kubo (port 4001); bootstrap2 **618 / 371**. Only 6 connections each on kwaainet ports.
- The unreachable peers (metro, draak, christophermayfield) were **connected to both bootstraps** at the time, yet absent from their routing tables and holding no relay reservation there. An inbound peer is tabled only if its k-bucket has room; the buckets are full of kubo entries.
- Cause of the kubo entries: the swarm-wide `V1Lazy` negotiation makes our dialer mark `/kwaai/kad` "confirmed" before the remote answers, so every kubo peer ever dialed was inserted and never removed, then served in every FIND_NODE reply (#197 describes it).
- Relay slots drift: the relay manager treats every hop-advertising peer as a candidate in a 16-entry oldest-out list, so the bootstraps fall out once 16 kubo peers identify. When a reservation ends the slot goes to the next candidate within ~200 ms, while the bootstrap sits in a 30 s backoff. Measured twice; pinning the bootstraps in `trusted_relays` only wins the startup race.
- On the same node, kubo relay reservations lasted a mean of 104 min; **bootstrap reservations a mean of 31 min, two thirds ending within 15 min**. The bootstraps drop reservations early (cause still open).
- Inference itself works. On an isolated 3-node LAN (Mac, Mac mini, Jetson Orin as bootstrap+relay) a 2-hop `shard run` produced correct text: direct, relayed through our relay, and relayed through a **stock 0.6.8** relay with the direct path firewalled. Same hop protocol as production.
[...]

PR #199, opened 2026-09-08
fix(kad): answer FIND_NODE from a peerstore, as the Go bootstraps did

dazwin:
Part of #198 (item 2). Relates to #179.

## Problem
rust-libp2p answers an inbound FIND_NODE from its k-buckets alone. On production the bootstraps' buckets are full of Amino peers (222 and 371 kubo connections on 2026-09-07), so fleet peers that dial in get no slot, and every walk for them fails with `peer not found in DHT (no addresses)` while the peer is connected to the very bootstrap being asked. The Go p2pd bootstraps did not have this gap: go-libp2p's `handleFindPeer` includes the target "if present in peerstore, even if it is self, the requester, or not a DHT server".

## Change
- In the vendored `libp2p-kad` patch we already carry: `Behaviour::set_peerstore_addresses`, and the target is prepended to a FIND_NODE reply when the owner has vouched for it.
- kwaai-p2p feeds it: each connected peer's identify addresses, table and connection addresses, and a circuit through this node for a peer holding a relay reservation here; refreshed on identify, reservation and disconnect. Nested circuits are dropped.
- Dialing side: a walk's answer for the target is kept and used (kad reports the addresses but only tables peers it connected to).

Nothing new on the wire; no client change. A 0.6.7/0.6.8 peer's ordinary walk finds a NATed peer as soon as the bootstraps run this.

## Evidence
- Integration tests (`find_node_peerstore.rs`): a connected-but-evicted peer is found by a plain walk and dialed; a NATed peer's answer carries a circuit through the bootstrap it reserved on. Both fail with the kad answer disabled.
- Isolated 3-node LAN (Mac, mini behind a relay-only firewall rule, Jetson Orin as bootstrap): `p2p peers find <mini>` from the Mac returned `/ip4/…/tcp/8000/p2p/<orin>/p2p-circuit/p2p/<mini>`; before the patch the same walk timed out.

## Deploy
Bootstraps first (with #197). Then peers in 0.6.9.


RezaRassool (commented):
## Review — rework, nine findings

I verified the vendored patch applies cleanly to pristine `libp2p-kad 0.48.0`, read the patched `behaviour.rs` in context, and traced the requester side (`discovered` → `QueryPeers::into_peerinfos_iter` → `ClosestPeersIter::into_result`) plus `libp2p-relay 0.21.1`'s `Event` enum. **The mechanism is sound** — kad's `handle_pending_outbound_connection` does chain in-flight query addresses, so the peerstore circuit really is used for the dial. CI is green, and the fetch script stamps the patch's sha256, so a changed patch correctly re-fetches.

The problems are all in what gets fed *into* the peerstore and how it's invalidated. The four that shape the rework:

1. **`core/crates/kwaai-p2p/src/service.rs:1969` — `relay::Event::ReservationClosed` is unhandled, so `relay_reservations` goes stale while a connection survives.** The new match handles `ReservationReqAccepted` and `ReservationTimedOut` only; `ReservationClosed` falls into `other => trace!`. In relay 0.21.1 reservations are tracked **per connection** (`behaviour.rs:277-294`) and `ReservationClosed` fires from `on_connection_closed`. The other clearing path, `SwarmEvent::ConnectionClosed` at `service.rs:1807`, is gated on `last`. So with two connections to a peer — which demonstrably happens; #206 in this same queue exists because of it — the reservation-bearing connection can close while the peer stays connected. Relay drops the reservation, kwaai keeps `relay_reservations` populated, and the bootstrap keeps answering FIND_NODE with `/…/p2p/<self>/p2p-circuit` for a peer it can no longer hop for. Every dialer's circuit dial is then denied — the exact `peer not found` this PR exists to remove. (`ReservationTimedOut` removing the peer wholesale even when it holds a reservation on another connection is the lesser half, since it fails closed.)

2. **`service.rs:2298` — the peerstore is fed regardless of `speaks_kad`.** Twenty lines above, the routing-table feed *is* gated on `speaks_kad`. The peerstore is a **more** public surface than the routing table — it's literally what FIND_NODE hands out — so a bootstrap now vouches for every foreign peer connected to it, and your own body cites 222 and 371 kubo connections. On a migration-window bootstrap built with `kad-multi-protocol` (serving `/ipfs/kad/1.0.0`), an Amino requester asking for one of those kubo peers gets a useful answer, re-inserting the bootstrap into the public DHT's serving path that the kad rename, `patches/README.md`, #197 and #174 are all working to exit — the documented p2pd-OOM vector.

3. **`service.rs:1110` — `dial_candidates` is deliberately unfiltered, and its output is now advertised to third parties.** It uses `|_| true` and is documented as keeping "loopback and LAN, which is how two nodes on one host reach each other" — correct for our own dials, wrong for a FIND_NODE answer. Two consequences: **(a)** a peer on the bootstrap's own LAN (the isolated 3-node bed in your body, or the two-node-on-one-Mac workflow) makes the bootstrap publish `/ip4/192.168.x.y/tcp/…` for that peer to the whole DHT, so remote dialers burn attempts probing their *own* subnet; **(b)** for an inbound connection `Connection.addr` is `send_back_addr` (`service.rs:1736`) — the peer's ephemeral source port, which the identify comment at `service.rs:2239` already calls "useless to anyone else". The new comment's "with port reuse the address a peer dialed in from is its listen port" holds only under `PortUse::Reuse` *and* a non-symmetric NAT; under symmetric NAT that mapping is valid for the bootstrap alone. Note the loopback tests depend on the unfiltered behaviour, so this wants a deliberate decision rather than a one-line swap. Relatedly the peerstore vector has **no per-peer cap at all**, where `add_routing_address` caps the same unverified identify list at 6.

4. **`core/crates/kwaai-p2p/tests/find_node_peerstore.rs:132` — the relay-reservation branch isn't actually exercised by the test that claims to cover it.** The relay's `external_addr` (and only listener) is `/ip4/127.0.0.1/tcp/PORT`. In `refresh_peerstore` the `ours` list filters on `is_announceable`, which rejects loopback, so `ours` is empty and **no** circuit is synthesised. The assertion is satisfied by the target's own identify-reported circuit listen address instead. Delete the whole `if self.relay_reservations.contains(&peer)` block and this test still passes — the code path that ships to bootstraps has no coverage. Giving the test relay a non-loopback `external_addr` (an RFC5737 documentation address, which `is_routable_v4` deliberately permits) and dropping the identify circuit would fix it.

### Also

5. **`service.rs:1163`** — `seed_walk_addrs` calls `kad.add_address` directly, bypassing `add_routing_address`. A walk answer *is* a remote claim, and kad's `Addresses` (`src/addresses.rs:29`) applies no bound, so one peer answering a walk with N addresses grows the target's entry to N, uncapped and unfiltered. Second-order: `forget_exhausted_entry` evicts an entry precisely so "a lookup can replace it", and `seed_walk_addrs` then re-seeds the just-failed addresses from the walk that eviction enabled, so `flush_routed` → `dial_routed` retries the dead ones. Filed with #206 and #185 as #208.

6. **`service.rs:2302`** — `is_announceable` instead of `is_announceable_with(self.require_global_ips)`, inconsistent with the line above it. `addresses.rs:127` says of the latter: "The one entry point callers should use, so the `require_global_ips` decision lives in one place." A bootstrap running `require_global_ips: true` will publish a peer's RFC2544/RFC5737 addresses through the peerstore while refusing to table them.

7. **`service.rs:1133`** — the nested-circuit filter doesn't reject circuits missing the relay peer id. `retain` drops ≥2 `P2pCircuit` segments, but `/ip4/…/tcp/…/p2p-circuit` (no `/p2p/<relay>`) passes both this and `is_announceable`. `addresses.rs:201` documents that rust-libp2p "refuses to dial outright — `Missing relay peer id.`" for exactly that shape, and it's been observed live on metro-win. A go-libp2p peer reporting the shortened form would have the bootstrap serve it as that peer's FIND_NODE answer; with finding 5 it then lands in every requester's routing table uncapped and is retried forever.

8. **`core/patches/libp2p-kad.patch:41`** — the inserted `set_peerstore_addresses` sits *between* `find_closest_local_peers`'s doc comment and the function, so the rustdoc is now attached to the wrong function. Compiles, but it's a trap for the next person rebasing onto a new kad. Move it after.

9. **`core/patches/README.md:59`** — "The entire delta is `libp2p-kad.patch` (two added methods, two files)" is untrue after this PR: three methods, plus a new `Behaviour` field and a behavioural change in the `FindNodeReq` handler. The section heading is also still "## libp2p-kad (multi-protocol names)" while the file now carries two unrelated changes.

### Checked, not a finding

Whether the walk answer can be dropped for a target the dialer couldn't itself query: `ClosestPeersIter::into_result` yields only `PeerState::Succeeded` peers, so `seed_walk_addrs` sees nothing there. A real limit, but not a regression — the old `found` computation had the same blind spot, and the addresses would be undialable anyway.

dazwin:
Rebased onto main (clean) and pushed `a58b87d4`. All nine addressed.

1. **ReservationClosed — fixed.** Reservations are now counted per connection (mirrors relay 0.21.1); `ReservationClosed` and `ReservationTimedOut` both release one, and the peer keeps its circuit while another connection holds one. Unit tests: two connections/one closes, renewal, unknown release.
2. **Ungated feed — fixed.** Peerstore feed gated on `speaks_kad`. Test: a `/foreign/kad/1.0.0` target reserved on a third relay gets an empty walk answer; fails with the gate removed.
3. **Unfiltered `dial_candidates` — decision: leave our own dials unfiltered, filter what we publish.** `refresh_peerstore` no longer reads `dial_candidates` or `send_back_addr`; the feed is identify listen addrs through `is_announceable_with(require_global_ips)`, a dialable-shape check, minus circuits through us, capped at `MAX_ADDRESSES_PER_PEER`. Unit tests cover LAN, loopback and the cap.
4. **Reservation branch untested — fixed, different shape.** An RFC5737 relay address cannot coexist with a walk that has to query the target (the walk only reports peers it could reach — your own "checked, not a finding"). Instead the identify-supplied circuit through us is dropped from the feed in code (needed for 1 anyway, or identify's stale circuit outlives the reservation) and the reservation branch synthesises from the swarm's external addresses, which is what libp2p-relay hands out. The test asserts the exact `<external>/p2p/<bootstrap>/p2p-circuit` string and fails with the reservation block disabled; a companion test with the relay on a third node exercises the identify branch and fails with the feed removed.
5. **`seed_walk_addrs` — fixed.** Goes through strip-dest-p2p → `is_announceable_with` → dialable shape → minus the entries `forget_exhausted_entry` just evicted → `add_routing_address`. Unit test with a fully-qualified evicted entry.
6. Fixed: `is_announceable_with(self.require_global_ips)` throughout the feed; unit test flips strict on a 198.18 address.
7. Fixed: new `addresses::circuit_relay` / `is_dialable_shape` — a circuit must carry exactly one `/p2p-circuit` with `/p2p/<relay>` immediately before it; the shortened go-libp2p form is rejected (four unit tests).
8. Fixed: method moved after `find_closest_local_peers`; patch regenerated with `diff -u` against pristine 0.48.0, fetch script re-stamped.
9. Fixed: README heading and delta wording; peerstore section moved under the kad section with the admission rules stated.

On #208: on this branch `add_routing_address` only caps; the announceable filter sits at each call site (identify, and now the walk). Folding it into `add_routing_address` is the repo-wide change that belongs to the #208 PR.

Checks: fmt clean, clippy clean, `cargo test -p kwaai-p2p` 218 passed (find_node_peerstore 3/3).

RezaRassool:
## Rebase needed — one file

**Note the interaction with #208.** That issue covered three PRs bypassing `add_routing_address` — this one, #206, and #185. **#185 has now landed**, so the rule is no longer hypothetical: `LearnedAddrs` and the reworked `candidate_addresses` are on `main`, and this PR's `seed_walk_addrs` and peerstore feed rebase onto them rather than alongside them.

That should make the review findings easier rather than harder — routing the walk-answer seeding through `add_routing_address`, and gating the peerstore feed on `speaks_kad` with `is_announceable_with`, are now changes against code that already works that way, instead of arguments about what the invariant should be.

---

dazwin:
Rebased onto main `8c5660b7` and pushed `23d33106`. The `service.rs` conflict was the `use` list only; the other two were `core/patches/README.md` placement against #204's cudarc section (kad subsection kept under `## libp2p-kad`, cudarc after it).

`seed_walk_addrs` still calls the one `add_routing_address` (service.rs:1241 → 1655), and the "minus what eviction removed" set is read from `routed_attempts[target].evicted`, which `forget_exhausted_entry` fills after `kad.remove_peer`. One correction to my own framing: #185 did not change `add_routing_address` at all; its `learned_addrs.forget` sits in the dial-failure handler. `a_walk_does_not_reseed_what_eviction_just_removed` passes.

Not done here, flagged as a follow-up: `learned_addrs.rs`'s own module doc says the FIND_NODE change "will answer from this same map". Feeding the peerstore from learned records is a policy change (vouching for a disconnected peer's signed claim versus a peer connected now), so it belongs in its own PR after this one.

Patch fetch verified from a pristine 0.48.0 crate: applies clean, stamp matches, no `.rej`; cudarc coexists. Red/green: with `refresh_peerstore` handing kad an empty list, 2 of 3 `find_node_peerstore` tests fail; restored, 3/3. Checks: gate clean, fmt, clippy, `kwaai-p2p` 147 unit + 110 integration passed.

PR #200, opened 2026-09-08
fix(relay): keep a configured relay's slot through its backoff

dazwin:
Part of #198 (item 5).

## Problem
Every hop-advertising peer was an equal relay candidate in a 16-entry oldest-out list. On production that list is kubo, so the bootstraps fell out of it once sixteen kubo peers had identified, and when a bootstrap reservation ended its slot went to the next candidate within 200 ms while the bootstrap sat in a 30 s backoff. Measured twice on 2026-09-07: one Wi-Fi blip moved both slots to foreign relays in 18 s, with the bootstraps pinned in `trusted_relays`. Kubo relays are fine introducers; the loss is that a peer's advertised addresses then point at relays that rotate, and the bootstraps hold no reservation to answer for it.

## Change
- Initial peers are configured relays: the tier tried first and never evicted from the candidate list (kubo's `Swarm.RelayClient.StaticRelays` is the same distinction).
- A free slot waits for a configured relay through its backoff, one 90 s hold per outage (two first-tier retries), before a discovered relay may take it. A relay that keeps failing still yields the slot.

## Evidence
- Unit tests: initial peers are configured candidates; a configured relay's slot waits out its backoff; a repeatedly failing configured relay yields.
- Isolated LAN: restarting the bootstrap ended the NATed mini's reservation; it waited 33 s and returned to the bootstrap. Another hop-advertising peer never took the slot.

## Deploy
Peers, 0.6.9.

### DIFF
--- a/core/crates/kwaai-p2p/src/relay_manager.rs
+++ b/core/crates/kwaai-p2p/src/relay_manager.rs
@@ -55,12 +55,15 @@ const BACKOFF_RESET_AFTER: Duration = Duration::from_secs(60);
 /// The swarm reports success or failure in practice, but a permanent slot leak
 /// in a process expected to run for months needs more than "in practice".
 const PENDING_DIAL_TIMEOUT: Duration = Duration::from_secs(60);
-/// How long a free slot waits for a configured relay in backoff before a
-/// discovered relay may take it. Two first-tier backoffs, jitter included: a
-/// bootstrap that dropped a reservation gets two tries before whichever peer
-/// last advertised hop inherits its slot. Without this the slot went to the
-/// next candidate within 200 ms of every "reservation ended", and a node
-/// drifted onto foreign relays in seconds (measured 2026-09-07).
+/// How long a configured relay that dropped out keeps a claim on its slot
+/// before a discovered relay may take it. Two first-tier backoffs, jitter
+/// included: a bootstrap that dropped a reservation gets two tries before
+/// whichever peer last advertised hop inherits its slot. Without this the slot
+/// went to the next candidate within 200 ms of every "reservation ended", and
+/// a node drifted onto foreign relays in seconds (measured 2026-09-07).
+///
+/// The same span also bounds an outage: a failure more than this long after
+/// the relay's previous one is a new outage, and starts a new hold.
 const CONFIGURED_HOLD: Duration = Duration::from_secs(90);
 /// Cap on identify-discovered candidates. The list is a convenience, not a
 /// routing table, and an unbounded one is a slow memory leak on a busy node.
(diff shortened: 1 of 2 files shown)


RezaRassool (commented):
## Review — rework, four findings

The direction is right — a configured tier that is never evicted is the correct shape — and `with_initial_peers`/`parse_configured` dedup correctly. But "tried first" currently holds only for the very first fill.

Method note: I built the branch in a temporary git worktree (symlinking the gitignored `core/patches/*` crates and reusing the main `target/`) and wrote throwaway probe tests against the real `RelayManager`. All four findings below are reproduced against running code, not inferred.

### Findings

1. **`core/crates/kwaai-p2p/src/relay_manager.rs:553` — the hold doesn't actually hold the slot.** It only blocks discovered candidates while the configured relay's backoff is unexpired, and `pick_candidate` then scans from the rotation cursor — which sits in the discovered region after startup and after every `fail()`. So the slot still goes to a discovered relay roughly 30 s after a bootstrap's reservation ends.

   Reproduced twice: with 2 configured + 8 discovered and 2 slots, `on_tick(+45s)` dials two discovered peers; with 1 configured + 4 discovered and 1 slot, the bootstrap gets zero retries inside its own 90 s hold.

   Worth knowing why this looks green: `a_configured_relays_slot_waits_out_its_backoff` passes only because with exactly two candidates the cursor wraps to index 0. Nothing in `pick_candidate` prefers the configured prefix once the cursor has moved, and the hold expires with the backoff rather than lasting the documented 90 s. A third candidate in that test would fail it.

2. **`relay_manager.rs:443`** — `holds.entry(..).or_insert(now)` never refreshes, and `holds` is cleared only by `on_new_listen_addr`, so a configured relay whose outage outlasts 90 s loses its claim permanently. Every later outage then hands its slot straight to a discovered relay. Reproduced.

3. **`relay_manager.rs:566`** — `held_for_configured` is one boolean applied to every slot, so a single configured relay in a hold blocks *all* free slots. With one configured relay and the default `max_relay_reservations: 2`, the node holds zero circuits for the hold window where it previously held one — a regression in the common single-bootstrap case. Reproduced.

4. **`core/crates/kwaai-p2p/src/service.rs:442`** — uses `config.initial_peers` while the adjacent line uses `config.effective_initial_peers()`. So a config naming bootstraps only in `bootstrap_peers` (`with_kwaai_bootstrap()`, `with_petals_bootstrap()`, the builder) dials them but gets an empty configured relay tier, and the whole policy is inert for those callers.

dazwin:
Rebased onto main (clean) and pushed `2291bcca`. All four fixed, and your reproductions are now tests (all five fail on the pre-fix code).

1. `pick_candidate` scans the configured list first, in order, regardless of the cursor; the cursor rotates over discovered relays only. Tests: 2 configured + 8 discovered, 2 slots, tick at +45 s dials both configured; 1 configured + 4 discovered, 1 slot, the bootstrap is retried inside its hold; `a_configured_relays_slot_waits_out_its_backoff` now has a third candidate.
2. Holds carry `started` and `last_failure`. A failure within `CONFIGURED_HOLD` of the previous one continues the outage; a failure after a longer quiet spell starts a fresh hold. A relay that keeps failing still yields after ~90 s; a later outage gets its own hold (test).
3. The hold is per relay: the discovered tier may take a slot only while `free_slots > held_configured_relays`. Test: 1 configured, 2 slots — a discovered relay fills the second slot during the hold, a third candidate is refused, the bootstrap gets its slot back at +45 s.
4. `with_initial_peers(&config.effective_initial_peers())`.

Found on the way, pre-existing on main and not touched here: `on_tick` retains only unexpired backoffs, so `consecutive_failures` resets at every 15 s tick and the exponential backoff never escalates in practice. Worth its own issue. Also no preemption: a recovered configured relay waits for a discovered relay's slot to free rather than reclaiming it; say if you expect reclaim.

Checks: fmt clean, clippy clean, `cargo test -p kwaai-p2p` 111 lib + all integration suites green.

PR #201, opened 2026-09-08
fix(announce): a whole-model node announces no block range

### LINKED ISSUE #183: On macOS a whole-model node announces state=2 and shard_loading=true together, permanently
Split out of the review of #182, which shipped without this. Line references are against `main` @ `18b027b7`.

## Summary

On macOS, a node that spawned a shard child announces `state=2` (ONLINE) **and**
`shard_loading=true` together, permanently. The map then shows a forever-"loading" shard on a
node that has nothing to load and is serving whole-model inference perfectly well — the exact
ambiguity `shard_loading` was added to remove, on the platform #182 exists to fix.

## Mechanism

`announce_shard_loading()` (`config.rs:1073`) reduces to `!shard_is_ready() && shard_child_alive`
and never consults `whole_model_is_ready()`. Each link verified:

| step | result |
|------|--------|
| `kwaainet start --shard` / `contribute.shards: true` spawns the child, parent writes its pid (`main.rs:331`, `:454`, `node.rs:1490`, `:2542`) | pid file present |
| child takes the whole-model branch (`shard_cmd.rs:339` → `serve_whole_model_via_ollama`) | writes **no** ready sentinel, blocks on `ctrl_c().await` (`shard_cmd.rs:307`) |
| `shard_is_ready()` = ready-file ∧ `is_running()` | **false** |
| `is_running()` reads pid, checks liveness | **true** — forever |
| `announce_shard_loading()` | **true** — forever |
| `refresh_whole_model_ready` writes `whole_model.ready` → `announce_state()` | **2** |

So every announce carries `state=2` with `shard_loading=true`.

`to_msgpack` (`announce.rs:255`) also emits the key regardless of `state`, which contradicts its
own doc comment ("Only meaningful while `state` is JOINING").

## Impact

Cosmetic, not a routing fault — every consumer filters on `state == 2` and ignores
`shard_loading`, so nothing mis-routes. The cost is that the one signal distinguishing
"mid-load" from "up, serving nothing" is now permanently wrong on Apple-silicon nodes, which
makes it useless for the diagnosis it was introduced for.

## Suggested fix

Either or both:

- Gate the encode site on `state == 1`, which makes the field match its documented meaning.
- Add `&& !ShardManager::whole_model_is_ready()` to `announce_shard_loading()`.

The first is the more durable of the two: it makes the invariant structural rather than
dependent on remembering every readiness source.

dazwin:
Part of #198 (item 6). Resolves #183.

## Problem
A macOS node serving the whole model through Ollama announced its configured block range anyway, ONLINE, so every chain build discovered it as a 0–32 block server, pinned it first, and had the hop refused: `remote does not support protocol /kwaai/inference/1.0.0` (christophermayfield-macos, v0.6.8, on every production chain build on 2026-09-07).

## Change
The announced range depends on what the node serves: a block shard announces its range; a whole-model node announces an empty one, so no per-block record is written and it is never chosen as a hop. Its `_kwaai.inference.nodes` entry is unchanged, which is how `p2p://auto` finds it. A node serving neither keeps the configured range and its JOINING state.

## Evidence
- Unit tests for the three cases.
- Isolated LAN: a Mac on the Ollama route logged `Announced 0 blocks` while its model and inference-node records were still stored.

## Deploy
Peers, 0.6.9.

### DIFF
--- a/core/crates/kwaai-cli/src/announce.rs
+++ b/core/crates/kwaai-cli/src/announce.rs
@@ -759,6 +759,61 @@ pub fn build_unannounce_records(
     ctx: &AnnounceContext<'_>,
     server_info: &DHTServerInfo,
 ) -> Result<Vec<StoreRequest>> {
+    let subkey = rmp_serde::to_vec(&ctx.peer_id.to_base58())?;
+    let node_info = NodeInfo::from_peer_id(ctx.peer_id);
+    let expiration = get_dht_time() + ANNOUNCE_TTL_SECS;
+
+    let mut records = Vec::with_capacity(2);
+    records.extend(tombstone_blocks(
+        ctx,
+        server_info,
+        server_info.start_block..server_info.end_block,
+    )?);
+
+    // The VPK record is withdrawn with the *live* VPK value, not the offline
+    // ServerInfo — it has no state field to flip, so the only signal available
+    // is that it stops being refreshed.
+    if let Some(ref vpk) = server_info.vpk_info {
+        records.push(StoreRequest {
+            auth: Some(RequestAuthInfo::new()),
+            keys: vec![dht_id(VPK_NODES_KEY)],
+            subkeys: vec![subkey],
+            values: vec![vpk.to_msgpack_bytes()?],
+            expiration_time: vec![expiration],
+            in_cache: vec![false],
+            peer: Some(node_info),
+        });
+    }
+
+    Ok(records)
+}
+
+/// Tombstone the block records a previous announce wrote under
+/// `previous.0..previous.1` that the current range no longer refreshes, so
+/// a node whose range shrank — or collapsed to nothing — stops reading ONLINE
+/// for those blocks until the 360 s TTL runs out.
+///
+/// Blocks still in the current range are left alone: the fresh ONLINE record
+/// carries the same expiration and would be rejected as not strictly newer.
+pub fn build_dropped_block_records(
+    ctx: &AnnounceContext<'_>,
+    server_info: &DHTServerInfo,
+    previous: (i32, i32),
+) -> Result<Vec<StoreRequest>> {
+    let current = server_info.start_block..server_info.end_block;
+    let dropped = (previous.0..previous.1).filter(|b| !current.contains(b));
+    Ok(tombstone_blocks(ctx, server_info, dropped)?
+        .into_iter()
+        .collect())
+}
+
+/// One STORE request writing the `state = -1` tombstone under each of `blocks`;
+/// `None` when there is nothing to write, so no all-empty request goes out.
+fn tombstone_blocks(
+    ctx: &AnnounceContext<'_>,
+    server_info: &DHTServerInfo,
+    blocks: impl Iterator<Item = i32>,
+) -> Result<Option<StoreRequest>> {
     let offline_info = DHTServerInfo {
         state: -1, // OFFLINE — tells map.kwaai.ai to remove the node immediately
         throughput: 0.0,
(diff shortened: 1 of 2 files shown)


RezaRassool (commented):
## Review — five findings, but the top one depends on merge order

The core change is sound: the guard in `build_announce_records`, the new `announced_range` helper, and its three unit tests all check out, and I ran them (3 passed). I also confirmed downstream consumers of an empty range are safe — `discover_chain` only queries block keys, `discover_inference_nodes` ignores start/end, and the coverage/rebalancer slices are clamped so `s == e` yields an empty slice rather than a panic.

**Sequencing first, because it changes what's worth doing here:** finding 1 below is the most severe, and #191 (drop the Go p2pd daemon) deletes the code it's about. If #191 merges first — and it's currently second in the merge order, right after #192 — finding 1 evaporates and this PR is left with four smaller items. Probably not worth fixing the p2pd path if it's about to be removed; worth a quick check of the merge order before spending time on it.

### Findings

1. **`core/crates/kwaai-cli/src/node.rs:504` — the p2pd announce path is untouched.** This PR edits only `announce.rs` and `node_native.rs`, while `node.rs` still announces `config.start_block()` … `config.effective_end_block()` unconditionally. So a node with `native_p2p: false` — a documented rollback — still publishes ONLINE block records for 0–32 and reproduces #183. `announced_range` is private to `node_native.rs`, so the two paths now disagree about the range. Moot once #191 lands.

2. **`core/crates/kwaai-cli/src/node_native.rs:809`** — `announced_range` gates on the readiness sentinels rather than on the state actually announced. With `announce_online_without_shard: true` the node still announces state=2 *plus* the full range, and the new "node with neither" test's stated rationale ("announces JOINING for it") doesn't hold in that config.

3. **`core/crates/kwaai-cli/src/announce.rs:476`** — when the range collapses to empty, the previously published block records are neither refreshed nor tombstoned, leaving up to 360 s of stale ONLINE records. `build_unannounce_records` also never got the `keys.is_empty()` guard and sends an all-empty StoreRequest.

4. **`node_native.rs:681`** — the Ollama-recovery arm calls `refresh_server_info` without `refresh_whole_model_ready` first, so it re-announces from a stale sentinel: state=1 plus the full 0–32 block range, which is exactly the set of records this PR removes.

5. **`node_native.rs:509`** — the `[4/4]` banner still prints `Blocks : 0–32` immediately after "Announced 0 blocks". Also worth noting there's no regression test for the new guard itself: every record-shape test uses a 3-block fixture, so reverting the guard keeps the suite green.

Minor: `refresh_server_info`'s doc comment was left attached to the newly inserted `announced_range`, which now opens with a sentence describing a different function.

dazwin:
Rebased onto main and pushed `b110f8ef`.

1. **Moot.** After #191 `node.rs` has no announce site; only `node_native.rs` announces.
2. Fixed: `announced_range` now takes the state `announce_state()` actually reports. ONLINE without a ready shard (whole model, or `announce_online_without_shard`) announces an empty range; JOINING keeps the configured range as intent. Tests rewritten with the corrected rationale. Note this changes what the kwaaiai-env NAT nodes show on the map (Idle rather than Serving) — the honest state.
3. Fixed: the node remembers the last range that reached the DHT and prepends tombstones for blocks the previous round published and this one no longer covers (only the dropped ones; a still-covered block gets its fresh ONLINE record, which hivemind would reject behind a same-expiration tombstone). `build_unannounce_records` shares a `tombstone_blocks` helper that returns `None` for an empty key set. Test covers collapse, move, unchanged and grown.
4. Fixed one level up: `refresh_server_info` calls `refresh_whole_model_ready` itself, so all four arms (SIGHUP, tick, reachability settle, Ollama recovery) re-derive the sentinel.
5. Fixed: banner prints the announced range or "none"; regression test `an_empty_range_announces_no_block_records` uses a `start == end` fixture and asserts no zero-key request.
Minor: doc comment back on `refresh_server_info`.

Checks: fmt clean, clippy clean, `cargo test -p kwaainet --bin kwaainet` 252 passed.

RezaRassool:
## Rebase needed — one file

Conflicts in `core/crates/kwaai-cli/src/node_native.rs` only. Smaller than it looks.

**Some of this PR may now be redundant.** Its most severe review finding was that the p2pd announce path in `node.rs` still published an unconditional 0–32 block range. #191 deleted that path — `node.rs` went from 2,657 lines to 679 and the block-range announce is gone entirely. So that finding is moot, and the remaining work is the four smaller items: the `announce_online_without_shard` gate, tombstoning stale records when the range collapses, the Ollama-recovery arm re-announcing from a stale sentinel, and the `[4/4]` banner still printing `Blocks : 0–32`.

Worth re-reading the diff against the new `node_native.rs` before resolving — the conflict may be smaller than the original change.

---

dazwin:
Rebased onto main `8c5660b7` and pushed `5f023bde`. Both commits kept. The only conflict was `NativeNode::announce`: #185 made `server_info` `&mut` and inserted the `signed_addrs` refresh; this branch made `self` `&mut`, added `announced_blocks` and the tombstone prepend. Kept both, main's refresh first. Diffing the rebased file against the pre-rebase tip shows only #185's hunks.

Nothing in this PR became redundant: main did not touch `refresh_server_info`, `announced_range`, the banner or the tick arm. All four refresh arms still go `refresh_server_info` → `refresh_whole_model_ready` → `announce_state()` → `announced_range` (node_native.rs:588/632/699/712 → 855).

Red/green, one revert each: `an_empty_range_announces_no_block_records`, `online_without_a_shard_announces_no_blocks` and `a_shrunken_range_tombstones_only_the_dropped_blocks` each fail with their guard removed and pass restored. The `announce()` wiring itself (tombstones from `announced_blocks`) has no direct test; it composes two tested helpers. Say if you want an integration test for it before merge.

Checks: gate clean, fmt, clippy, `kwaainet --bin` 289 passed.

PR #202, opened 2026-09-08
fix(cli): apply log_level from config.yaml to the daemon's log filter

dazwin:
Part of #198 (item 9).

`log_level` in config.yaml was read only to print it back; the daemon's filter was a fixed `info` unless `RUST_LOG` was set, so `log_level: debug` changed nothing. `RUST_LOG` still wins; the config level is the default under it, read from the file without creating it and before logging exists.

Evidence: unit test on the filter; isolated LAN node at `log_level: debug` writes DEBUG lines with no `RUST_LOG`.

### DIFF
--- a/core/crates/kwaai-cli/src/config.rs
+++ b/core/crates/kwaai-cli/src/config.rs
@@ -1031,7 +1031,6 @@ impl KwaaiNetConfig {
         Self::load_or_create().unwrap_or_else(|_| self.clone())
     }
 
-    /// Load config from `~/.kwaainet/config.yaml`, creating it with defaults if absent.
     /// `log_level` from the config file, without creating the file or
     /// parsing the rest of it. Read before logging is initialised, so it
     /// must not fail, create or log anything.
@@ -1044,6 +1043,7 @@ impl KwaaiNetConfig {
         serde_yaml::from_str::<LogLevelOnly>(&text).ok()?.log_level
     }
 
+    /// Load config from `~/.kwaainet/config.yaml`, creating it with defaults if absent.
     pub fn load_or_create() -> Result<Self> {
         let cfg_file = config_file();
         std::fs::create_dir_all(cfg_file.parent().unwrap())?;
@@ -1211,7 +1211,17 @@ impl KwaaiNetConfig {
             "blocks" => self.blocks = value.parse().context("blocks must be a number")?,
             "port" => self.port = value.parse().context("port must be a number")?,
             "use_gpu" => self.use_gpu = parse_bool(value)?,
-            "log_level" => self.log_level = value.to_string(),
+            "log_level" => {
+                self.log_level = parse_log_level(value)
+                    .ok_or_else(|| {
+                        anyhow::anyhow!(
+                            "log_level must be one of {}, got '{}'",
+                            LOG_LEVELS.join(", "),
+                            value
+                        )
+                    })?
+                    .to_string()
+            }
             "public_name" => self.public_name = Some(value.to_string()),
             "public_ip" => self.public_ip = Some(value.to_string()),
             "public_port" => {
@@ -1291,6 +1301,18 @@ impl KwaaiNetConfig {
         }
         Ok(())
     }
+}
+
+/// Accepted `log_level` values, least to most verbose.
+pub const LOG_LEVELS: [&str; 6] = ["off", "error", "warn", "info", "debug", "trace"];
+
+/// The canonical name for a `log_level` value, or None if it is not one.
+pub fn parse_log_level(s: &str) -> Option<&'static str> {
+    let s = s.trim();
+    LOG_LEVELS
+        .iter()
+        .copied()
+        .find(|l| l.eq_ignore_ascii_case(s))
 }
 
 fn parse_bool(s: &str) -> Result<bool> {
@@ -1690,6 +1712,22 @@ mod start_block_provenance {
 }
 
 /// `cargo test` must not be able to touch the developer's real config.
+#[cfg(test)]
+mod log_level_values {
+    use super::*;
+
+    #[test]
+    fn set_key_rejects_a_level_the_filter_would_not_understand() {
+        let mut cfg = KwaaiNetConfig::default();
+        for typo in ["warning", "verbose", ""] {
+            assert!(cfg.set_key("log_level", typo).is_err(), "{typo:?}");
+        }
+        cfg.set_key("log_level", " Debug ").expect("set");
+        assert_eq!(cfg.log_level, "debug");
+        assert_eq!(parse_log_level("OFF"), Some("off"));
+    }
+}
+
 #[cfg(test)]
 mod test_isolation {
     use super::*;
--- a/core/crates/kwaai-cli/src/config.rs
+++ b/core/crates/kwaai-cli/src/config.rs
@@ -1032,6 +1032,18 @@ impl KwaaiNetConfig {
     }
 
     /// Load config from `~/.kwaainet/config.yaml`, creating it with defaults if absent.
+    /// `log_level` from the config file, without creating the file or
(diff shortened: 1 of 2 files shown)


RezaRassool (commented):
## Review — one blocking finding, three smaller

The change does what it claims: `RUST_LOG` still wins, config `log_level` becomes the default, and the daemon path is genuinely covered because `run-node` (spawned in `daemon.rs`) goes through the same `main()` init. CI is green.

All four findings were confirmed by running the exact filter strings through tracing-subscriber 0.3.23 — the version in `Cargo.lock` — rather than by reasoning about them.

### Findings

1. **`core/crates/kwaai-cli/src/main.rs:1936` — an unrecognised `log_level` silently disables *all* logging, ERROR included.** The bare invalid word parses as a *target* at TRACE, `kwaainet=<invalid>` is dropped lossily by `parse_lossy`, and because the resulting directive table is non-empty EnvFilter never installs its ERROR fallback.

   Reproduced with the real crate name and the locked subscriber version: `log_level: warning` and `log_level: verbose` — both plausible typos, and neither a valid level name — print nothing at all, where `info` and `debug` print normally.

   Two things make this worth fixing before merge rather than after: `config set log_level` validates nothing (`config.rs:1270` is a bare string assignment), and before this PR the key was only read back for display, so a typo was harmless. Now it silently blinds the daemon. Suggest validating against the five level names and falling back to `info` with a warning on the way through.

2. **`main.rs:1941`** — the unconditional `hnsw_rs=warn,kwaai_storage=warn,tantivy=warn` pins now act as floors as well as ceilings, so at `log_level: off` those crates still print WARN while kwaainet's own ERROR is suppressed. Pinning at `min(level, warn)` fixes it.

3. **`main.rs:1934`** — the bare global directive raises *every* dependency (libp2p, quinn, hyper, sqlx) to the configured level, which contradicts the doc comment's "for kwaainet itself". A persisted `log_level: trace` then floods the unrotated append-only `~/.kwaainet/logs/kwaainet.log` on every daemon start. Either the comment or the global directive wants changing.

4. **`config.rs:1083`** — `load_or_create`'s doc comment ("creating it with defaults if absent") got orphaned onto the newly inserted `peek_log_level`, whose own doc promises the opposite; `load_or_create` is left undocumented. This pattern shows up in #201, #199 and #203 too — new function inserted between an existing doc comment and the function it documented.

dazwin:
Rebased onto main (clean) and pushed `a7630a26`. All four fixed; the tests drive real `EnvFilter`s through the locked tracing-subscriber 0.3.23 with a capturing writer rather than asserting on filter strings.

1. `log_level` is validated against `off, error, warn, info, debug, trace` (case-insensitive). Unknown values fall back to `info` and `main()` emits a warning right after `.init()`; `config set log_level` rejects anything else and lists the accepted names. Test: `warning` and `verbose` still print kwaainet ERROR/INFO.
2. The noisy crates are pinned at `min(level, warn)`. Test: at `off`, ERROR from `kwaainet`, `kwaai_storage` and `hnsw_rs` are all dropped; at `error`, `hnsw_rs` ERROR passes and WARN does not.
3. Decision: the configured level applies to our own crates only. Filter is `info,hnsw_rs={pin},kwaai_storage={pin},tantivy={pin},kwaai={level}` — `EnvFilter` matches targets by prefix, so `kwaai=` covers `kwaainet` and every `kwaai_*` crate, and the longer `kwaai_storage=` directive still wins. Doc comment rewritten. Test: at `trace`, `kwaai_p2p` TRACE passes while `libp2p_swarm` and `quinn` stay at info.
4. `peek_log_level` moved; `load_or_create` has its doc comment back.

Checks: fmt clean, clippy clean, `cargo test -p kwaainet --bin kwaainet` 252 passed.

RezaRassool:
Pushed one doc-only commit (`7cb85040`) to this branch — say the word and I'll drop it if you'd rather write it yourself.

**What it fixes.** `mod log_level_values` was inserted between `mod test_isolation` and the `///` block describing it, so *"`cargo test` must not be able to touch the developer's real config"* ended up on a module about log-level parsing, and `test_isolation` was left with no doc at all. `///` attaches to the next item, so nothing complains: rustc, clippy and rustdoc all accept it.

The commit gives each module its own: `log_level_values` gets one saying why the test exists (an unparseable level silently disabled every log line *including ERROR*, because the bare word parsed as a target), and the isolation doc goes back where it came from. Four lines of prose, one file, no code change. If the wording on your test reads better your way, please just amend it.

**Why it surfaced now.** #214 added a CI check for exactly this pattern — it reached review five times in this queue, including in the commit that *fixed* the original finding here. This PR was one of two it flagged. Verified before pushing: gate passes, `cargo test -p kwaainet --bin kwaainet config::` is 50 passed / 0 failed, including both affected modules.

**Everything else on this PR is confirmed fixed** — I re-read the code rather than the commit messages. `parse_log_level` validating and falling back to `info` *with* a warning is a better answer than I suggested; `min(level, warn)` and the `info,…,kwaai={level}` scoping both check out.

One thing when you next pull: `main` is now `8c5660b7` and #204 added a third patched crate, so re-run `core/patches/fetch-patches.sh` or `cargo` won't load the workspace at all.

dazwin:
Thanks for `7cb85040`; kept it. Your version gives `log_level_values` its own doc, which is better than the move I had queued locally, so I dropped mine. Rebased your head onto main `8c5660b7` (clean) and pushed `56e1f00a`.

Verified at the rebased head: orphaned-docs gate clean, fmt clean, `config::` 50 passed, `log_filter_tests` 4 passed. Red/green on `unknown_levels_fall_back_to_info_with_a_warning`: with `default_log_filter` reverted to the pre-validation shape it fails at the `warning.unwrap()`; restored, it passes.

RezaRassool:
Read the whole diff at `56e1f00a` before merging. All four original findings are fixed, and verified by reading the code rather than the commit messages:

- **`set_key` validates.** `parse_log_level` plus an error naming the valid levels, so a typo cannot reach the filter at all — this is the stronger half of the fix and I had only asked for the filter side.
- **The fallback warns, and the warning is visible.** `default_log_filter` returns `(String, Option<String>)` and `main` emits it *after* `.init()`, with `kwaai=info` in force, so it actually reaches the log.
- **`min(level, warn)` is correct.** Traced all six levels: `off`/`error` sit below the pin, `info`/`debug`/`trace` clamp to `warn`, and the `unwrap()` in `rank` cannot panic because `level` is always a `LOG_LEVELS` member.
- **Doc ownership restored** on `load_or_create` / `peek_log_level`.

The tests are the strongest in this queue — `hits()` builds a real subscriber, emits through it and checks the buffer, so none of these assertions can pass vacuously. Given how many weak tests the September review turned up, that is worth saying explicitly.

## One note, not blocking, logged here because it will outlive the PR

**`off_is_really_off` overstates itself.** At `log_level: off` the filter is:

```
info,hnsw_rs=off,kwaai_storage=off,tantivy=off,kwaai=off
```

The bare `info` global still admits **libp2p, quinn and hyper at INFO**. The test asserts only that `kwaainet::node`, `kwaai_storage::index` and `hnsw_rs::hnsw` are silent — it makes no dependency assertion at all. Your own `the_level_applies_to_our_crates_and_not_to_dependencies` proves the mechanism: it asserts `hits(filter, "libp2p_swarm", INFO)` is true, and the global is the same hardcoded `info` at every level.

The behaviour is deliberate and the function doc says so ("dependencies at info"), so this is not a defect. But an operator who sets `off` to quiet a node will still get dependency INFO into the unrotated, append-only `~/.kwaainet/logs/kwaainet.log` — which was part of what the original finding about the global directive was worried about. Either the test name wants narrowing to match what it checks, or `off` should also silence the global. Your call; I have recorded it in the merge commit either way.

PR #203, opened 2026-09-08
fix(shard): rank pinned-path candidates by reachability before span

dazwin:
Part of #198 (item 4). Relates to #179. **Stacked on #185** — base is `feat/announce-dial-addrs`; retarget to `main` when #185 merges.

## Problem
The pre-connect dialled every server in the chain and then ignored the outcome: `build_pinned_path` ranked hops by block span alone, on the first pin and every rebuild. On production, with 7 of 10 peers undialable, a session pinned a peer it had just failed to reach, failed the hop, rebuilt onto a peer with no address at all, and failed again — while three connected servers sat idle.

## Change
After the pre-connect the daemon is asked which chain peers it holds a connection to. Pinning skips failed peers outright and ranks connected peers above unreached ones before span, at every site. An unreached peer stays a candidate of last resort.

## Evidence
- Unit test covering the ranking and the rebuild case.
- Production (2026-09-07): with the same chain, the session pinned a connected peer and made a real hop instead of two rebuilds onto undialable ones.

## Deploy
Peers, 0.6.9.

### DIFF
--- a/core/crates/kwaai-cli/src/shard_api.rs
+++ b/core/crates/kwaai-cli/src/shard_api.rs
@@ -242,12 +242,14 @@ async fn run_inference(
     let mut client_guard = state.client.lock().await;
     let mut failed_peers: std::collections::HashSet<libp2p::PeerId> =
         std::collections::HashSet::new();
+    let mut unreached = crate::shard_cmd::unreached_peers(&mut client_guard, &state.chain).await;
 
     // Pin peer path for this request so KV-caches stay coherent.
-    let mut pinned_path = match crate::shard_cmd::build_pinned_path(
+    let mut pinned_path = match crate::shard_cmd::build_pinned_path_ranked(
         &state.chain,
         state.total_blocks,
         &failed_peers,
+        &unreached,
     ) {
         Ok(p) => p,
         Err(e) => {
@@ -275,6 +277,7 @@ async fn run_inference(
             request,
             Some(&state.our_peer_id),
             &mut failed_peers,
+            &unreached,
             None,
             None,
         )
@@ -283,10 +286,13 @@ async fn run_inference(
             Ok(r) => r,
             Err(e) => {
                 // Try rebuilding path excluding failed peer
-                match crate::shard_cmd::build_pinned_path(
+                unreached =
+                    crate::shard_cmd::unreached_peers(&mut client_guard, &state.chain).await;
+                match crate::shard_cmd::build_pinned_path_ranked(
                     &state.chain,
                     state.total_blocks,
                     &failed_peers,
+                    &unreached,
                 ) {
                     Ok(new_path) => {
                         pinned_path = new_path;
@@ -308,6 +314,7 @@ async fn run_inference(
                             retry,
                             Some(&state.our_peer_id),
                             &mut failed_peers,
+                            &unreached,
                             None,
                             None,
                         )
@@ -674,12 +681,15 @@ fn detect_gguf_path(
         }
     }
 
-    // For auto-detection, require all blocks to be hosted by our peer
+    // For auto-detection, require our own entries alone to cover every block.
+    let ours: Vec<BlockServerEntry> = chain
+        .iter()
+        .filter(|e| e.peer_id == *our_peer_id)
+        .cloned()
+        .collect();
+    let none = std::collections::HashSet::new();
     let covers_all =
-        crate::shard_cmd::build_pinned_path(chain, total_blocks, &std::collections::HashSet::new())
-            .ok()
-            .map(|path| path.iter().all(|e| e.peer_id == *our_peer_id))
-            .unwrap_or(false);
+        crate::shard_cmd::build_pinned_path_ranked(&ours, total_blocks, &none, &none).is_ok();
 
     if !covers_all {
         return None;
--- a/core/crates/kwaai-cli/src/shard_cmd.rs
+++ b/core/crates/kwaai-cli/src/shard_cmd.rs
@@ -1424,7 +1424,7 @@ pub async fn cmd_shard_run(args: ShardRunArgs) -> Result<()> {
             .connect_peer_with_addrs(&entry.peer_id, &entry.dial_addrs)
             .await;
     }
-    let unreached = unreached_peers(&mut client, &chain).await;
+    let mut unreached = unreached_peers(&mut client, &chain).await;
 
     // ── Inference loop ────────────────────────────────────────────────────────
     let mut generated_ids: Vec<u32> = Vec::new();
@@ -1485,6 +1485,7 @@ pub async fn cmd_shard_run(args: ShardRunArgs) -> Result<()> {
             request,
             Some(&our_peer_id),
             &mut failed_peers,
+            &unreached,
             show_stats.then_some(&mut token_hops),
             reputation.clone(),
         )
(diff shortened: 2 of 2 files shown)


RezaRassool (commented):
## Review — rework, five findings

Reviewed against the stacked base `feat/announce-dial-addrs` rather than `main`. The core ranking change (`max_by_key(|e| (!unreached.contains(..), e.end_block))`) is **correct and terminates safely** — I checked that preferring a narrower reached span can never introduce a coverage gap, because any wider candidate at position `p` still covers the new position. The added unit test is sound too.

The findings are about where the ranking doesn't reach, and one thing it changes downstream that's silent.

Heads-up on CI: this PR has no test run — the only check is the release workflow's `plan` job, so `Test`/`Clippy`/`Windows check`/`WASM`/`Format` never ran. Worth pushing an empty commit or retargeting to get a real run before merge.

### Findings

1. **`core/crates/kwaai-cli/src/shard_cmd.rs:3167` — the reachability ranking is thrown away at hop time.** `forward_through_chain` is handed `&pinned_path` as its `chain` and re-selects a candidate itself at every position, sorting `b.end_block.cmp(&a.end_block)` first — span alone, then trust, then self. So whenever the pinned path still contains an unreached wide-span entry (chain = connected A 0–15, unreached U 0–31 → ranked path `[A, U]`), the hop loop at `pos = 0` sorts U first because 32 > 16 and dials the undialable peer anyway, on every token. The fix only takes effect where ranking removes the unreached peer from the path entirely; where it's retained "as a candidate of last resort" it is in fact tried *first*.

2. **`shard_cmd.rs:2804` — the ranking breaks the path's non-overlap invariant by construction, and the consequence is silent.** Preferring a narrower reached entry guarantees the next pick's `start_block` precedes the previous entry's `end_block` (A 0–15 reached, U 0–31 unreached → path `[A(0–15), U(0–31)]`). `InferenceRequest` (`block_rpc.rs:70`) carries no block range — a server always runs its own configured range — so a hop that overlaps its predecessor re-applies layers already applied.

   Reachable: U fails at `pos = 0` with a *transient* error (so `forward_through_chain` deliberately doesn't blacklist it), A runs 0–15, then U is the only candidate covering 16 and succeeds, applying layers 0–31 to A's output. The result is silently corrupted logits rather than an error. It also makes the printed "Pinned path" table show overlapping ranges.

3. **`core/crates/kwaai-cli/src/shard_api.rs:247` — not applied "at every site".** `shard_api.rs` has the identical pre-connect loop (lines 874–879) and then pins with the unranked `crate::shard_cmd::build_pinned_path` at 247, rebuilds with it at 286, and again at 679. `kwaainet shard api` is the long-running OpenAI-compatible server — the most likely production consumer — and it retains the exact bug in the problem statement. It already holds `client_guard` and `state.chain`, so threading `unreached` through is mechanical.

   Also `shard_cmd.rs:3004`: `cmd_shard_circuit_create` pins unranked *and persists the result to disk with a TTL* for reuse "across invocations". A circuit created while most peers are undialable bakes them into a path that outlives the session — strictly worse than the transient case being fixed. `client` is in scope two dozen lines above.

4. **`shard_cmd.rs:2777` — our own peer is always classified unreached, so the local self-bypass loses to any remote.** A node running `shard serve` appears in its own discovered chain (see the check at 2534 and the self-dispatch at 3196), the pre-connect's dial-to-self always fails, and `list_peers` never reports self — so the self entry is in `unreached` unconditionally and now ranks below *any* connected remote. Concretely: we serve 0–31 locally with the local port file present, a connected remote serves 0–15, and the new ranking pins the remote for 0–15 — sending activations over the network when `local_inference_call` was available, which is the cheapest possible hop and one `forward_through_chain` explicitly tries to prefer. `unreached_peers` should exclude `our_peer_id`.

5. **`shard_cmd.rs:1493` — `unreached` is a one-shot snapshot reused for the life of the session.** Computed once after the pre-connect, then passed unchanged to every `build_pinned_path_ranked` call in the recovery path (same at 1933 in `run_streaming_inner`). Since `forward_through_chain` classifies `stream reset`/`early eof`/`connection closed` as transient and deliberately doesn't add the peer to `failed_peers`, a peer whose connection dies mid-session is neither blacklisted nor demoted — it's still ranked "connected" — so every rebuild re-pins it. That's the same rebuild-onto-a-dead-peer loop the PR is trying to break, with a different trigger. Re-reading `unreached_peers` on rebuild is one cheap local daemon query.

### Smaller

- **`shard_cmd.rs:2757`** — the greedy-walk doc block was left in place with the new `unreached_peers` doc appended, so the whole thing now documents `unreached_peers` and describes behaviour it doesn't have. `build_pinned_path` is left with a one-liner and `build_pinned_path_ranked` never states the greedy contract. Stale `build_pinned_path` reference in the module-map comment at 1690 too.
- **`shard_cmd.rs:2836`** — the diff removes the `// ── Circuits ──` section header and the `Circuit` struct's doc comment with nothing replacing them, leaving a bare `#[derive]` after the pinned-path helpers. Looks accidental.
- `unreached_peers` swallows a `list_peers` error via `unwrap_or_default()`, which marks every chain peer unreached and silently degrades to the old span-only ranking — benign, but indistinguishable from a genuinely empty connection table.

dazwin:
Rebased onto the new `feat/announce-dial-addrs` head (dceb9532) and pushed `81ad2eba`. That also gives it a real CI run.

1. **Ranking thrown away at hop time — fixed.** Hop selection is now `hop_candidates()`: only entries whose `start_block == pos` qualify, ranked reached-first, then span, trust, self. `forward_through_chain` takes `unreached` and follows a pinned path as the plan; a failed hop bails so the caller rebuilds instead of re-sorting by span. Test `hop_candidates_follow_the_plan_and_rank_fallbacks`.
2. **Overlap by construction — fixed.** `build_pinned_path_ranked` is a depth-first walk that extends only from the previous `end_block` and backs out of dead ends (memoised), so a path is contiguous by construction and the table cannot show an overlap. Your A(0–15 reached)/U(0–31 unreached) case pins `[U]`; with a reached B(16–31) it pins `[A, B]`; A(0–15)/B(8–31) is now an error naming block 16 rather than silently re-applying 8–15. That last overlap was reachable with the old span-only greedy too, so this is a semantics change worth knowing: a chain whose entries only overlap errors instead of producing wrong logits. Test `pinning_never_overlaps_ranges`.
3. **Not applied everywhere — fixed.** `shard api` reads `unreached_peers` under the client guard per request and again on rebuild, pins ranked at both sites, and passes it to both forwards; `circuit create` gained the pre-connect loop and a ranked pin before persisting. The llama-cpp `detect_gguf_path` site asks "do our own entries cover every block", so it filters the chain to our peer id rather than ranking by reachability. The span-only `build_pinned_path` became dead code and is removed.
4. **Self classed unreached** — fixed in the stacked #210 (`unreached_in` excludes our peer id, alongside the daemon loopback in #209), not duplicated here.
5. **One-shot snapshot — fixed.** `unreached` is re-read with one `list_peers` on every rebuild in `shard run`, `run_streaming_inner` and the `shard api` handler.

Smaller: docs on `unreached_peers`/`build_pinned_path_ranked` and the module map corrected, `// ── Circuits ──` header and `Circuit` doc restored, and a `list_peers` error now warns and yields no information rather than "everything unreached".

Checks: fmt clean, clippy clean, `cargo test -p kwaainet --bin kwaainet` 273 passed. The `llama-cpp` cfg block could not be compile-checked here (its cmake build fails in this environment); it is six lines on already-imported symbols.

PR #204, opened 2026-09-08
feat(jetson): CUDA shard serving on Orin — cudarc sync-allocation opt-out

dazwin:
Relates to #198 (goal 3: partial model on the Orin Nano).

## Problem
Serving blocks 0–8 of Llama-3.1-8B in f16 on a Jetson Orin Nano's GPU failed at block 3 with `CUDA_ERROR_OUT_OF_MEMORY` and 1.8 GB free. cudarc routes every allocation through `cuMemAllocAsync` whenever the device reports memory-pool support, and on L4T r36 / CUDA 12.6 that pool hard-caps at about a third of system memory: 2.5 GB on this 7 GB board, ~20 GB on a 64 GB AGX Orin (NVIDIA/TensorRT-LLM#10894, unresolved). Plain `cuMemAlloc` reaches 6.5 GB. No JetPack release in the Orin's line addresses it; JetPack 7 targets Thor.

## Change
- `core/patches/cudarc.patch` + `fetch-cudarc.sh`, in the existing fetch-and-patch convention: cudarc 0.19.7 with a four-line change so `CUDARC_DISABLE_ASYNC_ALLOC=1` selects synchronous allocation. Unset, the crate behaves exactly as upstream on every platform. Only the CUDA features pull it in. The right home for this is upstream cudarc (an opt-out beside the existing runtime check from cudarc#174); this carries it until then.
- `cuda-no-flash` feature on `kwaainet`: plain candle CUDA, the same as `cuda-windows`, named for what it is on Linux.
- `docs/JETSON.md`: build and run recipe, memory notes, measurements.

## Evidence (isolated 3-node chain, Orin → Mac → mini)
| | CPU | GPU |
|---|---|---|
| Orin hop, 8 blocks + embeddings | 403 ms/token | ~150 ms/token (median 156, min 109) |
| Orin prefill forward, 17 tokens | 730 ms | 120–170 ms |
| chain throughput | 1.8 tok/s | 2.7 tok/s |
| resident | 4.5 GB | 5.25 GB of 7 |

Output identical and correct across runs. Bandwidth floor for this shard is ~66 ms, so candle's per-op CUDA path has ~2× left; flash-attn not tried on sm_87.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -177,6 +177,15 @@ jobs:
         working-directory: core
         run: cargo test -p kwaai-p2p --features kad-multi-protocol
 
+      # Same trap for cudarc, which no CI job compiles (CUDA is release-only):
+      # `[patch.crates-io]` only applies when the patched version satisfies
+      # what candle asks for, and a candle bump past cudarc 0.19 turns the
+      # Jetson fix into a cargo *warning*. Assert the patch is what resolves.
+      - name: Check the cudarc patch is still in the graph
+        if: runner.os == 'Linux'
+        working-directory: core
+        run: bash patches/check-cudarc-patch.sh
+
   # ── 3b. Windows type check (#146) ──────────────────────────────────────────
   # `main` did not compile on Windows for five days between #130 and #142: a
   # field added to a struct was populated only in the `#[cfg(unix)]` arm, and
--- a/.gitignore
+++ b/.gitignore
@@ -110,6 +110,7 @@ tests/rag-bench/results/*.md
 tests/rag-bench/.env
 core/patches/multistream-select/
 core/patches/libp2p-kad/
+core/patches/cudarc/
 
 # Human-readable renderings of tracked source docs (.docx/.pdf from .md).
 # Generated on request, never the source of truth — mirrors the source path,
--- a/core/Cargo.lock
+++ b/core/Cargo.lock
@@ -1192,8 +1192,6 @@ dependencies = [
 [[package]]
 name = "cudarc"
 version = "0.19.7"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "1cea5f10a99e025c1b44ae2354c2d8326b25ddbd0baf76bde8e55cfd4018a2cc"
 dependencies = [
  "float8",
  "half",
--- a/core/Cargo.toml
+++ b/core/Cargo.toml
@@ -29,6 +29,8 @@ default-members = ["crates/kwaai-cli"]
 [patch.crates-io]
 multistream-select = { path = "patches/multistream-select" }
 libp2p-kad = { path = "patches/libp2p-kad" }
+# Behaves exactly as upstream unless CUDARC_DISABLE_ASYNC_ALLOC is set; see patches/README.md.
+cudarc = { path = "patches/cudarc" }
 
 # Root package for examples
 [package]
--- a/core/crates/kwaai-cli/Cargo.toml
+++ b/core/crates/kwaai-cli/Cargo.toml
@@ -125,6 +125,9 @@ default = ["storage", "rag"]
 cuda = ["kwaai-inference/flash-attn"]
 # cuda-windows: CUDA without flash-attn (flash-attn requires Linux nvcc and cannot build on Windows)
 cuda-windows = ["kwaai-inference/cuda"]
+# The same thing under a name that is not a lie on Linux: CUDA without
+# flash-attn, which is what a Jetson builds (docs/JETSON.md).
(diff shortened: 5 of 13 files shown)


RezaRassool (commented):
## Review — one hard break, five smaller

The core idea and the patch itself are sound, and I verified the mechanics rather than eyeballing them: downloaded cudarc 0.19.7 and confirmed the SHA, confirmed `cudarc.patch` applies cleanly with no fuzz, confirmed all four `has_async_alloc` constructor sites are covered and that the alloc/free paths (`core.rs:809`, `1512`, `1535`) branch on the same per-context flag — so there's no alloc/free mismatch. Also built a worktree, ran `fetch-patches.sh`, and confirmed `cargo metadata --locked` passes (the `Cargo.lock` edit is exactly what cargo generates) and that the patch dirs are *not* auto-adopted as workspace members, so `cargo test --all` and the Windows `cargo check --workspace` are unaffected. cudarc 0.17.8 (via `ug-cuda`) stays unpatched but is never activated — `candle-core`'s `cuda` feature uses `candle-ug?/cuda`, and `ug` is off.

Only the first finding is a hard break.

### Findings

1. **`nix/crane.nix:87` — every `nix build` fails.** The new `[patch.crates-io] cudarc` path isn't materialized in the Nix `src` derivation, so the build dies with `failed to load source for dependency 'cudarc'`. Verified empirically: exit 101 with the directory hidden, 0 with it present. `crane.nix` materializes only `multistream-select` and `libp2p-kad`, and this PR touches no `.nix` file at all — while `core/patches/README.md`, which it *does* edit, is where the requirement is written down ("`nix/crane.nix` materializes each patched crate from…").

   Sequencing note: #192 moves `nix/` → `distrib/nix/`, so once that lands the fix belongs in `distrib/nix/crane.nix`. The two PRs also both edit `core/patches/README.md`. Easiest path is to let #192 merge first and then rebase.

2. **`docs/JETSON.md:27`** — the OOM fix is env-var-only. Nothing sets, defaults, or warns about `CUDARC_DISABLE_ASYNC_ALLOC`, so a service-managed Orin gets the same bare `CUDA_ERROR_OUT_OF_MEMORY` this PR exists to fix. Given the target audience is running a headless box, a default or a startup warning would carry more of the value than the doc line does.

3. **`core/Cargo.toml:33`** — the patch pins cudarc 0.19.7 against candle's `^0.19.4`, so a candle bump past 0.19 makes the patch a silent no-op with only a cargo warning, quietly regressing the Jetson. `libp2p-kad` has a CI guard for exactly this; cudarc has none.

4. **`docs/JETSON.md:12`** — the recipe does `cd core` and then `core/patches/fetch-patches.sh`, which only resolves from the repo root, so the line fails outright.

5. **`docs/JETSON.md:4`** — "blocks 0–8" is nine blocks; the command and the closing four-Orin arithmetic both mean 0–7.

6. **`core/crates/kwaai-cli/Cargo.toml:131`** — `cuda-no-flash` duplicates `cuda-windows` verbatim instead of `cuda-windows = ["cuda-no-flash"]`, so the two can drift despite the commit message calling one an alias.

dazwin:
Rebased onto main (#192 moved nix, README merged cleanly) and pushed `901d38ed`. All six addressed.

1. `distrib/nix/crane.nix` materialises `core/patches/cudarc` the same way as the two libp2p crates (fetchurl of the 0.19.7 crate with the sha from `fetch-cudarc.sh`, `patch -p1`, copied into the src derivation). **Not verified with nix** — none on this machine; please `nix build` once before merge.
2. The binary now sets the opt-out itself: `to_candle_device()` (the one site every CUDA entry point goes through) sets `CUDARC_DISABLE_ASYNC_ALLOC=1` on Linux aarch64 when `/etc/nv_tegra_release` exists and the variable is unset, logging one INFO line. The decision is a pure function with a unit test over all four branches. Docs updated. One consequence worth knowing: the patch tests presence, not value, so there is no way to opt back *in* on a Jetson; if you want that, the patch needs `!= "0"` and a re-stamp.
3. New `core/patches/check-cudarc-patch.sh` plus a CI step after the kad guard: asserts `cargo metadata --locked --all-features` resolves `cudarc` at the fetch script's version from the patch path. Verified it fails when the `[patch.crates-io]` line is removed.
4. Recipe path fixed (`bash patches/fetch-patches.sh` after `cd core`).
5. "0–8" → "0–7"; the command and the four-Orin arithmetic were already right.
6. `cuda-windows = ["cuda-no-flash"]`; `cargo tree` confirms `kwaai-inference/cuda` is still reached.

Checks: fmt clean, clippy clean on kwaai-inference and kwaainet, `cargo test -p kwaai-inference --lib` 25 passed. No CUDA build here (no nvcc).

PR #205, opened 2026-09-08
feat(shard): serve a block range with MLX on Apple Silicon

dazwin:
Relates to #198 (stretch goal) and resolves the premise of #117: Macs can serve block shards, on MLX.

## Problem
`shard serve` on macOS routes to whole-model-via-Ollama because candle's Metal backend decodes slower than CPU (#117). That is a property of one backend's kernels, not of Apple GPUs. `kwaai-inference` already had an MLX shard (`mlx_shard.rs`, `mlx` feature) but it decoded incorrectly on `main` and nothing in the CLI could enable or route to it.

## Change
- Ports the two RoPE fixes that lived only on `wip/mlx-shard`: apply RoPE after the transpose (`fast::rope` takes axis −2 as position), and use the uncompiled block forward by default so the decode offset is honoured (`KWAAINET_MLX_COMPILE=1` opts back in). `forward_first/middle/last` ported; clippy-clean (`static mut` → `OnceLock<Mutex<_>>`).
- `kwaainet` gains an `mlx` feature; the loaded shard becomes `LoadedShard::{Candle, Mlx}` behind the same f16 wire format, so MLX and candle halves mix in one chain.
- `shard serve --mlx` (implies `--force-blocks`; errors on a binary built without the feature).

## Evidence (isolated LAN, Llama-3.1-8B f16)
| | candle CPU | MLX |
|---|---|---|
| M1 Max, 16 blocks | ~150 ms/token | 25 ms/token |
| M4 mini, 16 blocks + LM head | ~180 ms/token | ~90 ms/token |
| 2-hop chain decode | 2.4–2.6 tok/s | 5.4–5.7 tok/s |
| 2-hop chain prefill, 15–17 tokens | 5–6 s | 0.3–1.3 s |

Output identical and correct. Compute-only figures are within ~75–80 % of each chip's memory-bandwidth floor. Tests: `mlx_shard` 16 pass / 1 ignored with `--test-threads=1` (MLX's scheduler does not tolerate parallel test threads); `block_rpc` 10/10; clippy clean with and without the feature.

## Known rough edges (follow-ups, not blockers)
- First request after load pays 20–48 s of Metal kernel warmup and can trip a client's 30 s call timeout; a warmup forward at load would fix it.
- MLX loads kernels from `mlx.metallib` (88 MB) resolved via a compile-time path or a file beside the executable; release packaging must ship it.
- Building needs the Metal Toolchain (`xcodebuild -downloadComponent MetalToolchain`), ~5 min per profile for mlx-sys.

### DIFF
--- a/core/Cargo.lock
+++ b/core/Cargo.lock
@@ -3039,6 +3039,7 @@ dependencies = [
  "criterion",
  "dirs",
  "half",
+ "memmap2",
  "mlx-rs",
  "safetensors 0.6.2",
  "serde",
--- a/core/crates/kwaai-cli/Cargo.toml
+++ b/core/crates/kwaai-cli/Cargo.toml
@@ -126,6 +126,8 @@ cuda = ["kwaai-inference/flash-attn"]
 # cuda-windows: CUDA without flash-attn (flash-attn requires Linux nvcc and cannot build on Windows)
 cuda-windows = ["kwaai-inference/cuda"]
 llama-cpp = ["llama-cpp-2"]
+# mlx: serve blocks with MLX on Apple Silicon. Needs Xcode's Metal Toolchain to build.
+mlx = ["kwaai-inference/mlx"]
 storage = ["kwaai-storage", "uuid"]
 rag = ["kwaai-rag", "uuid"]
 # Bootstrap-grade build: also serve the legacy /ipfs/kad/1.0.0 so peers that
--- a/core/crates/kwaai-cli/src/block_rpc.rs
+++ b/core/crates/kwaai-cli/src/block_rpc.rs
@@ -36,7 +36,49 @@ use tracing::{debug, error, info, warn};
 /// serving inference).  The background load task writes `Some(shard)` when
 /// model weights have been fully loaded.  Inference handlers read this cell
 /// without blocking the load task.
-pub type ShardCell = Arc<RwLock<Option<Arc<TransformerShard>>>>;
+pub type ShardCell = Arc<RwLock<Option<Arc<LoadedShard>>>>;
+
+/// The engine behind one `shard serve`: candle everywhere, MLX when built for it.
+/// MLX forwards take `&mut self`, hence the mutex — one request at a time per shard.
+pub enum LoadedShard {
+    Candle(TransformerShard),
+    #[cfg(feature = "mlx")]
+    Mlx(std::sync::Mutex<kwaai_inference::mlx_shard::MlxTransformerShard>),
+}
+
+impl LoadedShard {
+    pub fn blocks(&self) -> (usize, usize) {
+        match self {
+            Self::Candle(s) => (s.start_block, s.end_block),
+            #[cfg(feature = "mlx")]
+            Self::Mlx(m) => {
+                let m = m.lock().unwrap();
+                (m.start_block, m.end_block)
+            }
+        }
+    }
+    pub fn is_first(&self) -> bool {
+        match self {
+            Self::Candle(s) => s.is_first(),
+            #[cfg(feature = "mlx")]
+            Self::Mlx(m) => m.lock().unwrap().is_first(),
+        }
+    }
+    pub fn is_last(&self) -> bool {
+        match self {
+            Self::Candle(s) => s.is_last(),
+            #[cfg(feature = "mlx")]
+            Self::Mlx(m) => m.lock().unwrap().is_last(),
+        }
+    }
+    pub fn gc_sessions(&self) {
+        match self {
+            Self::Candle(s) => s.gc_sessions(),
+            #[cfg(feature = "mlx")]
+            Self::Mlx(m) => m.lock().unwrap().gc_sessions(),
+        }
+    }
+}
 
 // ── Protocol constant ─────────────────────────────────────────────────────────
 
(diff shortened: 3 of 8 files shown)


RezaRassool (commented):
## Review — five findings

The two core fixes are **correct**, and I checked them rather than taking them on trust: `fast::rope` does take axis −2 as the position, so rotating after the `[0,2,1,3]` transpose (and moving the KV concat from axis 1 to axis 2) is right, and it matches candle's `[b, n_heads, seq, head_dim]` RoPE convention — so a mixed candle/MLX chain stays consistent.

Also verified: `LoadedShard` is `Send + Sync` (mlx-rs `Array` is `Send`, so `Mutex<MlxTransformerShard>` is both), `half` resolves to a single version so `as_slice::<half::f16>()` links, and no other consumer of `ShardCell` was missed. `rope_scaling` and explicit `head_dim` are ignored by candle too, so no new divergence there, and the KV-retry/duplication and unbounded-session concerns are identical in the candle path — pre-existing, not yours.

### Findings

1. **`core/crates/kwaai-cli/src/block_rpc.rs:396`** — `is_first()`/`is_last()`/`blocks()` lock the MLX mutex from async context, parking tokio workers for the whole forward, which by your own numbers is a 20–48 s warmup. `shard_cmd.rs:675`'s `gc_sessions()` does the same every 30 s. These fields are immutable after load, so caching them outside the mutex would remove the contention entirely.

2. **`block_rpc.rs:443`** — one panic under `m.lock().unwrap()` poisons the mutex permanently, and because the locks at 396–398 run outside `spawn_blocking` the panic escapes the handler future. The node then stops answering while still announcing coverage, which is the worst shape for the rebalancer — it keeps routing work to a node that can't do any.

3. **`block_rpc.rs:193`** — `f16_bytes_to_array` checks only even byte length, then `Array::from_slice` `assert_eq!`s length against the shape product and panics; the candle twin returns `Err` for the same input. Also `d as i32` wraps for dims above `i32::MAX`.

4. **`core/crates/kwaai-inference/src/mlx_shard.rs:452`** — `std::fs::read` pulls every safetensors file fully into RAM where candle mmaps, plus an f32 intermediate per tensor. A 16 GB Mac loading a 16-block shard from a full 8B f16 snapshot reads ~16 GB before touching a tensor — which on the machines this PR targets is the difference between working and swapping.

5. **`mlx_shard.rs:886`** — `test_rope_position_axis` has **zero assertions** and always passes, and `test_mlx_vs_candle_block0` is `#[ignore]`d and only asserts lengths. So the RoPE fix — the substance of the PR — ships with nothing that fails on the old code. Given the fix is subtle and axis-ordering regressions are silent (wrong logits, not a crash), this is the finding I'd most want addressed: a numeric assertion against a known-good rotation on a small fixture would lock it in.

dazwin:
Rebased onto main (clean) and pushed `22bbc0bb`. All five addressed in one commit, since 1 and 2 share the `LoadedShard` reshape.

1. `LoadedShard::Mlx` now carries `blocks`/`is_first`/`is_last` and a sessions handle copied out at load, so no async context touches the shard mutex. `gc_sessions` takes only the session table, and `run_blocks` no longer holds it across a forward: the session is taken out and put back on success, so the GC never waits on a 20–48 s forward either.
2. The forward was already under `spawn_blocking`; the escape was the three async-side locks, now gone. Inside the forward the lock recovers from poison (`into_inner`), which is sound because a panicking forward mutates nothing durable: weights are read-only and the in-flight session was removed from the table, so it is dropped on unwind rather than left half-appended. Test `lock_recover_survives_poison`.
3. `f16_bytes_to_array` errors on `i32` overflow and on `product(shape) != len`. Worth knowing: the candle twin did not check either — candle 0.10's `Tensor::from_vec` ignores the element count for a plain shape, and a parity test showed `f16_bytes_to_tensor(4 elems, [1,5])` returning `Ok`. Both twins now check, with a test each.
4. Weights are mmapped (`memmap2`, already in the lock via candle, optional under `mlx`); `load_tensor` reinterprets the mapped bytes in place when aligned and hands them to `Array::from_slice` — one copy, MLX-owned, BF16→F16 on the GPU, no f32 detour. Contiguity confirmed by the diagnostic matmul (0.7–0.9 ms) and 18.4 tok/s decode on the 8B benchmark.
5. `test_rope_position_axis` asserts every `[a,c,:]` row equals a reference half-split rotation at position `c`, and that rows differ from the axis-1 reading. New `test_attention_matches_reference` runs one attention layer (2 heads, 1 kv head, seq 3) against a scalar reference and **fails on the pre-fix ordering** (checked by reverting the rope/transpose order and concat axis). `test_mlx_vs_candle_block0` is un-ignored, finds the model via `KWAAI_TEST_MODEL` or the HF cache (skips if absent), and asserts per-position cosine > 0.999 (measured 0.9999+).

Not re-measured: release load time on a 16 GB machine after the mmap change; the debug-build full-8B load was 355 s before the in-place path landed and the budget did not allow a second run. Also unchanged and pre-existing in both paths: a coordinator retry at `seq_pos > 0` after a failed forward runs against an empty KV cache.

Checks: fmt clean, clippy clean with and without `--features mlx`, `cargo test -p kwaai-inference --features mlx --lib mlx_shard` 16 passed, block_rpc tests 13 passed with mlx and 12 without.

PR #206, opened 2026-09-08
fix(p2p): close AutoNAT dial-backs instead of leaking a duplicate connection

dazwin:
AutoNAT's server side dials the probing peer from a fresh port with `PeerCondition::Always` to prove a *new* dial reaches it, then never closes the result. Identify's 5-minute stream keeps every connection under the 10-minute idle timeout, so each probe leaves a permanent duplicate: one per probing peer per refresh, against `max_connections` 100.

The dial-back is recognised by `PortUse::New` (nothing else here allocates a new port, DCUtR included), narrowed to a probe we saw requested and an address it named, and closed on establishment, after seeding kad from it: on a network where identify's addresses are filtered this dial is the only outbound confirmation kad gets.

Measured on the isolated LAN (Orin bootstrap, Mac public peer, mini relay-only): the Mac held **79** connections to the Orin, 78 from ephemeral ports and all owned by the node daemon, accumulating about one per identify round over 8.5 h. The relay-only mini, never asked to dial back, held 1. Rebuilt on this fix: 1 connection, unchanged across identify rounds and a 3-hop inference run. Every reachable production node dials back each bootstrap's probes, so bootstraps and public peers accumulate the same way.

Part of #198 (peers list).

### DIFF
--- a/core/crates/kwaai-p2p/src/service.rs
+++ b/core/crates/kwaai-p2p/src/service.rs
@@ -96,6 +96,38 @@ const RELAY_TICK_INTERVAL: Duration = Duration::from_secs(15);
 /// layer, so a count cap is the closest available approximation.
 const MAX_ADDRESSES_PER_PEER: usize = 6;
 
+/// How long an AutoNAT dial-back stays recognisable after its probe was
+/// requested. The probe's verdict is no guide: its 30 s request timeout fires
+/// while the dial may still be walking up to 16 addresses at the transport's
+/// 10 s each, and the late connection must still be closed.
+const AUTONAT_DIALBACK_TTL: Duration = Duration::from_secs(180);
+
+/// Addresses AutoNAT's server side is about to dial back, each with its expiry.
+///
+/// Keyed by the dial address (it carries `/p2p/<peer>`), so a match is exact.
+/// Entries age out on `AUTONAT_DIALBACK_TTL`, never on the probe's outcome —
+/// bounded by 16 addresses per probing peer. See `take_autonat_dialback`.
+#[derive(Debug, Default)]
+struct PendingDialBacks {
+    expires: HashMap<Multiaddr, Instant>,
+}
+
+impl PendingDialBacks {
+    /// Remember a probe's addresses, dropping whatever has aged out.
+    fn note_request(&mut self, addrs: &[Multiaddr], now: Instant) {
+        self.expires.retain(|_, expiry| now < *expiry);
+        for addr in addrs {
+            self.expires
+                .insert(addr.clone(), now + AUTONAT_DIALBACK_TTL);
+        }
+    }
+
+    /// Consume `addr` if a live probe named it.
+    fn take(&mut self, addr: &Multiaddr, now: Instant) -> bool {
+        self.expires.remove(addr).is_some_and(|expiry| now < expiry)
+    }
+}
+
 /// A connection we are tracking for `list_peers`.
 #[derive(Debug, Clone)]
 struct Connection {
@@ -131,10 +163,9 @@ pub struct NetworkService {
     /// Live connections, per peer, keyed by connection so multiple connections
     /// to one peer are tracked independently.
     connections: HashMap<PeerId, HashMap<ConnectionId, Connection>>,
-    /// Peers with an AutoNAT dial-back probe in flight → the addresses they
-    /// asked us to dial. Consulted once, when the dial-back connects; see
-    /// `is_autonat_dialback`.
-    autonat_probes: HashMap<PeerId, Vec<Multiaddr>>,
+    /// Addresses AutoNAT's server side has been asked to dial back. Consulted
+    /// once, when the dial-back connects; see `take_autonat_dialback`.
+    autonat_dialbacks: PendingDialBacks,
     /// Addresses we were told about peers — the peerstore rust-libp2p does not
     /// have. Consulted ahead of the routing table by
     /// [`Self::candidate_addresses`]; see [`crate::learned_addrs`] for why a
@@ -446,7 +477,7 @@ impl NetworkService {
             pending_routed: HashMap::new(),
             routed_attempts: HashMap::new(),
             connections: HashMap::new(),
-            autonat_probes: HashMap::new(),
+            autonat_dialbacks: PendingDialBacks::default(),
             learned_addrs: LearnedAddrs::new(local_peer_id),
             last_connected: HashMap::new(),
             observed_addrs: HashMap::new(),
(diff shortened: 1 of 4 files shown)


RezaRassool (commented):
## Review — one blocking finding, three smaller

The mechanism is sound, and I checked it against the vendored `libp2p-autonat-0.15.0` / `libp2p-swarm-0.47.1` sources rather than assuming: `AsServer::on_outbound_connection` answers the probe inside `on_swarm_event`, `InboundProbeEvent::Request` really is emitted *before* the `ToSwarm::Dial` it describes, `filter_valid_addrs` appends `/p2p/<peer>` so the `ConnectedPoint::Dialer.address` is genuinely `contains`-comparable, and nothing else in the tree or in dcutr/relay/kad uses `allocate_new_port`. The `PortUse::New` discriminator and the `connections.contains_key` guard are both correct.

Built a worktree at the head, ran the whole `kwaai-p2p` suite (77 tests, green), and confirmed the new test genuinely guards — removing `close_connection` makes it fail with `left: 3, right: 1`. Also merged `origin/main` (the branch is 4 commits behind, merges cleanly) and re-ran: still green.

### Findings

1. **`core/crates/kwaai-p2p/src/service.rs:1653` — the kad seed bypasses both the announceable filter and the per-peer cap.** This is the one I'd block on. `kad.add_address` is called directly instead of through `add_routing_address`, whose own doc says "Every path that learns an address from a *remote claim* goes through here" — and a dial-back address *is* a remote claim: AutoNAT replaces only the IP, the peer chose the port.

   Verified empirically: with the diff, bob's routing table gains `/ip4/127.0.0.1/tcp/53421/p2p/<alice>`; with just those six lines stubbed out it gains nothing. Two concrete consequences. **(a)** Loopback and RFC1918 addresses now enter kad and get served to remote peers in query responses — exactly what the identify path at ~2212 refuses to do, with the metro-win incident in its comment ("A peer's `127.0.0.1` resolves to *us*… crowds out the circuit address that would actually have worked"). `is_announceable` rejects those ranges regardless of `require_global_ips`, which defaults false, so this is the default-config path. **(b)** `MAX_ADDRESSES_PER_PEER` is skipped, and an AutoNAT client sends its listen addrs as the probe addresses — so a symmetric-NAT peer's rotating public ports get seeded uncapped on every 5-minute refresh, which is verbatim the scenario that constant's doc comment exists to prevent.

   Fix is one line: `if is_announceable_with(&stripped, self.require_global_ips) { self.add_routing_address(&peer_id, stripped); }`.

   Filed as #208 — #199 and #185 add the same class of bypass, so it seemed better to agree the rule in one place than three times.

2. **`service.rs:1658` — no `speaks_kad` gate on the seed.** The identify path only seeds addresses for peers whose protocol list contains a kad protocol name. This path has no such gate, and `add_address` *creates* the routing entry (verified: routing table went 0 → 1). Any peer that probes us — a relay-only client, or anything not speaking our kad protocol after the #167 migration — is admitted to our k-buckets on the strength of one dial-back.

3. **`service.rs:2142` — `InboundProbeEvent::Error` doesn't mean the probe's dial has finished, so the leak can still happen.** AutoNAT's request_response timeout is 30 s, but the dial-back walks up to `max_peer_addresses` (16) addresses *sequentially* (`override_dial_concurrency_factor(1)`) with no per-dial deadline. A NATed peer offering several black-holed ports — the fleet's common case — trips `InboundFailure::Timeout` → `Error` while the dial is still walking. This code then removes `autonat_probes[peer]`, so when a later address does connect, `is_autonat_dialback` returns false and the duplicate connection is left open exactly as before the fix. Same shape for a second probe refused with "dial-back already ongoing". Keying by the dial's `ConnectionId` (available from `DialOpts::connection_id()`, though autonat doesn't expose it) or expiring on a timer would decouple it from the probe verdict.

4. **`core/crates/kwaai-p2p/tests/swarm.rs:578` — the regression guard can silently stop guarding.** The only synchronisation is a bare `sleep(8s)` with a single un-retried assertion, and the assertion's passing value (1) is also its vacuous value. It depends on autonat's 5 s `boot_delay` *and* on identify having first delivered an external-address candidate; if that slips past 8 s on a loaded runner, the test passes having tested nothing. The file's own `eventually`/`SETTLE_TIMEOUT` convention plus a positive check that the count first rose above 1 would make it deterministic. Note also that both nodes probe each other (the failure mode shows 3 connections, not 2), so bob transiently holds alice's in-flight dial-back — sampling inside that window can fail spuriously.

dazwin:
Rebased onto main (clean) and pushed `ed56f16a`. All four addressed.

1. The seed now goes `speaks_kad` → `is_announceable_with(stripped, require_global_ips)` → `add_routing_address` (cap and dedupe). The swarm test asserts the loopback dial-back address does not enter the routing table, and fails with the raw `kad.add_address` restored.
2. `speaks_kad` gate added; a peer with no identify yet reads as no, so no entry is created. No isolating test on loopback: finding 1's filter rejects every address there, so the gate cannot be observed separately without a non-loopback listener. Noted as a follow-up if you want one.
3. Pending dial-backs are keyed by dial address with a 180 s expiry (16 addresses × the transport's 10 s connect timeout, past autonat's 30 s). `InboundProbeEvent::Error`/`Response` no longer clear anything; entries are consumed on match and purged on the next request. Unit tests cover recognition at +60 s, single consumption and expiry. I did not write the black-holed-address integration test: no reliable black hole exists on loopback and it would need >30 s, outside the file's 20 s convention.
4. The bare sleep is replaced by `eventually` gates. The dial-back is closed before it is ever inserted into `connections`, so the count cannot be seen rising; the test instead captures the service's "closing autonat dial-back" debug line for both sides, then waits for the count to settle at 1, re-samples, and checks the routing table. Removing `close_connection` makes it time out. One flake found and fixed on the way: with both `boot_delay`s firing together one side's request can be routed over the other's ms-long dial-back and die with it, so the swarms are spawned 1 s apart (12/12 after, 1/13 before).

Forced by finding 1: four pre-existing loopback tests (`dht_service::routing_peers_reach_the_storage` and three routed-dial tests in `service_unary.rs`) only passed because a dialling-in peer leaked into the bootstrap's routing table via its loopback dial-back address. They now seed through the operator path (`add_kad_address`) and their comments are corrected.

One behavioural note for production: a request that request-response happens to route over the ms-long dial-back connection fails with `ConnectionClosed` and autonat retries 30 s later. Negligible against the leak, but it will show up as an occasional "probe failed once".

Checks: fmt clean, clippy clean, `cargo test -p kwaai-p2p --no-fail-fast` 207 passed.

RezaRassool:
## Rebase needed — one file

**#185 has landed**, which matters here: it is one of the three PRs in #208, and it reworked the address surface this PR seeds into. The `take_autonat_dialback` / `PendingDialBacks` work should carry over cleanly, but the seeding call now sits next to `LearnedAddrs` — worth confirming the dial-back address still takes the `speaks_kad` + `is_announceable_with` path and is not duplicated by the learned-address route.

The fixes on this PR were the cleanest of the batch, so this is expected to be a short rebase.

---

dazwin:
Rebased onto main `8c5660b7` and pushed `6744bbba`. Two hunks, both the struct field and initialiser where #185's `learned_addrs` landed beside `autonat_dialbacks`; kept both. Interdiff against the pre-rebase branch differs only in context lines.

Your two questions, read at the rebased head: the seed is `take_autonat_dialback` → `speaks_kad` → `is_announceable_with(stripped, require_global_ips)` → `add_routing_address` (service.rs:1824–1836), and it runs before the connection is inserted into `connections`, so a recognised dial-back never becomes a tracked connection. `learned_addrs` has one insert site, `ConnectPeerWithAddrs`; no identify, AutoNAT or connection-established path writes it, so a dial-back rejected by this gate cannot be admitted through the learned route.

Swarm suite three times: 16/16 each. Red/green on `an_autonat_dial_back_does_not_leave_a_duplicate_connection`: with the handler disabled it times out waiting for bob to close the dial-back; restored, passes. One honest note on that test: its failure signal is the "closing autonat dial-back" log line, because a recognised dial-back is closed before it could ever appear in `list_peers`, so the `== 1` count cannot be the load-bearing assertion.

Checks: gate clean, fmt, clippy, `kwaai-p2p --no-fail-fast` all suites green, `kwaainet --bin` 284.

PR #209, opened 2026-09-09
fix(p2p): a call to our own peer id loops back to the local handler

dazwin:
libp2p refuses to dial its own peer id, so the local node was the one peer the daemon could never connect to or call. Every client of the daemon had to special-case itself, and the chain builder's pre-dial reported the local node as unreachable on every session (the "you: unreachable" row in the GUI).

A routed request addressed to the local peer id now resolves inside the daemon:

- connect succeeds (the node is trivially connected to itself),
- a unary call dispatches to the registered handler exactly as an inbound call does; no handler registered is the same protocol refusal a remote gives during negotiation,
- a raw stream is refused, since there is no loopback transport for one.

Tests: loopback round trip, refusal without a handler, connect to self, in `tests/service_unary.rs`.

Demonstrated on the isolated LAN: `p2p peers connect --addr /p2p/<own id>` answers Connected, and a 3-hop session's local hop runs through the daemon at 34 ms per token.

The client-side cleanup that this enables is #210. Part of #198.

### DIFF
--- a/core/crates/kwaai-p2p/src/service.rs
+++ b/core/crates/kwaai-p2p/src/service.rs
@@ -1172,6 +1172,10 @@ impl NetworkService {
     /// first when we have none — Go's routed-host semantics (see
     /// [`RoutedRequest`]). With a live connection, dispatch is immediate.
     fn dispatch_routed(&mut self, peer: PeerId, request: RoutedRequest) {
+        if peer == *self.swarm.local_peer_id() {
+            self.dispatch_local(request);
+            return;
+        }
         if self.swarm.is_connected(&peer) {
             self.forward_routed(peer, request);
             return;
@@ -1188,6 +1192,50 @@ impl NetworkService {
 
         if !self.dial_routed(peer) {
             self.start_routed_lookup(peer);
+        }
+    }
+
+    /// A routed request addressed to ourselves: the swarm refuses to dial its
+    /// own peer id, but we are trivially connected and our unary handlers are
+    /// right here. A raw stream has no loopback transport and is refused.
+    fn dispatch_local(&mut self, request: RoutedRequest) {
+        let local = *self.swarm.local_peer_id();
+        match request {
+            RoutedRequest::Connect { reply } => {
+                let _ = reply.send(Ok(local));
+            }
+            RoutedRequest::Unary { proto, data, reply } => {
+                // The same refusal a remote gives during negotiation.
+                if !self.unary_handlers.contains_key(&proto) {
+                    let _ = reply.send(Err(unary::UnaryError::UnsupportedProtocol(proto)));
+                    return;
+                }
+                let (responder, result) = oneshot::channel();
+                self.dispatch_unary(
+                    local,
+                    unary::InboundRequest {
+                        proto: UnaryProtocol::new(proto),
+                        data,
+                        responder,
+                    },
+                );
+                tokio::spawn(async move {
+                    let outcome = match result.await {
+                        Ok(Ok(data)) => Ok(data),
+                        Ok(Err(e)) => Err(unary::UnaryError::Remote(e)),
+                        Err(_) => Err(unary::UnaryError::Wire(
+                            "local handler dropped the call".to_string(),
+                        )),
+                    };
+                    let _ = reply.send(outcome);
+                });
+            }
+            RoutedRequest::RawStream { protos, reply } => {
+                let _ = reply.send(Err(raw_stream::RawStreamError::Io(format!(
+                    "no loopback transport for a stream to self ({})",
+                    protos.join(", ")
+                ))));
+            }
         }
     }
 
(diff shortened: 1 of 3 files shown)


RezaRassool (commented):
## Review — reviewed at `b5529972`

The design is right: the swarm refuses to dial its own peer id, so a node was the one peer it could not call, and looping back through the local handlers in `dispatch_local` is the correct shape. The three arms are all sensible — a `Connect` to self succeeds trivially, an unsupported protocol gives the same refusal a remote would give during negotiation, and a raw stream is refused because there genuinely is no loopback transport. The tests cover all three, and I confirmed the unary loopback test actually exercises the new path.

Things I checked and cleared, so they needn't be re-argued: there is **no re-entrancy deadlock** — `add_unary_handler_boxed` spawns per call and the daemon's client reader dispatches `RequestHandling` frames on its own task, so a self-call from inside a handler of the same protocol does not stall the dispatcher. The `contains_key` pre-check followed by `dispatch_unary`'s own lookup is redundant, but both outcomes map to the same error. And the `tokio::spawn` inside a command handler deviates from the file's "must never await or block" rule without actually blocking the loop.

### Finding

**`service.rs` → `dispatch_local`, the `RoutedRequest::Unary` arm — the loopback path enforces no timeout, so a hung local handler hangs the caller forever.**

Both remote paths bound the handler by `Config::request_timeout` (30 s by default):

- outbound — `unary.rs:486`: `match tokio::time::timeout(timeout, exchange).await { … Err(_) => Err(UnaryError::Timeout) }`
- inbound serving — `unary.rs:407`: `match tokio::time::timeout(timeout, result_rx).await { … Err(_) => Err("handler timed out") }`

`dispatch_local` instead does a bare `tokio::spawn(async move { … result.await … })`, and nothing upstream adds a bound.

What makes this worth fixing rather than noting is that callers are *told* not to bring their own timeout. `node_native.rs`, on `REQUEST_TIMEOUT`:

> A 30 s cap on every `call_unary_handler`. The native handle enforces it itself (`NetworkConfig::request_timeout`), so `send_records_via_handle` needs no timeout of its own.

So this is a hole in a documented guarantee, not a missing nicety. Concretely, and in this PR's own motivating case: `shard run` with the local node in the chain reaches `block_rpc::call_inference` → `call_unary_handler(own_id, INFERENCE_PROTO, …)`, which deliberately has no timeout of its own. If the local block server stalls — the known llama.cpp session-pooling and concurrent-GPU issues — a *remote* hop returns `Timeout` and the session fails that peer over, while the *local* hop blocks the generation loop indefinitely with no failover and no error. Wrapping the `result.await` in `tokio::time::timeout(self.config.request_timeout, …)` → `UnaryError::Timeout` restores the contract.

### Two notes, not findings

- **`Connect` to self answers `Ok(local)` but registers nothing.** Nothing is inserted into `self.connections`, so `list_peers` still omits the node and `DisconnectPeer` for the same id returns `PeerNotFound`. `p2p peers connect --addr /p2p/<own id>` therefore prints Connected, and `p2p peers list` then omits it. I could not find a caller that connects-then-verifies and would actually be misled — and #210 deliberately compensates by excluding self from `unreached` — so this reads as a consistency wrinkle rather than a bug. Worth a comment saying self is deliberately invisible to `list_peers`, if that is the intent.
- **Only a bare `/p2p/<id>` self-connect is fixed.** `Command::ConnectPeer` reaches `dispatch_routed` only under `Some(peer) if strip_p2p(&addr).is_empty()`; anything carrying a transport component falls to `self.dial(addr)` and is rejected with `DialError::LocalPeerId`. A node failing to dial its own full address is arguably fine, so I am flagging it only in case the self check belongs in the `ConnectPeer` arm (or in `dial`) rather than inside `dispatch_routed`.

### Merge order

**This must merge before #210.** `dispatch_local` exists only here — I verified it is absent from #210, #203, #185 and `main`, and #210's ancestry is #203 → #185 → main, so it never picks this up. #210 deletes the local TCP bypass and relies on this loopback, so in the other order a self hop dials its own peer id, falls to a kad walk for itself, and fails where the bypass previously succeeded. Nothing in git enforces it.

dazwin:
Thanks, agreed on the timeout: cf96b0c6 bounds the loopback await by the unary behaviour's `request_timeout` and maps expiry to `UnaryError::Timeout`, with a test that hangs the local handler and expects `P2PError::Timeout`. It fails without the fix.

On the two notes: self stays out of `connections` on purpose. There is no connection to close and nothing to count, and #210 excludes self from `unreached` for the same reason. I added a comment on the Connect arm saying so. The full-address self-dial I have left as the swarm's `LocalPeerId` refusal, since the only caller that dials self is the chain builder and it goes through the bare `/p2p/<id>` path. Happy to move the check into `ConnectPeer` if you would rather it cover both.

Merge order: #210 now has this branch merged in, so it cannot land first.

PR #210, opened 2026-09-09
fix(shard): call our own shard like any other peer

dazwin:
Stacked on #203. Requires #209 on the daemon.

The chain builder carried three special cases for the local node: a hop that side-stepped the swarm through a local TCP port, a "prefer self" ranking gated on that port's file, and a pinning pass that counted the local node as unreached because the daemon never lists itself as a connection.

With the daemon looping a call to its own peer id back to the local handler, none of them is needed. The hop calls our own shard through the daemon like any other peer, the ranking prefers self unconditionally, and `unreached_peers` leaves ourselves out. The pre-dial now reports the local node as dialed instead of unreachable.

The local TCP bypass server itself stays: the standalone `shard run` path still reuses a running `shard serve`'s loaded model through it.

Part of #198.

### DIFF
--- a/core/crates/kwaai-cli/src/shard_api.rs
+++ b/core/crates/kwaai-cli/src/shard_api.rs
@@ -242,7 +242,9 @@ async fn run_inference(
     let mut client_guard = state.client.lock().await;
     let mut failed_peers: std::collections::HashSet<libp2p::PeerId> =
         std::collections::HashSet::new();
-    let mut unreached = crate::shard_cmd::unreached_peers(&mut client_guard, &state.chain).await;
+    let mut unreached =
+        crate::shard_cmd::unreached_peers(&mut client_guard, &state.chain, &state.our_peer_id)
+            .await;
 
     // Pin peer path for this request so KV-caches stay coherent.
     let mut pinned_path = match crate::shard_cmd::build_pinned_path_ranked(
@@ -286,8 +288,12 @@ async fn run_inference(
             Ok(r) => r,
             Err(e) => {
                 // Try rebuilding path excluding failed peer
-                unreached =
-                    crate::shard_cmd::unreached_peers(&mut client_guard, &state.chain).await;
+                unreached = crate::shard_cmd::unreached_peers(
+                    &mut client_guard,
+                    &state.chain,
+                    &state.our_peer_id,
+                )
+                .await;
                 match crate::shard_cmd::build_pinned_path_ranked(
                     &state.chain,
                     state.total_blocks,
--- a/core/crates/kwaai-cli/src/shard_cmd.rs
+++ b/core/crates/kwaai-cli/src/shard_cmd.rs
@@ -576,8 +576,8 @@ async fn cmd_shard_serve(args: ShardServeArgs) -> Result<ShardServeExit> {
     print_info("Loading model in background. Requests return 'warming up' until ready.");
     print_separator();
 
-    // Start local TCP bypass server so `shard run` on the same machine can
-    // call us without triggering libp2p's "dial to self" rejection.
+    // Local TCP bypass server for `shard local`, which runs without a daemon
+    // and reuses this loaded model. `shard run` reaches us through the daemon.
     let _ = std::fs::create_dir_all(crate::config::run_dir());
     match start_local_inference_server(shard_cell.clone(), device.clone()).await {
         Ok(port) => {
@@ -1755,8 +1755,8 @@ pub enum ShardRunEvent {
 ///   peer errors) are unchanged — they're acceptable in the daemon context
 ///   today and out of scope for this refactor.
 /// - The local node is honoured as a hop when it appears in the discovered
-///   chain — `forward_through_chain` already calls into the local TCP bypass
-///   server for `our_peer_id`, just like the CLI path.
+///   chain — `forward_through_chain` calls it through the daemon like any
+///   other peer, just like the CLI path.
 /// - On an empty chain (no peers serving the model) we poll every 2 s for up
 ///   to 30 s before yielding
 ///   `Error(anyhow!("no peers serving model …"))`.
--- a/core/crates/kwaai-cli/src/shard_cmd.rs
+++ b/core/crates/kwaai-cli/src/shard_cmd.rs
@@ -1439,7 +1439,7 @@ pub async fn cmd_shard_run(args: ShardRunArgs) -> Result<()> {
             .connect_peer_with_addrs(&entry.peer_id, &entry.dial_addrs)
             .await;
     }
-    let mut unreached = unreached_peers(&mut client, &chain).await;
+    let mut unreached = unreached_peers(&mut client, &chain, &our_peer_id).await;
 
     // ── Inference loop ────────────────────────────────────────────────────────
     let mut generated_ids: Vec<u32> = Vec::new();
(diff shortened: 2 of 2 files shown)


RezaRassool (commented):
## Review — reviewed at `b43cb717`

The refactor is coherent, and I verified the mechanical parts rather than assuming them: all seven `unreached_peers` call sites were updated, `hop_candidates`' dropped `self_dispatchable` parameter has no other callers, `local_inference_call`/`local_server_port_file` remain live for `cmd_shard_local`, and both the p2p handler and the TCP bypass call `handle_inference_request` on the same `shard_cell`, so switching transports preserves the session KV cache. `cargo check -p kwaainet --tests` is clean and the `shard_cmd` unit tests pass, including the new `our_own_node_is_never_unreached`.

Treating our own shard as just another peer is the right direction — it removes a whole special case rather than patching it.

### Finding — merge order, and it is not enforced by git

**This depends on #209 and must not merge first.** The daemon loopback this PR relies on is `dispatch_local`, added in #209. I verified it is absent from this branch, from its base `fix/pin-dialed-peers` (#203), from #185, and from `main` — this PR's ancestry is #203 → #185 → main, and #209 branches off `main` separately, so this branch never picks it up.

Merged in this state, a self hop parks in `dispatch_routed`, dials its own peer id (which the swarm rejects with `LocalPeerId`), falls through to a kad walk for itself, and eventually fails — exactly where the bypass this PR deletes used to succeed. The regression is silent until someone runs a chain that includes the local node, which is the common single-machine case.

Nothing in the branch topology or CI enforces the ordering, so it needs to be held manually or rebased onto #209.

### Two stale comments

Both are cheap, and the second is the kind that causes a later mistake:

- **`shard_cmd.rs` → `run_streaming`'s doc comment** still says: "`forward_through_chain` already calls into the local TCP bypass server for `our_peer_id`, just like the CLI path." This PR removes exactly that.
- **`shard_cmd.rs` → the `shard serve` bypass-server comment** still reads: "Start local TCP bypass server so `shard run` on the same machine can call us…". After this diff only `shard local` uses it — so the comment invites a future reader to delete a server that is still needed.

### A design question rather than a finding

Excluding self from `unreached` unconditionally is what I asked for in the #203 review, so this is me arguing with my own earlier note rather than with the PR.

But the two changes together — self is never `unreached`, *and* the port-file bypass is gone — leave nothing checking whether our own shard is actually serving. Previously `self_dispatchable` (port file present) was that signal. So a stale self-announcement now gets pinned ahead of live remotes, costing a failed first hop and a lost KV cache on every fresh session.

Not obviously wrong: the daemon call failing is itself a signal, just a more expensive one. Worth a deliberate answer either way. Relatedly, there is no fallback if the daemon-mediated call cannot work — an old daemon left running after an auto-update is a documented failure mode in this tree — and the bypass server and port file are still present and free to retry against.

dazwin:
Both comments fixed in 19d44322, and this branch now carries #209 as an ancestor (8df04423) so the ordering is enforced by git rather than by memory. The diff will shrink to the shard changes once #209 merges.

On the design question: yes, the intent is that a failed daemon call is the liveness signal for our own shard, not a port file. The port file only proved that `shard serve` had started, not that the model was serving, and a stale self-announcement costs one failed hop either way. With #209's timeout that failure is now bounded at `request_timeout` rather than hanging, which is what made me comfortable dropping the pre-check. I have not added a bypass fallback for the old-daemon-after-auto-update case: the fix for that is the daemon restart, and a silent transport fallback would hide it. If you think the fallback belongs here I will add it as a follow-up rather than widen this PR.

RezaRassool:
## Retarget and rebase — a different problem from the other five

This one is not a content conflict. Its dependencies have all landed, but **via squash**, so git cannot see that.

**What happened.** This PR's base is still `fix/pin-dialed-peers`, and its history carries 22 commits — including #185's whole series, #203's (`81ad2eba`) and #209's (`cf96b0c6`). All three merged to `main` as single squashed commits under new SHAs, so none of the original commits is an ancestor of `main`. Git therefore sees the same changes twice, from two unrelated histories, and reports a conflict in `shard_api.rs` and `shard_cmd.rs` — even though the content is already there.

**The fix is to retarget, not to merge `main` in.** Merging would replay work that has already landed. Roughly:

```bash
git fetch origin
git checkout fix/self-dial-client
git rebase --onto origin/main 81ad2eba    # drop everything already squashed to main
```

and set the PR's base to `main`. What should survive is just this PR's own work: `b43cb717` (call our own shard like any other peer) and `19d44322` (the bypass-server comment). If the rebase brings anything else, it is a commit that already landed.

**Good news on the ordering hazard.** My earlier review said this must not merge before #209, because it deletes the local TCP bypass and relies on the daemon loopback `dispatch_local` provides. **#209 merged** (`8c5660b7`) and `dispatch_local` is on `main`, so that constraint is discharged — nothing to sequence around any more.

**One thing that does still need fixing: this PR has no test CI.** Its only check is the release workflow's `plan` job — no `Test`, `Clippy`, `Format`, `Windows check` or `WASM`. It reads green on GitHub and is not. #203, #193 and #190 were all in the same state and picked up real runs when they rebased, so the retarget above should fix this too — but worth confirming the eight checks appear before it goes in, because nothing else in the queue is currently unverified this way.

**Already addressed from the review, thanks:** `19d44322` covers the stale `shard serve` comment that still named `shard run` as the bypass consumer.

**Still open, and worth a decision rather than a fix:** self is now never `unreached`, *and* the port-file bypass is gone, so nothing checks whether our own shard is actually serving — a stale self-announcement gets pinned ahead of live remotes. That is me arguing with my own #203 note, which asked for the first half. The daemon call failing is itself a signal, just a more expensive one. Either answer is defensible; it would be good to have it stated.

---

**Context.** `main` is now `8c5660b7`. Nine PRs landed 10–11 September: #214, #185, #193, #204, #205, #200, #203, #194, #209. Re-run `core/patches/fetch-patches.sh` after pulling — #204 added a third patched crate (`cudarc`) and `cargo` will not load the workspace without it. A new CI gate is also live (#214, *Orphaned doc comments*); it is not flagging this PR today, but it will flag a rebase that introduces one.

dazwin:
Retargeted to `main` and rebased with `git rebase --onto origin/main 81ad2eba`; pushed `aba43657`. Exactly two commits and two files survive (`shard_cmd.rs`, `shard_api.rs`; 50 insertions, 64 deletions). #209's loopback commit was skipped after checking that the sorted +/- lines of the whole #209 branch and of main's squash `8c5660b7` are identical in both directions.

On the open design question, stated for the record: with the port-file bypass gone, nothing pre-checks that our own shard is serving. A self hop goes through the same `call_block_forward` → `dispatch_local` path as any remote, bounded by `request_timeout`; a failure lands in the generic failed-hop arm, self joins `failed_peers`, and the caller's rebuild path re-pins without it. So a stale self-announcement costs one bounded failed hop per session, then is excluded. That is deliberate: the daemon call failing is the liveness signal, and a silent transport fallback would hide a daemon that needs restarting.

Uncovered by unit tests, said plainly: the removal of the local bypass in `forward_through_chain`. The `unreached_in` self filter has a red/green test (`our_own_node_is_never_unreached`), and the loopback itself is covered on main by `service_unary`.

Checks: gate clean, fmt, clippy, `kwaainet --bin` 285, `service_unary` 23. The eight CI checks should now run against `main`; please confirm they appear.

PR #220, opened 2026-09-16
docs: cap a PR pass at 5, matching the retrospective's rule 8

RezaRassool:
## Why

The 8–11 September pass reviewed 19 PRs in one sweep and the entire review went
stale before it could be acted on. #191 alone deleted 9,779 lines across 66 files
and forced a rebase across the queue, so findings written against the pre-merge
heads had to be re-checked rather than acted on.

`CLAUDE.md` already records "merge the churn first, then review". What it did not
record is a limit on how much churn to take in one sitting.

## What

A cap of **5 PRs per session** at the top of `## Merging PRs`, stating the reason
rather than only the number: the cost of a big batch is not the merging, it is
that every unmerged PR behind the batch is being reviewed against a base you are
about to change.

Five is **rule 8 of `docs/REVIEW_INTEGRITY_RETROSPECTIVE.md`** (#213), adopted
here so the repository states one cap rather than two. The corollary — do not open
a review you cannot finish and act on in the same session — follows from the same
observation rule 8 makes, that quality fell visibly with volume.

## Not answered here

§5 of the retrospective asks whether rules 1–7 belong in a distinct
`## Verification discipline` section, separate from `## Reviewing PRs`. This PR
deliberately does not settle that: it places one merge-batch rule in the section
that already governs merging, and leaves the structural question open.

Rules 1–7 and 9 of the retrospective are also not adopted here.

## Provenance

Written during the 14–15 September pass, which merged six PRs — #169, #164, #189,
#213, #217, #216 — and was itself the first test of the cap. It exceeded it: five
were merged under the rule, and #216 the following day.

https://claude.ai/code/session_012cFnT3UbZTE23ZSzrSkt36


dazwin (approved):
Approved. Docs-only, CI green, and the rule matches rule 8 of the retrospective it cites.

One optional wording nit, not blocking: "At most 5 PRs per session" reads on first pass as a limit on how many PRs a contributor may open. Suggest "At most 5 PRs reviewed and merged per session; there is no limit on how many a contributor opens."

PR #221, opened 2026-09-16
fix(installer): source cargo env in same shell; skip broken CUDA upgrade on musl fallback

RezaRassool:
## Summary
- The README's `curl | sh` install command left `kwaainet` unreachable in the same terminal — the installer only edits rc files for *future* shells. The recommended command now chains `[ -f "$HOME/.cargo/env" ] && . "$HOME/.cargo/env"` so it works immediately, plus a confirm step.
- On hosts with an old glibc (RHEL/EL9, Amazon Linux 2-family), the CI-injected CUDA auto-detection unconditionally "upgraded" any GPU host to the gnu-linked CUDA archive — the only one ever published — which then fails at runtime (`error while loading shared libraries: libcudart.so.12`) on exactly the systems that needed the musl fallback in the first place. The CUDA upgrade in `scripts/patch-installer-cuda.sh` is now restricted to `gnu` targets only.
- Added a README note pointing GPU + old-glibc hosts at the Nix build (`nix build github:Kwaai-AI-Lab/KwaaiNet`) as the way to still get CUDA acceleration, since it's unaffected by the host's system glibc.

## Test plan
- [x] Reproduced both bugs against a real, live release installer on an EL9/glibc-2.34 host with a real NVIDIA GPU.
- [x] Reconstructed a pristine (pre-CI-patch) installer and applied the fixed `patch-installer-cuda.sh` exactly once, mirroring real CI, then ran it end-to-end: installer now prints "NVIDIA GPU detected, but this system's glibc is too old for the CUDA-enabled build — installing CPU build instead", stays on the working musl build, and `kwaainet --version` succeeds in the same shell immediately after the two-line install command (no restart needed).
- [x] Unit-tested `_try_cuda_upgrade` in isolation for both branches: `gnu` target still upgrades to CUDA correctly (unchanged behavior); `musl` target now skips the upgrade and warns instead of installing a broken binary.
- [x] `sh -n` syntax-checked both the patch script and the resulting patched installer.
- [x] No Rust code changed — `cargo test` not applicable.

### DIFF
--- a/README.md
+++ b/README.md
@@ -108,6 +108,16 @@ This quickstart shows how to install the native Rust CLI, start a node, and send
 
 ```bash
 curl --proto '=https' --tlsv1.2 -LsSf https://github.com/Kwaai-AI-Lab/KwaaiNet/releases/latest/download/kwaainet-installer.sh | sh
+[ -f "$HOME/.cargo/env" ] && . "$HOME/.cargo/env"
+```
+
+The second line puts `kwaainet` on `PATH` in the *current* shell — the
+installer itself only updates `.bashrc`/`.profile`/`.zshrc` for future
+shells, so without it you'd need to open a new terminal first. Then
+confirm:
+
+```bash
+kwaainet --version
 ```
 
 **PowerShell installer (Windows):**
@@ -181,6 +191,13 @@ On Linux and Windows machines with an NVIDIA GPU, the installer automatically de
 ```bash
 kwaainet benchmark --gpu
 ```
+
+Only a glibc-linked CUDA build is published, so on a host whose glibc is
+too old for it (e.g. RHEL/EL9, Amazon Linux 2-family), the shell installer
+falls back to the CPU build instead of installing a binary that can't run.
+To get CUDA acceleration there anyway, use the Nix build above — it links
+against its own glibc and isn't affected by the host's system glibc at
+all.
 
 **Apple Silicon (Metal):**
 
--- a/scripts/patch-installer-cuda.sh
+++ b/scripts/patch-installer-cuda.sh
@@ -22,10 +22,18 @@ cat > "$CUDA_FUNC" <<'CUDA_PATCH'
 _cuda_selected=""
 _cuda_libs_staged=""
 _try_cuda_upgrade() {
-    # Only applies to x86_64 linux targets (gnu or musl fallback)
+    # Only applies to the x86_64 gnu-linux target. No musl-linked CUDA build
+    # is published, so a musl-fallback host (glibc too old for the gnu
+    # build) must not be "upgraded" to the gnu-cuda archive — it would
+    # install a binary that can't run there (missing libcudart at runtime).
     case "${_artifact_name:-}" in
         kwaainet-x86_64-unknown-linux-gnu*) ;;
-        kwaainet-x86_64-unknown-linux-musl*) ;;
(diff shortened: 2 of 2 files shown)


dazwin:
Thanks for chasing this down on a real EL9 box. I'd like to push back on both halves, though, because I think each fixes the symptom one level away from the cause.

## PATH: fix the install location, not the README

Popular `curl | sh` installers do tell people to reload their shell — that part is fine:

- **rustup** adds `. "$HOME/.cargo/env"` to rc files and prints *"To configure your current shell, you need to source the corresponding env file under ~/.cargo (note the leading DOT)"*. Rustup owns `~/.cargo`, so that's legitimate for it.
- **uv** is generated by the same cargo-dist we use, but sets `install-path = ["$XDG_BIN_HOME/", "$XDG_DATA_HOME/../bin", "~/.local/bin"]`. It lands in `~/.local/bin`, which most distros already have on PATH, and only if it had to edit an rc file does it print *"To add ~/.local/bin to your PATH, either restart your shell or run: source ~/.local/bin/env"*.

Our problem is that `[workspace.metadata.dist]` has no `install-path`, so cargo-dist falls into its `CARGO_HOME` mode: it creates `~/.cargo/env` on machines that have never had Rust and writes a rustup-style line into their rc files. A kwaainet user shouldn't know or care about a cargo environment. The fix is uv's `install-path` line in `core/Cargo.toml`, after which the installer's own post-install hint (it already prints the exact `source` line whenever it had to touch an rc file) is all the README needs.

The added README line also hardcodes `~/.cargo/env`, which is wrong whenever `KWAAINET_INSTALL_DIR` or `CARGO_HOME` is set — the `[ -f ]` guard makes it silently do nothing there.

## CUDA: the binary isn't too new for EL9, the archive is empty

I pulled `kwaainet-x86_64-unknown-linux-gnu-cuda.tar.xz` from v0.7.1 and checked:

| | |
|---|---|
| Highest `GLIBC_*` symbol version the binary imports | 2.28 |
| EL9 glibc | 2.34 |
| Files in the archive | `kwaainet` only |
| Release log, "Package CUDA archive" step | `CUDA libs bundled: 0 files` |
| `NEEDED` | libcudart.so.12, libcublas.so.12, libcurand.so.10, libcuda.so.1 |
| `RUNPATH` | none |

So the CUDA binary runs on glibc ≥ 2.28 — which is exactly why 2ba47e24 moved the build into `manylinux_2_28` and extended the upgrade to the musl-fallback path "for Rocky Linux 8 compat". This PR reverts that intent. The real cause of `error while loading shared libraries: libcudart.so.12` is that the archive ships no runtime libraries, and even when it does, the binary has no `$ORIGIN` rpath to find them next to itself. That fails the same way on Ubuntu without a system CUDA toolkit; the test plan passed on EL9 only because the change now selects the CPU build there.

What I think actually fixes it:

1. In `release.yml`, make the packaging loop actually find the libs. `find /usr/local/cuda-12.6/lib64 ...` almost certainly returns nothing because `lib64` is a symlink in the CUDA rpm layout and `find` doesn't follow it — `find -L` or a trailing slash.
2. Fail the job when the bundled count is 0 instead of echoing it.
3. Build the CUDA target with `-C link-arg=-Wl,-rpath,$ORIGIN` so the bundled libs are found without `LD_LIBRARY_PATH`.
4. Drop the musl skip and the Nix paragraph.

Happy to review a revised version along those lines.

PR #222, opened 2026-09-16
fix(inference): match reinstall hint to documented install shell

RezaRassool:
## Summary
- The GPU-detected-but-no-CUDA-support messages (`require_gpu()` and `detect_best_logged()` in `core/crates/kwaai-inference/src/lib.rs`) suggested reinstalling via `curl ... | bash`, while README.md documents `curl ... | sh` for the same installer.
- Both work identically on any POSIX shell, but the mismatch is confusing when a user compares the two side by side.

## Test plan
- [x] `cargo check -p kwaai-inference` passes.
- [x] `grep -n '| bash' core/crates/kwaai-inference/src/lib.rs` now returns nothing.


dazwin (approved):
Matches README, CONTRIBUTING and docs/getting-started-node.md; CI green. The reinstall hint itself is in the wrong crate (it cannot see packaged installs, and the string is now copied three times) — tracked separately rather than here.

PR #224, opened 2026-09-16
fix(ollama): trust /api/tags instead of a permission-blind filesystem scan

RezaRassool:
## Summary
- `ollama::readiness()` already called Ollama's HTTP API (`GET /api/tags`) to check it was up, then discarded the response body and fell back to `list_local_models()` — a manifest-directory filesystem scan that silently returns empty on any `read_dir` error.
- On a systemd-managed Ollama (models owned by a dedicated `ollama` user, as on this machine), that scan can't see anything — producing a false "No local Ollama models found" even though Ollama correctly reports a model over HTTP.
- `readiness()` now parses the `/api/tags` body for the real model list, only falling back to the filesystem scan if the body doesn't parse or names nothing.
- `main.rs`'s startup network-map check now calls `readiness()` instead of the raw filesystem scan, so it gets the same fix, falling back to the old scan only when Ollama isn't reachable at all.

## Test plan
- [x] `cargo test -p kwaainet` (minus one unrelated `--ignored` live test, see note) — 313 passed, 0 failed.
- [x] Ran the `#[ignore]`d live-Ollama test against this machine's actual systemd-managed Ollama: readiness now correctly reports "serving 1 model(s)" where it previously reported `NoModels`. (That test's *second* assertion — the whole-model-ready sentinel — still fails on Linux, but that's the pre-existing, deliberately macOS-only gate on `refresh_whole_model_ready`, unrelated to and untouched by this PR.)


dazwin (changes requested):
Thanks for this — trusting `/api/tags` over a scan that can't read `ollama`-owned manifests is clearly the right call, and the live test on the systemd host is exactly the evidence I wanted. Two things need to change before it lands, two smaller ones, and one note for the description.

### Blocking

**1. A `200 {"models":[]}` still falls through to the disk scan** (`ollama.rs`, `readiness()`)

The new code treats a parsed-but-empty list the same as an unparseable body and calls `list_local_models()`. That inverts the trust order the PR is establishing: a 200 from a running Ollama is authoritative, and a leftover manifest tree under any of the five roots the scan probes (`~/.ollama/models`, `~/Documents/Kwaai/ollama`, `/usr/share/ollama/...`, a stale `OLLAMA_MODELS`) turns "Ollama has no models" into `Ok([stale])`. On macOS that flips the whole-model-ready sentinel, the node announces state 2, and `serve_whole_model_via_ollama` registers the proxy for a model Ollama will reject.

Fall back only when the request fails or the body doesn't parse; a parsed empty list should be `Err(NoModels)`.

**2. API names carry `:latest`, the scan and the persisted config strip it** (`main.rs`, `pick_best_model` call site)

`manifest_path_to_ref` strips a bare `:latest`, so an existing `cfg.model` reads `llama3.1`, while `/api/tags` reports `llama3.1:latest`. Two consequences on an *unchanged* install:

- the exact-string keep-current check in `pick_best_model` misses, so the "stay on the configured model" branch is skipped;
- `match_score`'s full-name normalisation becomes `llama31latest`, which is not a substring of `llama318binstruct`, so the score drops from 2 to 1 (base-name match only).

Net effect: `start` prints `Switching model llama3.1 → llama3.1:latest` and rewrites `config.yaml` on every host that already had a model picked, and tag-scoring ties are now settled by Ollama's modified-time order instead of the old sorted list. `ResidentModel::matches` already trims `:latest` for exactly this reason. Normalise the API names the same way `manifest_path_to_ref` does (or compare with a `:latest`-tolerant helper).

### Should fix

**3. The `Err(_) => list_local_models()` arm re-runs the scan** (`main.rs`)

`readiness()` returns `NoModels` only after it has already walked the manifest roots and found nothing, so this arm repeats the walk for a result already known to be empty — and the comment above it says "only fall back if Ollama isn't reachable", which is not what the code does. Match `Err(OllamaNotReady::Unreachable { .. })` for the scan and map `NoModels` to `Vec::new()`.

**4. None of the three new branches has a test**

The only reachable unit test dials port 1 and returns before the parse; the live test is `#[ignore]`. Cheapest cover: pull the decision into a pure `fn models_from_tags(parsed: Option<TagsResponse>) -> Result<Vec<String>, OllamaNotReady>` and test names / empty / unparseable. That also pins the fix for (1) so a later change can't quietly flip it back.

### For the description

**5. This fixes discovery, not serving.** On the systemd host that motivated the PR, `start` will now select and persist an Ollama short ref, but every serve-time path (`build_inference_state` → `resolve_model_blob` → `find_manifest`, plus the equivalents in `shard_api.rs`) still resolves `cfg.model` through the same unreadable manifest scan. So the symptom moves from an honest "No local Ollama models found" at start to a "Model not found" on the first inference. Fine to leave that for a follow-up, but say so — and consider having `collect_manifest_models` / `find_manifest` log the `EACCES` instead of returning nothing silently, since that silence is what hid this in the first place.

Minor: this is the fourth independent `GET /api/tags` client (`NodeHealth` in `main.rs`, the watcher in `node.rs`, `rag_cmd.rs`, now `readiness()`). Not asking for the consolidation here, but a single API-only `installed_models(port)` that `readiness()` builds on would be the natural place once (1) is fixed.

Issue #212, opened 2026-09-10
Defaults should not be written to .kwaainet/config.yaml


dazwin:
It's problematic because a first-run will lock-in defaults to whatever is in that build. Future upgrades have no way to know the difference between this and a user-intended change.

Case in point, https://github.com/Kwaai-AI-Lab/KwaaiNet/pull/189, is effectively a no-op for existing installs, despite the desire to enable QUIC by default. We could write a config migration, but this might overwrite someone's intention to explicitly disabled QUIC support.

RezaRassool:
For the record: the reviewing agent reached the same conclusion about #189 — that the `Option<bool>` change fixes the mechanism but reaches no node currently on the network, because every v0.6.4–v0.6.8 install has `enable_quic: false` written to disk and the PR's own test enshrines it. That finding is now part of a written retrospective on the whole review pass, **#213** (`docs/REVIEW_INTEGRITY_RETROSPECTIVE.md`), which records this issue in its timeline and §1.2 as the contributor arriving at the finding independently.

Direct link on the PR branch: https://github.com/Kwaai-AI-Lab/KwaaiNet/blob/docs/review-integrity-retrospective/docs/REVIEW_INTEGRITY_RETROSPECTIVE.md

On the dilemma you raise — a migration can't tell a machine-written `false` from a user's — the codebase's existing answer is `migrate_default_peers`: rewrite only where the value exactly matches a known shipped default, and document that a genuinely UDP-blocked operator re-sets it. The alternative is to leave it and make `kwaainet config set enable_quic true` a documented upgrade step. Either is defensible; the retrospective's position is that the current state — the second option, silently, in a PR whose purpose is fixing the default — is not. That's a decision for @RezaRassool rather than something to settle in a comment.

RezaRassool:
One observation that makes this less risky than it reads, @dazwin.

Your concern is that a migration "might overwrite someone's intention to explicitly disabled QUIC support." Under the **old** default that intention is nearly unexpressible: `enable_quic` defaulted to `false`, so a user who wanted QUIC off had to do nothing. The only value worth setting deliberately was `true`.

So on any v0.6.4–v0.6.8 config:

- **`enable_quic: false`** is machine-written with near-certainty — it is what `save()` emitted because the field had no `skip_serializing_if`.
- **`enable_quic: true`** is unambiguously deliberate, and a migration that clears only `false` preserves it automatically.

[...]

PR #225, opened 2026-09-16
feat(shard): explain block-range auto-pinning at the moment it happens

RezaRassool:
## Summary
- The first time `kwaainet shard serve` auto-assigns a block range (no `--blocks`/`--start-block` given), it queries the DHT for a gap and writes `start_block`/`blocks`/`start_block_auto: true` back into `config.yaml`, with only a neutral "Updated config.yaml — pinned [...)" message.
- Nothing told the operator this range now persists across restarts, or how to fix it in place instead — this is what turned a single auto-assignment into several rounds of "why is this node stuck at N blocks" while diagnosing it live.
- Adds a one-time explanatory note, gated on this being the node's very first assignment (`cfg.start_block.is_none()` before the write), pointing at `kwaainet config set start_block <n>` for anyone who wants a genuinely fixed range instead of the auto-negotiated one.
- Repeat starts that just reuse an existing pin are unaffected — still silent, matching current behavior.

## Test plan
- [x] `cargo test -p kwaainet shard_cmd` — 20 passed, 0 failed (none assert on the exact print sequence, so unaffected).
- [x] Reproduced live earlier in this investigation: a fresh node's first `shard serve` auto-assignment now prints the new explanatory line right after "Updated config.yaml — pinned...".


dazwin (changes requested):
Requesting changes, on two grounds.

**The note is inaccurate in the default case.** `auto_rebalance` defaults to `false`, so once the range is written, `start_block_pinned()` is true on the next start and the gap query is skipped. The range does not "move automatically if the network's coverage needs change"; it is frozen unless the operator passes `--auto`/`--auto-rebalance` or sets `auto_rebalance: true`. The note tells the operator the opposite of what a stock install does, which is the confusion it set out to cure. It also nudges operators toward pinning by hand, which is the wrong direction for the network as a whole.

**This is #212.** Auto-assign writing its result into config.yaml is exactly the pattern that issue calls out, and `start_block_auto` exists only to tag the machine-written value so later code can tell it apart from an operator's. Rather than explaining the write-back, I would remove it: when `start_block` is unset, query the DHT for a gap on every `shard serve` start, and never persist the answer. `pick_gap_from_chain` already excludes our own peer id, so a restart lands on the same gap unless someone filled it while we were down, and moving in that case is the right outcome.

The one thing that needs a replacement is the IPC: the daemon learns the range by re-reading config.yaml on SIGHUP (`reload_block_range` / `announced_range` in node_native.rs). The shard's ready file is written empty today; have `shard serve` write the served range into it and have the daemon announce from that. Then `start_block_auto`, the pinned/user-pinned split, and the `a.auto = true` forcing in the rebalance loop all go, and migration is clean: `start_block_auto: true` on disk can only have been machine-written, so load can drop both keys.

Two small things if you keep any of this PR: the comment can be one line, and the hint at shard_cmd.rs:2611 still says `config --set`, which is not the subcommand.

PR #226, opened 2026-09-16
fix(cli): config set refuses while the node is running

RezaRassool:
## Summary
- `kwaainet config set <key> <value>` was a bare load-mutate-save with no awareness of the running daemon/shard — no lock, no signal to the running process.
- A live shard child can independently rewrite `config.yaml` on its own timer (`--auto-rebalance`, hardcoded whenever a real `kwaainet start --daemon --shard` spawns it), re-persisting `start_block`/`blocks` whenever it re-checks DHT coverage.
- Reproduced live earlier this session: `kwaainet config set blocks 32` while the daemon was running appeared to succeed (`config show` reflected it immediately), then reverted to the old pinned values on its own a few minutes later.
- `config set` now checks `DaemonManager::is_running() || ShardManager::is_running()` and refuses with `kwaainet is running — stop it first: kwaainet stop` instead of risking either direction of that race silently.

## Test plan
- [x] `cargo test -p kwaainet config` — 64 passed, 0 failed.
- [x] Live-tested against this machine's actual running node (14h uptime, shard serving): `config set blocks 32` now correctly refuses with the new error.
- [x] Live-tested the happy path: after `kwaainet stop`, `config set public_name ...` still succeeds normally.


dazwin (approved):
Approving as a stopgap — the diagnosis is right and the guard is the correct short-term answer. One required fix before merge, a few notes, and a proposal for the real fix as its own issue.

**Required: rustfmt.** The `bail!` fits on one line and rustfmt collapses it, so `cargo fmt --check -p kwaainet` in CI goes red:

```diff
-                        anyhow::bail!(
-                            "kwaainet is running — stop it first: kwaainet stop"
-                        );
+                        anyhow::bail!("kwaainet is running — stop it first: kwaainet stop");
```

**Notes, non-blocking**

- One refinement to the description: this is less a race than a second writer. The shard child re-persists the range it actually serves on every rebalance tick (https://github.com/Kwaai-AI-Lab/KwaaiNet/blob/af214699/core/crates/kwaai-cli/src/shard_cmd.rs#L406-L419), so a hand-set `blocks` is overwritten by reality rather than by timing.
- Both `is_running` calls only check that *some* process has that PID. After a crash the PID file lingers, and a reused PID makes `config set` refuse and tell the operator to `kwaainet stop` a process that isn't kwaainet. Pre-existing pattern shared with `start`, just worth knowing it's now a user-facing dead end.
- The GUI still edits `config.yaml` in place while the daemon runs (`KwaaiNetGUI/lib/src/daemon/config_file.dart`), so the CLI now refuses what the GUI does freely. Not this PR's job — it's the argument for the follow-up below.
- A test would be cheap: write the test's own PID into a temp run dir and assert the refusal.

**The larger fix — proposing a separate issue**

The canonical Unix answer is SIGHUP-to-reload, and kwaainet already has it: the daemon registers the handler at `node.rs` before writing its PID file, and the shard child sends it via `signal_reannounce()` after every persist. Two things make that insufficient today:

1. The reload is narrow — `reload_block_range` re-reads only `start_block` and `blocks`. Every other key waits for a restart.
2. The shard child, which owns the block range while it runs, has no reload path at all. A SIGHUP-driven `config set` would still lose to it.

So the real problem is ownership, not signalling. The shape I'd propose:

- Split config into operator intent vs runtime-persisted state (`start_block_auto` is already that split, half-made).
- One reload verb on the gRPC control surface, with SIGHUP (Unix) and the existing `reannounce.flag` file (Windows) as thin aliases of it. That's where daemons with a control channel have gone — Caddy, Consul, Vault, Docker — and it sidesteps Windows having no SIGHUP.
- `config set` then writes and asks the running daemon to reload, reporting which keys only take effect on restart, instead of refusing everything.

Happy to open that issue if you'd rather keep this PR small.

PR #228, opened 2026-09-16
feat(cli): start flags are ephemeral, and every relaunch reuses them

dazwin:
## Problem

Every restart path relaunched the node with no flags: `kwaainet restart`, `reconnect`, `update`, the post-auto-update respawn (`start --daemon` from `respawn_after_update`) and the GUI. Anything enabled only by argv was lost on the first restart:

- `--shard` came back only if `contribute.shards: true` was already in the file. After an auto-update, a node started with `start --daemon --shard` came back without a shard child.
- Most `start` overrides (`--port`, `--public-name`, `--identity-key`, `--no-relay`, …) never reached the daemon child at all. They mutated the in-memory config, only `--grpc-port` was forwarded, and the child reloads config from disk.
- `--mlx` / `--force-blocks` could never reach the daemon's shard child, which is spawned as a fixed `shard serve --auto-rebalance`.

## Change

Flags stay ephemeral. `config.yaml` is the operator's intent and `start` never writes it. What changes is that the running instance remembers its own shape:

- The node flags become `StartOverrides`, flattened into both `start` and `run-node`, so `start --daemon` forwards every flag to the child as argv and the child applies them in memory.
- `start --daemon` records the flags beside the PID file as `run/start-args.json` (the same idea as PostgreSQL's `postmaster.opts`). `restart`, `reconnect` and `update` relaunch with that record; the in-daemon auto-update respawn relaunches with its own argv; `stop` discards the record, so a bare `start` after a `stop` runs bare.
- `launch_daemon` is the one spawn path `start --daemon` and `restart` share.
- Two config keys for what a flag cannot reach: `grpc_port` (resolution is flag, then `KWAAINET_GRPC_PORT`, then config, then default) and `shard.backend: auto | candle | mlx`, read by the daemon's shard child. A manual `shard serve --mlx` / `--force-blocks` applies to that run only.
- `ContributePolicy::serve_shards` and its always-false `cli_no_contribute` parameter are removed. `--shard` / `--no-contribute` are applied to the in-memory config before the policy is read, so `--shard` still wins over an unset or opted-out config and `--no-contribute` still outranks it.

## Verification

- `cargo fmt --all --check`, `cargo clippy -p kwaainet --all-targets -- -D warnings`, `cargo test -p kwaainet`: 313 passed. New tests cover the argv round trip through the flattened flags, the JSON record round trip, the in-memory policy precedence and the backend selection; the argv-emission guard was mutated and confirmed red.
- End-to-end in a sealed throwaway container on the kwaaiai-env test bed (no default route, production names pinned to loopback, production bootstrap IP verified unreachable), with a staged model snapshot so the shard child actually spawns:

| Scenario | result |
|---|---|
| `start --daemon --shard --port 9001`: no flag in `config.yaml`, record written, `run-node` argv carries `--port 9001 --shard`, shard child up, node listening on 9001 | ✓ |
| daemon SIGTERMed, relaunched with its own argv (what `respawn_after_update` builds): fresh shard child, no flag in config | ✓ |
| `kwaainet restart`: shard child back, `--port 9001` forwarded again | ✓ |
| `--grpc-port 9555` then `restart`: gRPC bound on 9555, no `grpc_port` in config | ✓ |
| `--no-contribute` then `restart`: nothing spawned, no flag in config | ✓ |
| `stop` removes the record; a later bare `start --daemon` runs a bare `run-node` with no shard child | ✓ |
| `config set shard.backend mlx` drives a flagless `shard serve` | ✓ |

The same scenarios on the bed's 0.7.0 binary lose the shard child and the ports on every relaunch. The GitHub download step of auto-update itself was not exercised (sealed container); everything after it was.

## Not in this PR

First of the process-management changes. Still open: `run-node` owning and supervising its shard/storage children (today they are detached orphans tracked by PID file), the `ShardManager::stop_process` `waitpid`/`ECHILD` false success, SIGTERM handling in `shard serve` / `storage serve`, and the Unix `kwaainet update` path stopping only the daemon before the binary swap.


RezaRassool:
Reviewed at `5937f434fdea101eaf4c7b213604ef0fdeb32274`. This is sound and I'd
like to merge it — one thing worth three lines from you first, since you're
actively on the stack.

## `typed()` doesn't exercise three of the twelve flags

The design's safety rests on `to_argv()` emitting every field of
`StartOverrides`: a field that isn't emitted silently doesn't survive a restart,
which is the exact bug class this PR fixes.
`argv_round_trips_through_the_flattened_flags` is the right guard for that — it
parses `to_argv()` back through clap and asserts equality — but it can only
catch a dropped field if the fixture sets that field to a non-default value, and
`typed()` leaves three at their defaults:

```rust
public_ip: None,
announce_addr: None,
no_contribute: false,
```

So I mutated `to_argv()` to drop two of them:

```rust
// opt("public-ip", self.public_ip.clone());
// opt("announce-addr", self.announce_addr.clone());
```

and ran the suite:

```
test result: ok. 313 passed; 0 failed; 1 ignored
```

All 313 pass with two of twelve flags silently dropped. The code is correct as
written — I checked, all twelve are emitted — but the guard would not notice if
that stopped being true.

It reads as worth fixing here rather than later because `--public-ip` and
`--announce-addr` are precisely the NAT-addressing overrides whose loss across a
restart this PR exists to prevent. Giving all three non-default values in
`typed()` closes it.

## Minor, your call

`read_start_args()` ends in `.ok()` / `.unwrap_or_default()`, so a truncated or
corrupt record brings the node back **bare** — no `--shard`, no `--port` — with
nothing logged. `write_start_args` warns when it can't write; the read path is
silent when it can't parse. Safe direction, but quiet about it.

## Verified sound

- Merges clean onto current `main`, which has moved 8 commits (#221–#226, #229)
  since this branched. 313 tests, `clippy --all-targets -- -D warnings`, and
  `fmt --check` all clean on the **merged** tree, not just the branch.
- The `ContributePolicy::serve_shards` removal preserves semantics exactly. I
  checked the old `self.shards || (shard_flag && !no_contribute)` against the new
  apply-order for all four flag combinations; applying `no_contribute` after
  `shard` gives identical results.
- Record lifecycle is right: written at `start --daemon`, removed by `stop`, read
  by `restart`/`reconnect` and the in-daemon respawn. No record → bare start.
- git and the API agree on scope: 9 files, +530/-274.

## On the stack

Noted that #231 is based on this branch, and confirmed `#231`'s head contains
this one. When this merges I'll leave `feat/config-is-intent` undeleted so GitHub
retargets #231 to `main` rather than orphaning it. After the squash you'll want
`git rebase --onto origin/main <the-landed-sha>` on #231 rather than merging
`main` in, since none of this branch's original commits will be an ancestor of
`main`.

One small thing: #231's body opens with "Second PR of GitHub stack #232", and
`232` resolves to neither a PR nor an issue in this repo — worth correcting the
pointer.

RezaRassool (commented):
Reviewed at `5937f434fdea101eaf4c7b213604ef0fdeb32274`.

Solid. The core move — `StartOverrides` flattened into both `start` and `run-node`, applied in memory, forwarded as argv, recorded beside the PID file — is the right shape. The one place `cfg` gets persisted after `apply_to` (`main.rs` → `Command::Start`, the model-selection branch) correctly rebuilds from `cfg.reloaded()` and copies only the model fields, so the ephemeral contract holds; I checked every `save()` in the `Start` arm and every caller of `spawn_daemon_child` / `relaunch_args`. No field was lost moving `StartArgs` → `StartOverrides` (12 in, 12 out).

Three findings, none blocking. The first is the one I'd want fixed before this lands.

## 1. `start-args.json` has no `#[serde(default)]` — a future bool flag silently drops every recorded flag (medium)

`cli.rs → StartOverrides` is now an on-disk format, but derives `Deserialize` without `#[serde(default)]`. `Option<T>` fields tolerate absence; **`bool` fields do not**. Verified rather than assumed, with a standalone probe of the same struct shape:

```
missing bool field   -> Err: missing field `new_flag`
missing Option field -> Ok(V2 { ... })
```

`daemon.rs → DaemonManager::read_start_args` is `.ok().and_then(|s| serde_json::from_str(&s).ok()).unwrap_or_default()`, so that parse error becomes `StartOverrides::default()` with nothing logged.

Failure scenario: this ships; a node runs `start --daemon --shard`; a later release adds any new boolean flag; the operator upgrades via installer or package manager and runs `kwaainet restart`. The new binary cannot parse the record written by the old one, falls back to default, and the daemon comes back **without `--shard`** — the exact bug this PR exists to fix, reintroduced by format drift. (`kwaainet update` self-heals, because there the *old* binary is the one reading the record and the relaunched `start --daemon` rewrites it. The installer and package-manager paths do not.)

Two lines: `#[serde(default)]` on the struct, and a `warn!` in `read_start_args` when the file exists but will not parse, so the silent case becomes a loud one.

## 2. `shard serve` blames `--mlx` for a config setting (low)

`shard_cmd.rs → cmd_shard_serve`:

```rust
let backend = args.backend(&cfg);
let mlx = backend == crate::config::ShardBackend::Mlx;
...
if mlx && !cfg!(feature = "mlx") {
    bail!("--mlx needs a binary built with `cargo build -p kwaainet --features mlx`");
}
```

With `shard.backend: mlx` in config and no flag typed — the daemon's own shard child, which is the whole point of the new key — the operator is told to stop passing a flag they never passed. Worth naming `shard.backend` when the backend came from config; this PR is otherwise careful about saying where a setting came from.

## 3. `restart` stops the daemon, then bails on a bad config (low)

`main.rs → Command::Restart` now runs `mgr.stop_process()?` and then `KwaaiNetConfig::load_or_create()?`. The old code was `load_or_create().unwrap_or_default()`. With a malformed `config.yaml` the node is now left stopped and the command errors, where before it came back on defaults. Refusing to silently run a node on defaults is arguably the better failure — but the ordering makes it "stopped and gone". Either load before stopping, or say so in the error.

## Merge note

Merge order is forced by the stack: this one first, then #231 rebases with `git rebase --onto origin/main 5937f434` (squash merging breaks stacked branches). Don't delete `feat/config-is-intent` until GitHub has retargeted #231.

PR #231, opened 2026-09-17
feat(cli): run-node supervises its shard and storage children

dazwin:
Second PR of GitHub stack #232, on top of #228 (`feat/config-is-intent`). Its base is that branch, so this diff is only the three commits of this PR; merge #228 first and GitHub retargets this one to `main`.

## Problem

The shard and storage servers were spawned detached (`setsid`) by the short-lived `kwaainet start` process, their handles forgotten, tracked by PID file only. Nothing waited on them: a crashed shard stayed crashed, a SIGKILLed daemon left its children serving into a socket that no longer existed, and `stop`, `restart` and `update` each stopped the three in their own order. Foreground `start` and the launchd/systemd units spawned no children at all. Across an auto-update the old daemon's children outlived it, so the new daemon's shard doubled up on memory and its storage child lost the port.

## Change

`run-node` is the parent of its workers, in a new `supervisor` module.

- `supervisor::plan` decides from the contribution policy which children to run (the gating that lived in `launch_daemon`). `Supervisor` spawns each once the control socket answers a bare connect, waits on it, restarts it with exponential backoff (1 s to 60 s, reset after a stable minute) and terminates it, SIGTERM then kill after 5 s, when the node shuts down. No `setsid`: children share the daemon's session and process group.
- Before spawning anything it sweeps children a previous daemon left behind, by their PID files, so an upgrade from a pre-supervisor binary needs no manual restart.
- A supervised child's stdin is a **lifeline pipe** whose only write end the daemon holds. The child reads it on a detached thread from process start (the minutes of DHT startup included); EOF means stop. The kernel closes the write end when the daemon dies, however it dies, so the signal is exact and immediate on every platform: no PID to probe or see reused, no zombie case, no process handle, no `unsafe`, no platform-specific code. Closing the lifeline is also how the supervisor asks a child to stop, which gives Windows a graceful stop request; SIGTERM backs it up on Unix. `KWAAINET_SUPERVISOR_PID` marks stdin as a lifeline (a manual run has a terminal or /dev/null there). If the child's graceful path is not reached within 5 s, a watchdog removes the child's own PID file and exits. An earlier revision of this PR used a PID poll on Unix and a `SYNCHRONIZE` handle wait on Windows; testing on real Windows and macOS found three defects in that approach (handles keeping dead PIDs "alive", tokio's blocking pool being joined on runtime drop, stale sysinfo entries), and the pipe replaced it with less code.
- `shard serve` and `storage serve` stop on SIGTERM as well as Ctrl-C. A supervised `storage serve` exits non-zero on a control-socket or handler-registration failure instead of degrading to HTTP-only mode, so the supervisor retries it rather than running a store no Bob can reach; a manual `storage serve` keeps HTTP-only mode. It removes its PID file on every exit, bind failure included.
- `ShardManager::stop_process` no longer uses `waitpid` on a process that may not be its child (ECHILD was read as "exited" after half a second). All three managers, the daemon's included, share `stop_pid`: SIGTERM, a zombie-aware liveness probe, SIGKILL fallback; on Windows the crate's single `taskkill`, under `cfg(windows)`. A stale daemon PID file no longer makes `stop` fail.
- `start` no longer sleeps 3 s and spawns children itself: it launches `run-node` and prints the plan. `stop`, `restart` and `update` stop the daemon first, then sweep what an older binary left, after giving a dying daemon's children a few seconds to leave on their own. The startup update hint honours `KWAAINET_NO_AUTO_UPDATE`.
- Removed: `spawn_shard_child`, `spawn_storage_child`, the dead `libc_setsid` extern.

## Tests run

All on the final commit unless noted.

- **CI-equivalent, locally (macOS, Rust 1.93)**: `cargo fmt --all -- --check`; `cargo clippy --all-targets -- -D warnings` on the workspace; `cargo test --all` (1367 passed, 0 failed); `cargo test -p kwaainet --no-default-features --features storage,rag --bin kwaainet -- ipv6`; `scripts/check_memory_integrity.py`; `scripts/check_orphaned_docs.py --base origin/main`. All clean.
- **Unit tests**: plan decisions driven through a probe-free `plan_with` (the `--shard` waiver, config opt-in still gated on model and RAM, opt-out reason, `--no-contribute`, storage needs an initialised store), a skipped plan starting no tasks, and the own-PID-file helper. Each guarded line was mutated in turn and its test confirmed red.
- **Linux, sealed container on the kwaaiai-env test bed** (no default route, production names pinned to loopback, production bootstrap IP verified unreachable, `--init` so PID 1 reaps, staged fake model so the shard child really spawns). 33 of 33: parent link; `kill -9` on the shard → restarted with the exit logged; `stop` → no orphans, nothing to sweep; `kill -9` on the daemon → shard exits by itself; foreground `start` spawns the shard and SIGTERM takes both down; storage child survives `restart` as a fresh process with no churn; supervised `storage serve` with no daemon exits 1 with no pid file while a manual one keeps HTTP-only mode; **upgrade emulation with the real v0.7.0 binary as the old daemon**: its detached children are swept, the new pair runs under the new daemon, no churn.
- **macOS, local** (isolated `KWAAINET_HOME`, `KWAAINET_SOCKET`, `HF_HOME`, explicit `--model` so no map fetch, unroutable peers). 17 of 17: the same parent-link, crash-restart, stop, parent-death and upgrade-sweep scenarios on the whole-model Ollama path.
- **Windows 11, real hardware via a remote session** (release build, isolated env, remote TCP audited: only SYN_SENT to the two test addresses). `cargo check --workspace --all-targets` passes with no new warnings. All scenarios pass on the lifeline commit, with timings: shard gone 0.08 s after a daemon kill once registered, ~5 s when killed during DHT startup (the watchdog); identical with a PowerShell-held process handle on the parent, the case that defeats PID probes; shard and storage together each see their own EOF, so write ends do not leak between siblings; `stop`/`restart` without false "Orphaned"; storage on a Hyper-V-reserved port exits and restarts with 2/4/8/16/32 s backoff, every attempt really exiting and no stale pid file; supervised `storage serve` with no daemon exits 1 at its own error; **upgrade emulation from the v0.7.1 release zip** swept both old children. The last commit (watchdog removes its own PID file) was found by this run, is covered by a unit test, and was re-verified on Windows at the final commit: the pid file is written during the run and gone after the watchdog exit.
- **Not exercised**: the GitHub download step of auto-update itself (sealed environments); everything after it was.

## Follow-ups, not in this PR

- The kwaaiai-env node entrypoint launches `storage serve` for Eves itself; once the bed runs this, `run-node` spawns it and the entrypoint's copy should go.
- `kwaainet stop` on Windows is still a hard kill of the daemon (`taskkill /F`), so the node skips its DHT unannounce there; the children now stop gracefully via the lifeline.
- `stop` and the orphan sweep act on whatever PID a file records, with no check that it is still a kwaainet process; PID reuse after a crash or reboot could hit an unrelated process. Pre-existing, and worth its own change (record the start time or match the executable).
- No cap on restart churn for a child that dies the same way every time (an OOM shard is retried every 60 s forever).
- Readiness is still the `shard.ready` sentinel and `status` still reads PID files; moving that onto the control socket and gRPC is separate work.


RezaRassool (commented):
Reviewed at `652c1533fd78b1d56648916adee32b7fb04c9801` (diff read against its base #228 at `5937f434fdea101eaf4c7b213604ef0fdeb32274`).

The lifeline-pipe design is genuinely good. EOF on an inherited pipe is exact, immediate and platform-uniform, needs no `unsafe`, and sidesteps both traps the PR describes hitting: PID reuse, and a held handle keeping a dead PID readable as alive. The `stop_pid` consolidation correctly kills the old `waitpid`/ECHILD false success. I confirmed the children write their own PID files now that `spawn_shard_child` / `spawn_storage_child` are gone (`shard_cmd.rs:684`, `storage.rs:207`, plus the whole-model path in `daemon.rs:643`), so `status` and the orphan sweep still have a handle on them. I also checked that `mod libc { setsid }` is still live — `spawn_daemon_child` uses it for the daemon itself — so it is not a leftover.

Four findings, none blocking. The first is a straight regression against current `stop` behaviour and is a one-line guard.

## 1. `kwaainet stop` no longer stops orphans when the daemon PID file is gone (medium)

`main.rs → Command::Stop` is now:

```rust
mgr.stop_process()?;          // read_pid().context("No daemon is running")?
DaemonManager::remove_start_args();
print_success("KwaaiNet daemon stopped");
stop_orphaned_children();     // never reached if the ? fired
```

Before this PR the shard and storage managers were stopped *above* that `?`. `DaemonManager::read_pid` returns `None` when the PID file is absent — which is exactly what a cleanly-exited pre-supervisor daemon leaves behind, with its detached children still running.

Failure scenario: the old binary's daemon exits cleanly (PID file removed), its detached shard and storage children keep running, the operator upgrades and runs `kwaainet stop`. They get `Error: No daemon is running` and the shard server keeps holding its VRAM. The sweep does still happen on the next `start` (via `Supervisor::start → sweep_orphans`) or `restart`, so this is recoverable — but `stop` is the command an operator reaches for here, and it is the one that now refuses.

`Command::Restart` already has the right shape — `if mgr.is_running() { mgr.stop_process()?; }` followed by an unconditional `stop_orphaned_children()`. Mirroring that guard in `Stop` fixes it.

## 2. `sweep_orphans()` blocks a tokio worker for up to 10 s (low-medium)

`supervisor.rs → Supervisor::start` is a sync fn called from inside the runtime (it `tokio::spawn`s, so it must be), and it opens with `sweep_orphans()` → `ShardManager::stop_process` → `daemon.rs → stop_pid`, whose wait loop is `std::thread::sleep(Duration::from_millis(250))` against a 5 s deadline — twice, shard then storage.

On the upgrade path this sweep exists for, that is up to 10 s of a blocked runtime worker during node startup, concurrent with the gRPC server already spawned in `run_node`. `spawn_blocking`, or moving the sweep inside the supervise task, keeps it off the reactor.

## 3. A process-group SIGTERM respawns a child mid-shutdown (low)

`supervisor.rs → supervise`, the `status = child.wait()` arm. Children deliberately share the daemon's session and process group (no `setsid`), so a Ctrl-C on a foreground `start` SIGTERMs the daemon and its children together. The child then exits on its own, so the task takes the `wait()` arm rather than the `stopped()` arm: it logs "restarting in 1s" and sleeps. The inner `select!` on `stopped` only aborts that sleep if the stop flag flips within the second, and `run_native_node`'s DHT unannounce routinely takes longer than that.

The result is one spurious shard respawn — model load, VRAM allocation — torn down a few seconds later when `children.shutdown()` is finally reached. `kwaainet stop` is unaffected (it signals the daemon PID, not the group); this needs a real process-group signal, i.e. the foreground mode this PR added child spawning to. A `if *stop.borrow() { return; }` immediately after `child.wait()` closes it.

## 4. The `Supervisor` doc comment contradicts the code (nit)

> Dropping it without `shutdown` leaves the supervise tasks running; `kill_on_drop` still reaps the children when the runtime goes down.

Dropping `Supervisor` drops its `watch::Sender`, which makes `rx.changed()` return `Err` inside `stopped()` — which that helper treats as stopped and returns from. So the tasks do wake, take the `stopped` branch, `terminate` their children and return. The real behaviour is better than the documented one, but in a module where the comments carry this much of the design, the mismatch is worth correcting.

## Merge note

#228 has to land first; then rebase this one with `git rebase --onto origin/main 5937f434` rather than merging `main` in, since the squash will leave none of #228's commits as ancestors of `main`.

RezaRassool:
#228 is merged (`19f288d4`), GitHub has retargeted this to `main`, and the branch rebased cleanly — head moved `652c1533` → `86b279fc`.

Re-checked the review above against the new head rather than assume it carried over:

- `supervisor.rs` is byte-identical between `652c1533` and `86b279fc`, so findings **2** (`sweep_orphans()` blocking a tokio worker for up to 10 s inside `Supervisor::start`), **3** (a process-group SIGTERM respawning a child mid-shutdown in `supervise`'s `child.wait()` arm) and **4** (the `Supervisor` doc comment) stand verbatim.
- `main.rs` differs only by what `main` contributed in the eight commits this branch was behind; the `Command::Stop` arm is unchanged, so finding **1** — `stop_orphaned_children()` sitting behind `mgr.stop_process()?`, which bails with "No daemon is running" when the PID file is absent — also still stands.

So the review is accurate at `86b279fc`; only the SHA in its header is stale.

Nothing there blocks the merge. Finding 1 is the one I'd like fixed first, since it is a straight regression against the pre-PR `stop` ordering and the fix is the guard `Command::Restart` already uses:

```rust
if mgr.is_running() {
    mgr.stop_process()?;
}
```

Also worth folding in while you are here: the `#[serde(default)]` fix on `StartOverrides` from the #228 review. That struct is an on-disk format now, and the first new `bool` flag added to it silently drops every recorded flag on relaunch.

CI is re-running after the rebase; I am watching it.

dazwin:
Thanks — all four addressed, plus the `StartOverrides` one from #228 and one regression the e2e turned up that neither of us had spotted. Head is now `b900b7e8` (four commits on top of `86b279fc`, no rebase).

**1. `stop` with no daemon PID file — fixed (`f2812619`).** `stop` now attempts the daemon, always sweeps orphans, and returns "No daemon is running" only when there was nothing else to stop either, so the exit code for a genuinely empty `stop` is unchanged. I went slightly past the `is_running()` guard because a stale PID file should still be cleared. While there: the 5 s "children are leaving by themselves" wait is skipped when no daemon was just stopped — a pre-supervisor orphan never leaves by itself, so that was 5 s of nothing on exactly this path. Reproduced red on `86b279fc` (exit 1, orphan alive), green after (exit 0, 0.3 s).

**2. Sweep blocking a runtime worker — fixed (`e7b0401e`).** Each child's sweep runs on `spawn_blocking`: awaited before the first spawn for a child we run, fire-and-forget for one we skip. The skip case matters — shards are opt-in now, so the upgrade this sweep exists for is often one where the new daemon does *not* run a shard and the old one's is still holding VRAM.

**3. Group SIGTERM respawn — fixed (`e7b0401e`), but not with the suggested check.** `if *stop.borrow() { return; }` after `child.wait()` would not have closed it: the flag is only set by `children.shutdown()`, which is after `run_native_node` returns — the same delay that makes the backoff `select!` miss it. Instead the supervise loop also stops on the daemon's own SIGTERM/Ctrl-C (`stopping()` = stop flag ∨ `stop_requested()`), held as one pinned future for the life of the task so a signal landing between two `select!`s is not lost. Side effect: on `kwaainet stop` the children are now terminated at signal time rather than after the DHT unannounce, which is the better order anyway — they can still reach the control socket. In the sealed bed the pre-fix binary logs `shard serve exited … (signal: 15 (SIGTERM)); restarting in 2s`; the fixed one logs `shard serve stopped`. Unit test raises SIGTERM against the future with the flag unset.

**4. `Supervisor` doc comment — corrected (`e7b0401e`),** with a test pinning the behaviour (closed channel resolves `stopping`).

**`#[serde(default)]` on `StartOverrides` — done (`b900b7e8`),** plus the `warn!` when the record exists but will not parse, and a test that parses `{"shard": true}`; it fails with ``missing field `no_gpu` `` without the attribute.

**Not in the review — supervised shard had no PID file until the model loaded (`269eff8f`).** `spawn_shard_child` used to write it at spawn; with that gone, the only write left was the one after model load. For those minutes `status`, the orphan sweep and `announce_shard_loading()` (live PID ∧ no ready sentinel) could not see the shard, so a supervised node never announced `Loading`. A supervised `shard serve` now writes it at start. `storage serve` already did.

Verification: fmt / clippy `-D warnings` / `cargo test -p kwaainet` clean; sealed-container e2e — original S1–S6 20/20, new S7–S10 (group SIGTERM, orphan-without-PID-file, sweep of a skipped child, sweep-then-own-shard) 12/12 on the fixed binary, red on S8 and S10 against `86b279fc`. CI is green on `b900b7e8`, including the Windows check — the one place the `#[cfg(windows)]` call site of `stop_orphaned_children` gets compiled.

PR #234, opened 2026-09-21
fix(cli): auto-update fails loudly, streams its download, and trusts the OS roots

dazwin:
## Why

Windows operators report nodes that never upgrade themselves and have to be updated by hand. Reproduced on real Windows hardware (isolated node, stock v0.7.0 release → v0.7.1):

- **The stock path works** on a clean CPU box: check → download → swap → respawn in 5.1 s. The defects are in what happens when something goes wrong.
- **A failed release check is completely silent.** `maybe_auto_update` turned every error into `None`: no log line at any level (verified at `debug`), no cache entry, retried every ~5 min tick. A node that cannot reach `api.github.com` is indistinguishable from one with auto-update off.
- **`download_to` could not fetch the Windows `cuda-full` archive.** 970 MB under a 120 s *whole-request* timeout needs 68 Mbit/s sustained, and the body was buffered in RAM. Measured on a 44 Mbit/s link: failure at 120.6 s on every tick, working set climbing to ~775 MB, nothing written to disk, ~8 GB/hour re-downloaded forever. Any NVIDIA box installed with the cargo-dist installer (CPU zip, no `cublas64_12.dll` beside the exe) takes this path.
- **The updater trusted the bundled roots only**, so TLS inspection (AV HTTPS scanning, corporate middleboxes) breaks the check while the PowerShell installer, which uses the Windows store, keeps working. This is the leading explanation for the field reports but is **not confirmed** on an affected machine.

## What changes

- A failed check logs `WARN Auto-update check failed, no retry for at least …` with the full error chain. Check and install failures back off: 15 min, doubling, capped at 6 h. In-memory only, so no on-disk format changes. The delay is a floor; the retry rides the next announce tick.
- `download_to` streams to disk through a 1 MiB `BufWriter` under a 30 s connect timeout and a 60 s stall timeout, and removes a partial file on failure.
- The update runs on its own thread and current-thread runtime, reporting to the select loop over a channel. The old 120 s timeout was also the only bound on how long the update held the loop, and the install blocks (installer script, `tar`, a 1 GB extract), which no runtime worker should. On shutdown a download is cancelled at its next chunk. An install that has already started cannot be interrupted; shutdown waits up to 8 s for it (`kwaainet stop` SIGKILLs 10 s after SIGTERM) and then exits with a warning. That is no worse than before, when the install ran inline and a stop signal during it was simply lost, but it is **not** a guarantee that a swap is never cut short; see the rollback item below. A panic in the update thread is logged and backs off.
- The updater's clients (and only those) add the OS trust store to the bundled roots, loaded once on a blocking thread. Each OS cert is kept only if a client builds with it: `add_root_certificate` fails the whole `build()` on one malformed root, where reqwest's own native-roots feature skips them, and OS stores do hold such roots. An earlier revision enabled `rustls-tls-native-roots` workspace-wide; that made every one of the workspace's 42 client builds load the trust store synchronously (~133 ms each on macOS), several of them on the node's select loop, so it was dropped. Measured at this head: a plain client build is back to ~85 µs, the one-off store load is ~260 ms, and adding 159 roots costs ~65 µs per updater client.
- A non-2xx from GitHub is reported as such instead of `missing field tag_name` (rate-limit 403).
- `config.yaml` with a UTF-8 BOM parses. PowerShell 5.1 `Set-Content -Encoding utf8` and Notepad write one; with two or more lines serde_yaml fails with "more than one document" and the node refuses to start. Stock 0.7.0 and 0.7.1 both have this. Both config readers now parse through one function.

## Deliberately not here

- **`system-proxy`**: hyper-util's Windows matcher has no loopback bypass and the node's many `localhost` clients (VPK health, Ollama watchdog, RAG) would be proxied on exactly the machines it targets. Needs `.no_proxy()` on the loopback clients first. (Already true today for `HTTP_PROXY` env vars: `check_vpk_health` was observed going through the proxy.)
- How the updater picks `cuda-full` vs the lean zip, and whether CUDA should be redistributed at all.
- A post-respawn liveness check / rollback to `kwaainet.exe.old`. Today a new binary that cannot start leaves no node running while the log says "respawned daemon with new binary".
- After an auto-update the old `storage serve` survives on the old image and the new daemon never spawns its own. That is process ownership, i.e. #231.

## Tests

Each new test was run against the code with its fix removed and confirmed red, by the author and again independently from a fresh clone. The download tests were also looped 20x, and 10x under CPU load, with no flake:

| Test | Fix removed | Without the fix |
|---|---|---|
| `download_outlasts_the_stall_timeout_while_bytes_flow` | whole-request timeout + `bytes()` restored | red |
| `stalled_download_fails_and_leaves_no_partial_file` | partial-file cleanup removed | red (guards the new cleanup; the old code never wrote a partial file) |
| `cancelled_download_stops_and_leaves_no_partial_file` | cancel check removed | red |
| `rate_limited_check_reports_the_http_status` | `error_for_status` removed | red: `missing field tag_name` |
| `update_backoff_doubles_caps_and_resets` | doubling removed | red |
| `check_results_drive_the_backoff` | failure no longer arms the backoff; finding an update resets the count | red, both |
| `a_malformed_os_root_is_dropped_not_fatal` | root filtering removed | red |
| `config_with_utf8_bom_parses` | BOM strip removed from the shared parser | red: "more than one document" |

**Not covered by a test:** the `ready()` gate at the top of `maybe_auto_update`, the thread/channel handoff into the select loop, the bounded shutdown wait, the panic path, and OS-roots TLS against a real intercepting root. Those rest on the manual runs below.

## Verification

macOS: `cargo fmt --all --check`, `cargo clippy -p kwaainet --all-targets -- -D warnings`, workspace `cargo clippy --all-targets -- -D warnings`, `cargo test -p kwaainet` (321 passed), and a live `kwaainet update --check` against GitHub with the OS-roots client.

Windows 10, release build reporting 0.7.0, isolated node. **These runs were made at b4647e69. The update-thread and OS-roots changes (f4a106aa, 675c41f9) came after them and are being re-verified on the same machine, including a Ctrl-C during a download and the root filter against a real Windows certificate store; this section will be updated with those results before the PR leaves draft.**

| Scenario | Stock | b4647e69 |
|---|---|---|
| 970 MB `cuda-full` on a 44 Mbit/s link | fails at 120.6 s every tick, RAM → 775 MB | completes in ~187 s, working set flat at 23 MB, zip grows on disk, `%TEMP%` cleaned |
| Control socket during the download | n/a | `kwaainet status` answers immediately |
| Release check forced to fail (dead proxy), `info` logging | no log line, no cache file | one WARN with URL + OS error, no attempt for three ticks, then 1800 s |
| BOM + CRLF `config.yaml` | exit 1, "more than one document" | parses, daemon starts |

TLS inspection itself has not been reproduced; no test machine has an intercepting root installed.

To reproduce with a lower version: all 14 version pins in `core/Cargo.toml` must be lowered, not only `[workspace.package]`. Force a check by deleting `run/update_check.json`; any CLI command rewrites it.


RezaRassool:
Reviewed at `675c41f904bd3086658d09f9137e4f348c11eb4d`.

Not a defect in this PR — a gap it leaves open, found while trying to reproduce the field failure it targets on the one Windows box we have.

## The gates ahead of your warnings are still silent

This PR makes every *failure* to update audible: `note_check_result` warns and backs off, a failed install warns. Nothing reports the gates that return before any of that:

- `node.rs → maybe_auto_update`, first `if`: a packaged install, or `KWAAINET_NO_AUTO_UPDATE` — `return None`, silently.
- `node_native.rs → run_native_node`, the `next_announce` arm: `KwaaiNetConfig::load_or_create().map(…).unwrap_or(false)` — `contribute.auto_update: false` **and** a config that will not parse, both silently.

A node that was never going to update is indistinguishable from one already on the latest release.

## Why it matters here

metro-win (RTX A5000, NVIDIA, cargo-dist install) has been sitting on v0.7.0 since 2026-09-11 — through the v0.7.1 release on 09-16, ten days, daemon up the whole time. The cause is not any of the four defects this PR fixes:

- `api.github.com` answers 200 from that box; `X-RateLimit-Used: 3` — not rate-limited.
- TLS chain is genuine Sectigo, `Verify return code: 0` — **no interception**, so the OS-roots hypothesis does not apply to this machine.
- `cublas64_12.dll` is already beside the exe, so it takes the lean 18.9 MB CUDA zip, not the 970 MB `cuda-full` path.
- `config.yaml` has no BOM and parses.

It is `KWAAINET_NO_AUTO_UPDATE=1`, sitting in `HKCU\Environment` (set during the from-source CUDA work and never cleared), inherited by the daemon at spawn. `run/update_check.json` was untouched for the whole ten days, there is no auto-update line at any level in 200 MB of log, and `kwaainet update --check` — which consults none of these gates — kept reporting v0.7.1 as available.

Proof it is the gate and not a failing check: I forced a check so the cache was fresh and named 0.7.1, which makes `check(false)` a pure cache read that cannot fail. Three consecutive announce ticks (19:45:59, 19:51:02, 19:56:16, each logging `Re-announcing to DHT`, which comes *after* the auto-update block) produced no `Auto-update:` line. **This PR as it stands would not have made that node say anything.**

## Suggested patch

`run_native_node` logs the reason once at startup — INFO when somebody chose it, WARN when nobody did, since an unreadable config disables auto-update by accident. The escape-hatch string is parsed in one place so the gate and the line reporting it cannot drift, and the order is resolved by a pure function so it is testable without touching the process environment or the disk.

On metro-win the new line reads:

```
INFO kwaainet::node: Auto-update is off: KWAAINET_NO_AUTO_UPDATE is set in this environment — clearing it takes a restart
```

(verified by calling `auto_update_off()` on that machine; it returns `ByChoice("KWAAINET_NO_AUTO_UPDATE is set in this environment — clearing it takes a restart")`).

Each test was confirmed red with its fix removed: reordering the gates reds `the_first_gate_to_fire_is_the_one_reported`; reporting an unreadable config as a choice reds `an_unreadable_config_is_a_fault_not_a_choice`; loosening the escape-hatch parse reds `the_escape_hatch_reads_the_same_everywhere`. `cargo test -p kwaainet`: 329 passed, `cargo fmt --check` clean. Clippy is clean for these files; the crate has 19 pre-existing Windows-only lints (`rag_cmd.rs` unused `sock`, `updater.rs:305` nonminimal_bool, etc.) that CI does not see on Linux.

Applies on `675c41f9` with `git apply`:

<details>
<summary>patch</summary>

```diff
diff --git a/CHANGELOG.md b/CHANGELOG.md
index bd4ec7aa..33d16aad 100644
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -62,6 +62,7 @@ The historical record up to v0.3.7 follows.
 - The update download streams to disk under a stall timeout. It used a 120 s whole-request deadline with the body held in RAM, which the 970 MB Windows `cuda-full` archive cannot meet below ~68 Mbit/s: an NVIDIA Windows node re-downloaded it every ~5 min and never upgraded. The download also runs off the node's event loop.
 - The release check and download trust the OS certificate store as well as the bundled roots, so TLS inspection no longer breaks them.
 - A GitHub rate-limit response is reported as HTTP 403 instead of `missing field tag_name`.
+- **A node that is not going to update itself says so at startup.** The gates — a packaged install, `KWAAINET_NO_AUTO_UPDATE` in the environment, `contribute.auto_update: false`, a config that will not parse, a 1.0+ build — all returned silently, so a disabled node was indistinguishable from an up-to-date one. Each now logs which gate stopped it, as a warning when nobody chose it (an unreadable config).
 - `config.yaml` saved with a UTF-8 BOM (PowerShell 5.1 `Set-Content -Encoding utf8`, Notepad) parses; it used to stop the node starting with "more than one document".
 
 ## [v0.3.7] - 2026-03-04
diff --git a/core/crates/kwaai-cli/src/node.rs b/core/crates/kwaai-cli/src/node.rs
index 5190249d..6ac2df4e 100644
--- a/core/crates/kwaai-cli/src/node.rs
+++ b/core/crates/kwaai-cli/src/node.rs
@@ -332,6 +332,97 @@ pub(crate) fn spawn_auto_update(
         .ok()
 }
 
+/// The developer escape hatch, parsed in one place so the gate and the line
+/// that reports the gate can never disagree: set to anything but empty, `0`
+/// or `false`.
+fn env_disables(value: Option<&str>) -> bool {
+    value.is_some_and(|v| !v.is_empty() && v != "0" && !v.eq_ignore_ascii_case("false"))
+}
+
+fn no_auto_update_env() -> bool {
+    env_disables(std::env::var("KWAAINET_NO_AUTO_UPDATE").ok().as_deref())
+}
+
+/// Why this node will not install its own updates.
+///
+/// Every *failed* update now leaves a line in the log: a failed check warns
+/// and backs off, a failed install warns. The gates in `maybe_auto_update`
+/// and in the announce tick left none — they return before any of that, so a
+/// node that was never going to update looked exactly like one already on the
+/// latest release. On metro-win a `KWAAINET_NO_AUTO_UPDATE=1` left in the user
+/// environment held a daemon on v0.7.0 across the v0.7.1 release for ten days
+/// with nothing at any level in 200 MB of log to say so, while
+/// `kwaainet update --check`, which consults none of these gates, kept
+/// reporting the new version as available.
+#[derive(Debug)]
+pub(crate) enum AutoUpdateOff {
+    /// Somebody asked for this: the operator, the packaging, or the release line.
+    ByChoice(String),
+    /// Nobody asked for this — the node would update, but cannot read its config.
+    Fault(String),
+}
+
+/// Resolve the gates in the order they are actually applied. Split from
+/// [`auto_update_off`] so that order can be tested without touching the
+/// process environment or the disk.
+fn auto_update_off_for(
+    packaged: Option<&str>,
+    env_disabled: bool,
+    config_auto_update: Option<bool>,
+    pre_release: bool,
+) -> Option<AutoUpdateOff> {
+    if let Some(kind) = packaged {
+        return Some(AutoUpdateOff::ByChoice(format!(
+            "this is a {kind} package, which its package manager upgrades"
+        )));
+    }
+    if env_disabled {
+        return Some(AutoUpdateOff::ByChoice(
+            "KWAAINET_NO_AUTO_UPDATE is set in this environment — clearing it takes a restart"
+                .to_string(),
+        ));
+    }
+    match config_auto_update {
+        // `unwrap_or(false)` at the tick turns an unreadable config into an
+        // opt-out. That is the safe default, but it is nobody's decision.
+        None => Some(AutoUpdateOff::Fault(
+            "config.yaml could not be read, and the announce tick treats that as opt-out"
+                .to_string(),
+        )),
+        Some(false) => Some(AutoUpdateOff::ByChoice(
+            "contribute.auto_update is false in config.yaml".to_string(),
+        )),
+        Some(true) if !pre_release => Some(AutoUpdateOff::ByChoice(format!(
+            "v{} is past 1.0, and only pre-1.0 builds update themselves",
+            crate::updater::CURRENT_VERSION
+        ))),
+        Some(true) => None,
+    }
+}
+
+/// Why this node will not update itself, or `None` when it will. Reads the
+/// config from disk, exactly as the announce tick does.
+pub(crate) fn auto_update_off() -> Option<AutoUpdateOff> {
+    auto_update_off_for(
+        crate::updater::packaged_install(),
+        no_auto_update_env(),
+        KwaaiNetConfig::load_or_create()
+            .ok()
+            .map(|c| c.contribute.auto_update),
+        crate::config::is_pre_release(),
+    )
+}
+
+/// Say once, at startup, that this node will never update itself. Silence
+/// here means auto-update is live.
+pub(crate) fn log_auto_update_off() {
+    match auto_update_off() {
+        Some(AutoUpdateOff::ByChoice(why)) => info!("Auto-update is off: {why}"),
+        Some(AutoUpdateOff::Fault(why)) => warn!("Auto-update is off: {why}"),
+        None => {}
+    }
+}
+
 /// Check for a newer release and, if found, install it automatically.
 /// After a successful install the daemon exits cleanly so the OS service
 /// manager (systemd, launchd) or the user can restart it with the new binary.
@@ -354,11 +445,7 @@ pub(crate) async fn maybe_auto_update() -> Option<String> {
     // download or install an update from inside the running node.
     // A packaged install is upgraded by the distro package manager: installing
     // into ~/.cargo/bin here would shadow /usr/bin/kwaainet on PATH.
-    if crate::updater::packaged_install().is_some()
-        || std::env::var("KWAAINET_NO_AUTO_UPDATE")
-            .map(|v| !v.is_empty() && v != "0" && !v.eq_ignore_ascii_case("false"))
-            .unwrap_or(false)
-    {
+    if crate::updater::packaged_install().is_some() || no_auto_update_env() {
         return None;
     }
 
@@ -786,6 +873,70 @@ mod capacity_lease_dht_tests {
     }
 }
 
+#[cfg(test)]
+mod auto_update_gate_tests {
+    use super::*;
+
+    /// The escape hatch is a string, and every spelling of "off" must stay off.
+    #[test]
+    fn the_escape_hatch_reads_the_same_everywhere() {
+        assert!(env_disables(Some("1")));
+        assert!(env_disables(Some("yes")));
+        assert!(!env_disables(None), "unset means auto-update is live");
+        assert!(!env_disables(Some("")));
+        assert!(!env_disables(Some("0")));
+        assert!(!env_disables(Some("false")));
+        assert!(!env_disables(Some("FALSE")));
+    }
+
+    #[test]
+    fn a_live_auto_update_says_nothing() {
+        assert!(auto_update_off_for(None, false, Some(true), true).is_none());
+    }
+
+    /// The gates are reported in the order they are applied, so the line names
+    /// the one that actually stopped the update.
+    #[test]
+    fn the_first_gate_to_fire_is_the_one_reported() {
+        let packaged = auto_update_off_for(Some("deb"), true, Some(false), true);
+        assert!(
+            matches!(packaged, Some(AutoUpdateOff::ByChoice(ref why)) if why.contains("deb")),
+            "{packaged:?}"
+        );
+
+        let env = auto_update_off_for(None, true, Some(false), true);
+        assert!(
+            matches!(env, Some(AutoUpdateOff::ByChoice(ref why))
+                if why.contains("KWAAINET_NO_AUTO_UPDATE")),
+            "{env:?}"
+        );
+
+        let config = auto_update_off_for(None, false, Some(false), true);
+        assert!(
+            matches!(config, Some(AutoUpdateOff::ByChoice(ref why))
+                if why.contains("contribute.auto_update")),
+            "{config:?}"
+        );
+
+        let released = auto_update_off_for(None, false, Some(true), false);
+        assert!(
+            matches!(released, Some(AutoUpdateOff::ByChoice(ref why)) if why.contains("1.0")),
+            "{released:?}"
+        );
+    }
+
+    /// An unreadable config disables auto-update without anyone asking, so it
+    /// warns where a deliberate opt-out only informs.
+    #[test]
+    fn an_unreadable_config_is_a_fault_not_a_choice() {
+        let off = auto_update_off_for(None, false, None, true);
+        assert!(
+            matches!(off, Some(AutoUpdateOff::Fault(ref why)) if why.contains("config.yaml")),
+            "{off:?}"
+        );
+    }
+}
+
 #[cfg(test)]
 mod update_backoff_tests {
     use super::*;
diff --git a/core/crates/kwaai-cli/src/node_native.rs b/core/crates/kwaai-cli/src/node_native.rs
index b026ea24..cf64360f 100644
--- a/core/crates/kwaai-cli/src/node_native.rs
+++ b/core/crates/kwaai-cli/src/node_native.rs
@@ -574,6 +574,10 @@ pub async fn run_native_node(
         ollama_recovery_rx = crate::node::spawn_ollama_watcher(&config);
     }
     let mut pending_update_version: Option<String> = None;
+    // Say once why this node will never update itself. The gates are all
+    // silent returns, which leaves a node that cannot update looking exactly
+    // like one already on the latest release.
+    crate::node::log_auto_update_off();
     let (update_tx, mut update_done) = tokio::sync::mpsc::channel::<String>(1);
     let mut update_thread: Option<std::thread::JoinHandle<()>> = None;
     // Deadline for the reachability-change settle window; None = no change pending.
```

</details>

Yours to take or leave — happy to open it as a PR onto your branch instead if you'd rather not hand-apply.

